PR #1457, opened 2016-09-08
sysinfo: system logs

apahim:
- Drop the log collection code.
- Add log files to the sysinfo `files` list.
- Add `journalctl --boot --priority 5` command to the sysinfo `commands` list.

### DIFF
--- a/avocado/core/sysinfo.py
+++ b/avocado/core/sysinfo.py
@@ -404,22 +404,6 @@ class SysInfo(object):
 
         self._set_collectibles()
 
-    def _get_syslog_watcher(self):
-        syslog_watcher = None
-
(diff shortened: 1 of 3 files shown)


adereis:
Unfortunately the journal collects too much info, way more than the standard /var/log/messages file (syslog) used to. In part because these days all desktop apps (on a gnome system) tend to send their warnings to the systemd-journal. After a discussion on IRC, Amador and I agreed on using --boot --priority=5 as a compromise solution. Notice this is a config file, so we're looking for reasonable defaults. The user can fine-tune it to their preference.

apahim:
@adereis, updated based on our discussion.

PR #1502, opened 2016-09-29
use python-systemd for sysinfo journalctl

apahim:
Currently we use the journalctl command to collect the logs created
during a given test. This approach makes the KeyboardInterrupt to be
ignored by the test runner so the test ends with PASS instead of
INTERRUPTED.

This patch makes the sysinfo to use the python-systemd module, to
collect the messages, avoiding the issue and making the code less
dependent on external commands.

Reference: https://trello.com/c/dNgrjMOJhttps://trello.com/c/dNgrjMOJ

### DIFF
--- a/avocado.spec
+++ b/avocado.spec
@@ -7,13 +7,13 @@
 Summary: Avocado Test Framework
 Name: avocado
 Version: 41.0
-Release: 0%{?dist}
+Release: 1%{?dist}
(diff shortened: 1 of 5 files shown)


apahim:
systemd-python from pip requires a missing dependency (systemd-devel). Working on that.

PR #1522, opened 2016-10-06
remote.runner: be lenient with test references

apahim:
When copying tests for a remote execution, test references are making
_copy_files() to break.

This patch adds a check in _copy_files() so it can handle test
references.

Reference: https://trello.com/c/LvZrlNkV

### DIFF
--- a/avocado/core/remote/runner.py
+++ b/avocado/core/remote/runner.py
@@ -64,6 +64,9 @@ class RemoteTestRunner(TestRunner):
         paths = set()
         for i in xrange(len(self.job.urls)):
             url = self.job.urls[i]
+            test_function = None
+            if ':' in url:
+                url, test_function = url.split(':')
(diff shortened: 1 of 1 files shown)


apahim:
Ok, I like your proposal. Let me just share one thought. We currently kinda disable the loader for remote executions when the `--remote-no-copy` is provided. So the `remote_runner` receives an empty `test_suite`. Not a problem here, since you're relying on the `test_suite` only inside the `_copy_files()` and right now `_copy_files()` will always receive a 'loaded' `test_suite`. But I believe that with the current implementation of the `remote_runner`, the loader should not be locally engaged at all. Anyway, that's  something to solve if we ever decide to touch that behaviour. Unlikely given the remote discussions that are in place. You can go ahead and propose your solution. Thank you for the review.

PR #1555, opened 2016-10-21
selftests: Remove self.skipTest()

apahim:
Let's separate the check for requirements and the test code according to
the project's understanding of setup/test phases.

Reference: https://trello.com/c/mSrMLO0K

### DIFF
--- a/selftests/functional/test_basic.py
+++ b/selftests/functional/test_basic.py
@@ -554,9 +554,8 @@ class RunnerHumanOutputTest(unittest.TestCase):
         self.assertIn('skiponsetup.py:SkipOnSetupTest.test_wont_be_executed:'
                       '  SKIP', result.stdout)
 
+    @unittest.skipIf(missing_binary('echo'), 'echo binary not available')
     def test_ugly_echo_cmd(self):
-        if not os.path.exists("/bin/echo"):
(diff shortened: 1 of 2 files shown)


apahim:
[Do not use unittest `skipTest` inside tests](https://trello.com/c/mSrMLO0K/848-do-not-use-unittest-skiptest-inside-tests)

selftests/functional/test_basic.py
```diff
@@ -554,9 +554,8 @@ def test_output_skip(self):
         self.assertIn('skiponsetup.py:SkipOnSetupTest.test_wont_be_executed:'
                       '  SKIP', result.stdout)
 
+    @unittest.skipIf(missing_binary('echo'), 'echo binary not available')
     def test_ugly_echo_cmd(self):
-        if not os.path.exists("/bin/echo"):
-            self.skipTest("Program /bin/echo does not exist")
         os.chdir(basedir)
         cmd_line = ('./scripts/avocado run "/bin/echo -ne '
```
ldoktor: (the later requires the use of `self.skipTest` unless you want to run the same command twice)

apahim:
Ok, I got your point. Thank you.

PR #1556, opened 2016-10-21
selftests: Remove skip test [v2]

apahim:
v2:
Probe the binary only once and use if throughout the skip tests and the tests themselves.

v1: #1555 
Let's separate the check for requirements and the test code according to
the project's understanding of setup/test phases.

### DIFF
--- a/selftests/functional/test_basic.py
+++ b/selftests/functional/test_basic.py
@@ -96,12 +96,16 @@ class MyTest(Test):
 '''
 
 
-def missing_binary(binary):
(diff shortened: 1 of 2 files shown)


apahim:
[Do not use unittest `skipTest` inside tests](https://trello.com/c/mSrMLO0K/848-do-not-use-unittest-skiptest-inside-tests)

ldoktor:
OK, it does not want to run the checks, I ran them manually, merging...

PR #1559, opened 2016-10-21
Adding Multipath Module [v2]

narasimhan-v:
avocado.utils.multipath: Adds multipath related modules.

Changelog from v1 #1547 :
*) Used utils.service to restart and to check status of multipath service
*) Added a check to see if multipath devices exist, on all required functions, so that IndexError is not encountered.
*) Using logging.debug(open().read()) to display contents of the file, as per comments.
*) Based on comments, changed function name to device_exists
*) Removed raising of Exception, since its codepath is not existing anymore.
*) Mentioned that the module needs to be used only by root user.

Signed-off-by: Narasimhan V sim@linux.vnet.ibm.com

### DIFF
--- a/avocado/utils/multipath.py
+++ b/avocado/utils/multipath.py
@@ -1,0 +1,151 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Narasimhan V <sim@linux.vnet.ibm.com>
+
+"""
+Module with multipath related utility functions.
+It needs root access.
+"""
+
+import time
+import logging
+from . import distro
+from . import process
+from . import service
+
+
+def form_conf_mpath_file(blacklist="", defaults_extra=""):
+    """
+    Form a multipath configuration file, and restart multipath service.
+
+    :parm blacklist: Entry in conf file to indicate blacklist section.
+    :parm defaults_extra: Extra entry in conf file in defaults section.
+    """
+    mpath_fp = open("/etc/multipath.conf", "w")
+    mpath_fp.write("defaults {\n")
+    mpath_fp.write("    find_multipaths yes\n")
+    mpath_fp.write("    user_friendly_names yes\n")
+    if defaults_extra:
+        mpath_fp.write("    %s\n" % defaults_extra)
+    mpath_fp.write("}\n")
+    if blacklist:
+        mpath_fp.write("blacklist {\n")
+        mpath_fp.write("    %s\n" % blacklist)
+        mpath_fp.write("}\n")
+    mpath_fp.close()
+    logging.debug(open("/etc/multipath.conf", "r").read())
+    if distro.detect().name == 'Ubuntu':
+        svc_name = "multipath-tools"
+    else:
+        svc_name = "multipathd"
+    time.sleep(5)
+    service.SpecificServiceManager(svc_name).restart()
+
+
+def device_exists(path):
+    """
+    Checks if a given path exists.
+
+    :return: True if path exists, False if does not exist.
+    """
+    cmd = "multipath -l %s" % path
+    if process.system(cmd, ignore_status=True) != 0:
+        return False
+    return True
+
+
+def get_mpath_name(wwid):
+    """
+    Get multipath name for a given wwid.
+
+    :parm wwid: wwid of multipath device.
+
+    :return: Name of multipath device.
+    """
+    if not device_exists(wwid):
+        return
+    cmd = "multipath -l %s" % wwid
+    return process.system_output(cmd).split()[0]
+
+
+def get_multipath_wwids():
+    """
+    Get list of multipath wwids.
+
+    :return: List of multipath wwids.
+    """
+    cmd = "egrep -v '^($|#)' /etc/multipath/wwids"
+    wwids = process.system_output(cmd, ignore_status=True).strip("\n")
+    wwids = wwids.replace("/", "").split("\n")
+    return wwids
+
+
+def get_paths(wwid):
+    """
+    Get list of paths, given a multipath wwid.
+
+    :return: List of paths.
+    """
+    if not device_exists(wwid):
+        return
+    cmd = "multipath -ll %s" % wwid
+    lines = process.system_output(cmd).strip("\n")
+    paths = []
+    for line in lines.split("\n"):
+        if not (('size' in line) or ('policy' in line) or (wwid in line)):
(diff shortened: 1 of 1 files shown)


adereis:
When submitting a v(n+1) of a PR, please add a changelog, expliciting your changes from the previous versions (in the description of the Pull Request).

avocado/utils/multipath.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def form_conf_mpath_file(blacklist="", defaults_extra=""):
+    """
+    Form a multipath configuration file, and restart multipath service.
+
+    :parm blacklist: Entry in conf file to indicate blacklist section.
+    :parm defaults_extra: Extra entry in conf file in defaults section.
+    """
+    mpath_fp = open("/etc/multipath.conf", "w")
```
apahim: Please use the `with` statement when dealing with file objects. It assures that the file is properly closed, even if an exception is raised:
```python
with open("/etc/multipath.conf", "w") as mpath_fp:
    mpath_fp.write("defaults {\n")
    ....
```

Also, the content of the file could be given to a variable and then you would call the `.write(content)` only once.

avocado/utils/multipath.py
```diff
@@ -0,0 +1,151 @@
...
+        mpath_fp.write("blacklist {\n")
+        mpath_fp.write("    %s\n" % blacklist)
+        mpath_fp.write("}\n")
+    mpath_fp.close()
+    logging.debug(open("/etc/multipath.conf", "r").read())
+    if distro.detect().name == 'Ubuntu':
+        svc_name = "multipath-tools"
+    else:
+        svc_name = "multipathd"
+    time.sleep(5)
```
apahim: What are we waiting for?

avocado/utils/multipath.py
```diff
@@ -0,0 +1,151 @@
...
+    mpath_fp.write("defaults {\n")
+    mpath_fp.write("    find_multipaths yes\n")
+    mpath_fp.write("    user_friendly_names yes\n")
+    if defaults_extra:
+        mpath_fp.write("    %s\n" % defaults_extra)
+    mpath_fp.write("}\n")
+    if blacklist:
+        mpath_fp.write("blacklist {\n")
+        mpath_fp.write("    %s\n" % blacklist)
+        mpath_fp.write("}\n")
```
apahim: What happens if the system already have a `multipath.conf`? Are we going to overwrite it?

avocado/utils/multipath.py
```diff
@@ -0,0 +1,151 @@
...
+    mpath_fp.close()
+    logging.debug(open("/etc/multipath.conf", "r").read())
+    if distro.detect().name == 'Ubuntu':
```
apahim: From this point (line 47) on, it should be a helper function.

avocado/utils/multipath.py
```diff
@@ -0,0 +1,151 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Narasimhan V <sim@linux.vnet.ibm.com>
+
+"""
+Module with multipath related utility functions.
+It needs root access.
```
apahim: So let's handle this properly. We have `avocado.utils.process.can_sudo()` which returns `True`/`False` and can be used to fail early. Something like:
```python
...
if not process.can_sudo():
    raise ImportError('This module requires superuser rights.')
...
```
Then you have to execute everything that applies with `sudo=True`.

avocado/utils/multipath.py
```diff
@@ -0,0 +1,151 @@
...
+def get_mpath_name(wwid):
+    """
+    Get multipath name for a given wwid.
+
+    :parm wwid: wwid of multipath device.
+
+    :return: Name of multipath device.
+    """
+    if not device_exists(wwid):
+        return
```
apahim: This empty `return` has an implicit `None`. It's the same thing if you don't have a return. So this function will be more clear if written like this:

```python

def get_mpath_name(wwid):
    if device_exists(wwid):
        cmd = "multipath -l %s" % wwid
        return process.system_output(cmd).split()[0]

```

avocado/utils/multipath.py
```diff
@@ -0,0 +1,151 @@
...
+    :return: path checker policy.
+    """
+    if not device_exists(wwid):
+        return
+    cmd = "multipath -ll %s" % wwid
+    lines = process.system_output(cmd).strip("\n")
+    for line in lines.split("\n"):
+        if 'policy' in line:
+            return line.split("'")[1].split()[0]
+    return ""
```
apahim: This function is returning 3 different things: `None` (when the device does not exist), the `policy` or an empty string. This empty string is what bugs me. Since we have a `device_exists()`, I think we should have only the policy or `None`. Something like:
```python
def get_policy(wwid):
    """
    Docstring
    """
    if device_exists(wwid):
        cmd = "multipath -ll %s" % wwid
        lines = process.system_output(cmd).strip("\n")
        for line in lines.split("\n"):
            if 'policy' in line:
                return line.split("'")[1].split()[0]
```

That way `None` will be returned in both cases, when the device does not exist and also when there's no `policy` in `multipath -ll` result.

avocado/utils/multipath.py
```diff
@@ -0,0 +1,151 @@
...
+    :return: size of multipath device.
+    """
+    if not device_exists(wwid):
+        return
+    cmd = "multipath -ll %s" % wwid
+    lines = process.system_output(cmd).strip("\n")
+    for line in lines.split("\n"):
+        if 'size' in line:
+            return line.split("=")[1].split()[0]
+    return ""
```
apahim: Same case as in `get_policy(wwid)`

avocado/utils/multipath.py
narasimhan-v: Yes. Tester should take care of taking a backup, as demonstrated in https://github.com/avocado-framework-tests/avocado-misc-tests/pull/128

avocado/utils/multipath.py
narasimhan-v: Sure, will do

avocado/utils/multipath.py
narasimhan-v: Will change it.

avocado/utils/multipath.py
narasimhan-v: Restarting services after changing the conf file needs some time. Changes do not take effect if resarted imediately.
I could not find a better way to handle it.

apahim: So this should be after the `service.SpecificServiceManager(svc_name).restart()`, not before.

narasimhan-v: No, it is before that.

* change conf file -> restart service -> changes do not take effect.
* change conf file -> sleep(for say 5s) -> restart service -> changes take effect.

apahim: ok. Please address the other comments.

apahim:
@narasimhan-v, please send an updated version of this PR.

apahim:
*if you're still interested, of course.

narasimhan-v:
@apahim Sorry for the delay. working on it, and wil send it in a day or 2.

narasimhan-v:
Closing this PR.

PR #1566, opened 2016-10-26
Job ID: utility/command to find job results by ID

apahim:
Simple contrib script to find the Job results directory of a given Job
identified by the Job (partial) ID.

Reference: https://trello.com/c/vrp2F5n4

### DIFF
--- a/contrib/scripts/avocado-get-job-results-dir.py
+++ b/contrib/scripts/avocado-get-job-results-dir.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


apahim:
[Job ID: utility/command to find job results by ID](https://trello.com/c/vrp2F5n4/845-job-id-utility-command-to-find-job-results-by-id)

clebergnu:
@apahim even though error reporting is useful, the main goal of such a script is to hook into some other forms of automation.  To that, it's important to have a predictable output in STDOUT.  I suggest only printing the results directory, if found, to STDOUT, while error messages to STDERR.

apahim:
Oh, sure. v2 on the way.

PR #1567, opened 2016-10-26
avocado.utils.process: execute default sigint handler

apahim:
In order to exit subprocess gracefully, our library overrides the
default SIGINT handler with a custom handler that waits for the
subprocess. Since the default SIGINT handler raises a KeyboardExeption,
which is expected by the caller, and our custom handler does not, this
patch calls the default handler from inside our custom handler, after
the steps required to exit the subprocess gracefully.

Reference: https://trello.com/c/NZPg155x

### DIFF
--- a/avocado/utils/process.py
+++ b/avocado/utils/process.py
@@ -406,6 +406,7 @@ class SubProcess(object):
             def signal_handler(signum, frame):
                 self.result.interrupted = "signal/ctrl+c"
                 self.wait()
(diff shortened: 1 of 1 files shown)


apahim:
[BUG: avocado.utils.process is making KeyboardInterrupt to be ignored](https://trello.com/c/NZPg155x/842-bug-avocado-utils-process-is-making-keyboardinterrupt-to-be-ignored)

PR #1584, opened 2016-11-01
don't create a shell escaped command

apahim:
This is causing more problems than helping. Escaped commands can make
os.path.abspath() to misbehave since the command can be inside single
quotes. Let's remove this so we can behave correctly when some funky
characters are in the command.

### DIFF
--- a/avocado/core/loader.py
+++ b/avocado/core/loader.py
@@ -23,7 +23,6 @@ import imp
 import inspect
 import os
 import re
-import pipes
 import shlex
 import sys
 
@@ -691,8 +690,7 @@ class FileLoader(TestLoader):
                                                 subtests_filter)
             else:
                 if os.access(test_path, os.X_OK):
-                    return self._make_test(test.SimpleTest,
-                                           pipes.quote(test_path))
+                    return self._make_test(test.SimpleTest, test_path)
                 else:
                     return make_broken(test.NotATest, test_path)
         else:
--- a/selftests/functional/test_basic.py
+++ b/selftests/functional/test_basic.py
@@ -1,3 +1,4 @@
+# This Python file uses the following encoding: utf-8
 import aexpect
 import glob
 import json
@@ -609,7 +610,7 @@ class RunnerSimpleTest(unittest.TestCase):
     def setUp(self):
         self.tmpdir = tempfile.mkdtemp(prefix='avocado_' + __name__)
         self.pass_script = script.TemporaryScript(
-            'avocado_pass.sh',
+            'ʊʋʉʈɑʅʛʌ',
             PASS_SCRIPT_CONTENTS,
             'avocado_simpletest_functional')
         self.pass_script.save()


apahim:

selftests/functional/test_basic.py
```diff
@@ -609,7 +610,7 @@ class RunnerSimpleTest(unittest.TestCase):
     def setUp(self):
         self.tmpdir = tempfile.mkdtemp(prefix='avocado_' + __name__)
         self.pass_script = script.TemporaryScript(
-            'avocado_pass.sh',
+            'ʊʋʉʈɑʅʛʌ',
```
clebergnu: This is a good assertion that we're handling unicode test references.  In the future we may even think about a fuzzer for this kind of input.

selftests/functional/test_basic.py
```diff
@@ -1,3 +1,4 @@
+# This Python file uses the following encoding: utf-8
```
clebergnu: I'm not sure this was intentional (probably was), but it matches the regex `^[ \t\v]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+)"` given on PEP 263:

https://www.python.org/dev/peps/pep-0263/

Just in case it wasn't intentional, and you weren't aware of PEP263, you may choose to use a more standard notations:

```
   # coding=<encoding name>
```

But it's entirely up to you.  This one does the job nicely.

avocado/core/loader.py
```diff
@@ -691,8 +690,7 @@ def ignore_broken(klass, uid, params=None):
                                                 subtests_filter)
             else:
                 if os.access(test_path, os.X_OK):
-                    return self._make_test(test.SimpleTest,
-                                           pipes.quote(test_path))
+                    return self._make_test(test.SimpleTest, test_path)
```
apahim: Actually I could not determine the use case that led to the inclusion of the `pipes.quote()`. But it's clearly causing a regression. I hope @ldoktor can help me to clarify.

selftests/functional/test_basic.py
apahim: This is following the PEP263. There, we read: 
```
...
More precisely, the first or second line must match the regular
expression "^[ \t\v]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+)".
...
```
And the line was extracted from the second example on the same PEP:
```
    2. Without interpreter line, using plain text:

          # This Python file uses the following encoding: utf-8
          import os, sys
          ...
```

The notation you pointed as 'more standard' is not described in PEP263 as so. But yes, it is a _de facto_ standard and I can use it to make you happy :)

apahim:
@clebergnu:

> 1) By commands you mean test references that become test names with additional (automatic shell) escaping?

By commands I mean test references that are SimpleTetsts and will be executed as a shell command. That's the case where the loader is using `pipes.quote()` as part of `_discover()`

> 2) We should try to pinpoint at which point a test name is not a filename anymore, and if we're using things such as os.path.abspath() this would clearly be (other) errors.

The `abspath()` in acting in SimpleTest, assuming the `Test.name` is in fact a file and probing the full path for the file, which is supposed to be legit. The only issue I see right now is if the `Test.name` comes within quotes, the case this PR tries to fix.

avocado/core/loader.py
ldoktor: OK, the system of execution probably changed since then. I played with your version and it works well. To be sure not to break things I'd suggest updating the simple check selftest with ` ` (space) in the script name and at least one argument and execute it with the argument (eg. `[ "$1" != "correct_arg" ] && exit -1`). That should increase the test coverage.

selftests/functional/test_basic.py
clebergnu: Oops, I missed that second example on the same PEP.  They're all OK, and the second example is more informative.

clebergnu:
@apahim looks like you can respin this with the small suggestions given by @adereis and @ldoktor .

PR #1590, opened 2016-11-02
[RFC] setup: dynamically probe the Avocado version

apahim:
This dirty version will rely on the git tags to create the version. Here a doc page created out off this commit (notice the page title):
http://apahim-avocado.readthedocs.io/en/fix_versioning/

And here a posterior commit with a tag version:
http://apahim-avocado.readthedocs.io/en/89.0/

@clebergnu, @ldoktor, @adereis, I'd love to read your guys feedback on this.


apahim:
[BUG: Avocado documentation versioning problems](https://trello.com/c/hEhau3Ee/862-bug-avocado-documentation-versioning-problems)

clebergnu:
I have the feeling that this won't work on a lot of scenarios, including installing from pip and from tarballs, which is used in RPM builds.  Have you tried those?

apahim:
@clebergnu yes... maybe we can use a combined version, that uses the static file as a fall-back?

PR #1593, opened 2016-11-03
docs: include link for the version documentation

apahim:
We have docs available on Red The Docs for all versions from 36.0
onwards. Let's include in each release notes the link for the version
documentation so users can easily see the docs for that version.


apahim:
[BUG: Avocado documentation versioning problems](https://trello.com/c/hEhau3Ee/862-bug-avocado-documentation-versioning-problems)

PR #1607, opened 2016-11-07
don't create a shell escaped command [v2]

apahim:
Escaping shell commands was introduced by commit fe13199485. It probably
did make sense at that time, but it is now causing more problems than
helping. Escaped commands can make os.path.abspath() to misbehave since
the command can be inside single quotes. Let's remove this so we can
behave correctly when some funky characters are in the command.

Changes from v1 (#1584):
- Add white-space in selftest check.
- Add an argument in the selftest check.
- Improve the commit message pointing to the commit that introduced the shell-escaping code.

### DIFF
--- a/avocado/core/loader.py
+++ b/avocado/core/loader.py
@@ -23,7 +23,6 @@ import imp
 import inspect
 import os
 import re
-import pipes
 import shlex
 import sys
 
@@ -694,8 +693,7 @@ class FileLoader(TestLoader):
                                                 subtests_filter)
             else:
                 if os.access(test_path, os.X_OK):
-                    return self._make_test(test.SimpleTest,
-                                           pipes.quote(test_path))
+                    return self._make_test(test.SimpleTest, test_path)
                 else:
                     return make_broken(test.NotATest, test_path)
         else:
--- a/selftests/functional/test_basic.py
+++ b/selftests/functional/test_basic.py
@@ -1,3 +1,4 @@
+# This Python file uses the following encoding: utf-8
 import aexpect
 import glob
 import json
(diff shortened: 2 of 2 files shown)


apahim:

ldoktor:
I still get some failures:

```
avocado run "ʊʋʉʈɑ ʅʛʌ"
JOB ID     : d82fe37bcec5675fe40700e81a47d8ac6b1a9fe7
JOB LOG    : /home/medic/avocado/job-results/job-2016-11-08T15.51-d82fe37/job.log
TESTS      : 1
 (1/1) ʊʋʉʈɑ ʅʛʌ:  Error running method "start_test" of plugin "journal": You must not use 8-bit bytestrings unless you use a text_factory that can interpret 8-bit bytestrings (like text_factory = str). It is highly recommended that you instead just switch your application to Unicode strings.
ERROR (0.03 s)
Error running method "end_test" of plugin "journal": You must not use 8-bit bytestrings unless you use a text_factory that can interpret 8-bit bytestrings (like text_factory = str). It is highly recommended that you instead just switch your application to Unicode strings.
RESULTS    : PASS 0 | ERROR 1 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0
TESTS TIME : 0.03 s
JOB HTML   : /home/medic/avocado/job-results/job-2016-11-08T15.51-d82fe37/html/results.html

```

but maybe it's just an old plugin. I'll play with it in train...

apahim:
@ldoktor you ave to run:

```
$ avocado run "ʊʋʉʈɑ\ ʅʛʌ"
```

The same way with anything containing a white-space in the name.

ldoktor:
Hello @apahim, as we discussed please include the fix for the journal plugin so the support is final.

PR #1614, opened 2016-11-09
 Fix unicode issues [v3]

apahim:
v3:
- Fix the test names unicode handling in journal plugin.

v2: #1607 
- Add white-space in selftest check.
- Add an argument in the selftest check.
- Improve the commit message pointing to the commit that introduced the shell-escaping code.

V1 #1584
- Remove simple tests pipes.quotes() command conversion.

### DIFF
--- a/avocado/core/loader.py
+++ b/avocado/core/loader.py
@@ -23,7 +23,6 @@ import imp
 import inspect
 import os
 import re
-import pipes
 import shlex
 import sys
 
@@ -694,8 +693,7 @@ class FileLoader(TestLoader):
                                                 subtests_filter)
             else:
                 if os.access(test_path, os.X_OK):
-                    return self._make_test(test.SimpleTest,
-                                           pipes.quote(test_path))
+                    return self._make_test(test.SimpleTest, test_path)
                 else:
                     return make_broken(test.NotATest, test_path)
         else:
--- a/avocado/plugins/journal.py
+++ b/avocado/plugins/journal.py
@@ -86,7 +86,7 @@ class JournalResult(ResultEvents):
             status = None
 
         self.journal_cursor.execute(sql,
-                                    (str(state['name']),
+                                    (repr(state['name']),
                                      datetime.datetime(1, 1, 1).now().isoformat(),
                                      action,
                                      status))
--- a/selftests/functional/test_basic.py
+++ b/selftests/functional/test_basic.py
@@ -1,3 +1,4 @@
+# This Python file uses the following encoding: utf-8
 import aexpect
 import glob
 import json
@@ -28,8 +29,9 @@ basedir = os.path.join(os.path.dirname(os.path.abspath(__file__)), '..', '..')
 basedir = os.path.abspath(basedir)
 
 
-PASS_SCRIPT_CONTENTS = """#!/bin/sh
-true
+PASS_SCRIPT_CONTENTS = """#!/bin/bash
+# Checks if the first argument is "arg1".
+[ "$1" == "arg1" ] && exit 0 || exit 1
 """
 
 PASS_SHELL_CONTENTS = "exit 0"
@@ -611,7 +613,7 @@ class RunnerSimpleTest(unittest.TestCase):
     def setUp(self):
         self.tmpdir = tempfile.mkdtemp(prefix='avocado_' + __name__)
         self.pass_script = script.TemporaryScript(
-            'avocado_pass.sh',
+            'ʊʋʉʈɑ ʅʛʌ',
             PASS_SCRIPT_CONTENTS,
             'avocado_simpletest_functional')
         self.pass_script.save()
@@ -623,8 +625,9 @@ class RunnerSimpleTest(unittest.TestCase):
 
     def test_simpletest_pass(self):
         os.chdir(basedir)
+        path = self.pass_script.path
         cmd_line = ('./scripts/avocado run --job-results-dir %s --sysinfo=off'
-                    ' %s' % (self.tmpdir, self.pass_script.path))
+                    ' "%s arg1"' % (self.tmpdir, path.replace(' ', '\ ')))
         result = process.run(cmd_line, ignore_status=True)
         expected_rc = exit_codes.AVOCADO_ALL_OK
         self.assertEqual(result.exit_status, expected_rc,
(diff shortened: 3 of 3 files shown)


apahim:
@clebergnu, could you please take a look if this makes sense for the journal plugin?

avocado/core/loader.py
```diff
@@ -694,8 +693,7 @@ def ignore_broken(klass, uid, params=None):
                                                 subtests_filter)
             else:
                 if os.access(test_path, os.X_OK):
-                    return self._make_test(test.SimpleTest,
-                                           pipes.quote(test_path))
+                    return self._make_test(test.SimpleTest, test_path)
```
ldoktor: OK this little guy is actually very important, because we do support 2 ways of specifying test-ids. One is a simple path, which means it does not require escaping spaces. Then, for simple tests, we also allow providing path&args, which means the spaces inside location have to be escaped, because we do the shlex.split (line 499 in loaders).

Anyway our `utils.process` library, in case shell=False, also does the `shlex.split` so if we don't escape the url in (1) it splits the single name.

Words are words, let's talk in practice:

```
touch "foo bar"
chmod +x "foo bar"
avocado run "foo\ bar baz"  => executes SimpleTest "foo bar" with argument "baz" (path 2)
avocado run "foo bar" => executes SimpleTest "foo bar" (path 1)
```
but with this patch:
```
avocado run "foo\ bar baz"  => executes SimpleTest "foo\ bar baz" (path 2), which executes ["foo bar", "baz]
avocado run "foo bar" => executes SimpleTest "foo bar" (path 1), which executes ["foo", "bar"]
```

One thing I don't understand is why do you remove the `pipes.quote`, when I return it back it works well, the selftest passes and I'm able to execute the tests properly.

ldoktor: I mean this:

```diff
diff --git a/avocado/core/loader.py b/avocado/core/loader.py
index 97dbf5e..1bc116d 100644
--- a/avocado/core/loader.py
+++ b/avocado/core/loader.py
@@ -12,6 +12,7 @@
 # Copyright: Red Hat Inc. 2014
 # Authors: Lucas Meneghel Rodrigues <lmr@redhat.com>
 #          Ruda Moura <rmoura@redhat.com>
+import pipes
 
 """
 Test loader module.
@@ -693,7 +694,8 @@ class FileLoader(TestLoader):
                                                 subtests_filter)
             else:
                 if os.access(test_path, os.X_OK):
-                    return self._make_test(test.SimpleTest, test_path)
+                    return self._make_test(test.SimpleTest,
+                                           pipes.quote(test_path))
                 else:
                     return make_broken(test.NotATest, test_path)

```

works well, passes the `make check` and I'm able to run both examples above with expected results.

ldoktor: Oh I see, the trello card uses absolute path and none of the solutions (before and with this patch) works...

apahim: @ldoktor, could you clarify why this:
`avocado run "foo bar" => executes SimpleTest "foo bar" (path 1)`
is a requirement?

In my understanding, if you run such test, `foo` is the filename and `bar` is it's argument. Unless you escape the whitespace.

apahim: > The "foo bar" is a requirement. It's actually the primary usage.
Where do we state so? I'm not convinced. And it seems wrong and hacky. It's the first time I'm thinking this through, but right now we are also inconsistent, because on `master`, this does not work:
```
avocado run '/tmp/test script.sh'
```
while this works:
```
cd /tmp
avocado run 'test script.sh'
```
And none of them have anything to do with unicode. But `pipes.quotes()`, which is making the second example to work, is also breaking tests with unicode characters. I think we should remove this `pipes.quote()` and handle the spaces in test names properly, after making clear what are the requirements (like in a RFC).

clebergnu: The current state of master is a similar to the old "remote copy" feature, that is, it has caveats, and even worse, undocumented ones.

So, while the current state of master allows for one use case, it is inconsistent with others. I suggest we take the path of simplicity, at the cost of one extra feature.

I suggest we drop altogether the support for passing parameters, so that every test reference, in the case of being resolved by simple tests, maps to a path (alone).  Example:

* `foo bar` is an executable
* `foo` is an executable
* `avocado run "foo bar"` *always run `foo bar`, never runs `foo` with argument `bar`

Unless there's a major use case supporting passing arguments to executables as a test reference, I suggest we go this route.

A more confusing and costlier to implement solution (that I don't recomend, but that would be still consistent), would be to support partial matches for the path, with a given set of precedence.  Example:

* `foo` is an executable
* `foo bar` *does not* exist
* `avocado run "foo bar"` breaks the test reference and considers: executable = `foo`, arguments = `bar`.
* `foo bar` is created as an executable
* `avocado run "foo bar"` finds exact match first, and considers: executable = `foo bar`, arguments = `` (empty).

This results in wasted brain cycles for our users, and maybe even a bunch of "WTFs" from them.  Creating a script with test parameters built in is so much easier compared to supporting this.  Also, we already support passing parameters (including mux ones) as environment variables which executables can use.

clebergnu:
2016-11-09 12:28 GMT-03:00 Lukáš Doktor <notifications@github.com>:

> *@ldoktor* commented on this pull request.
>
> In avocado/core/loader.py
> <https://github.com/avocado-framework/avocado/pull/1614>:
>
> > @@ -694,8 +693,7 @@ def ignore_broken(klass, uid, params=None):
>                                                  subtests_filter)
>              else:
>                  if os.access(test_path, os.X_OK):
> -                    return self._make_test(test.SimpleTest,
> -                                           pipes.quote(test_path))
> +                    return self._make_test(test.SimpleTest, test_path)
>
> The "foo bar" is a requirement. It's actually the primary usage. We don't
> expect users to quotate the test names, but only for SimpleTests we *also*
> support manual quotation as they might want to use extra arguments.
>
> Take "passtest.py" for example. When you rename it to "pass test.py", you
> still can execute it by running avocado run "pass test.py" or avocado run
> pass\ test.py. The use of double quotating is only available for
> SimpleTests...
>

This is because Avocado will get "pass test.py" in both cases.  Then, the
sentence "is only available for SimpleTests..." is a sign of what we should
preferrably refrain from having.

avocado/core/loader.py
ldoktor: Yes, it's not optimal, but very handy. Right now in my CI I use `external-runner`, so I can cope with that, but it'd be impossible to combine different kind of tests together. What am I talking about is:

```
avocado run "./run 1" "./run 2" ... "./run 999"
```

can be turned into

```
avocado run --external-runner ./run 1 2 ... 999
```

but if I wanted to run various kinds of tests in one job

```
avocado run boot "./run 1" "./run 2" ... "./run 999" other_suite.py
```

I would have to use SimpleTest with args.

Right now I use different jobs to run functional, qemu-iotests and qemu-kvm-unittests, but who knows, I'd like to have the possibility.

On the other hand yes, it's not nice, it actually requires running the discovery twice for each reference so maybe we could remove it. But first we need to be pretty sure nobody uses it (or can replace it). Alternative solution would be to implement the test-type prefix as we discussed several times and allow SimpleTestWithArgs only when it's specified:

```
avocado run VT://boot "SimpleArgs://foo --bar --baz" passtest.py
```

where without the prefix it uses autodetect (as it works now), with the prefix it only involves the specified loader. The `SimpleArgs` loader would be disabled for autodetection (similar to external-runner).

PR #1615, opened 2016-11-09
Preserve test environment for post-mortem debug

apahim:
In order to preserve test environment for post-mortem debug, let's
create the job tmp_dir inside the job logdir when keep_tmp_files is set
to True in the config file, so users can have the temporary directory
created for the job preserved in a stable location.

Reference: https://trello.com/c/Q6I65OZl


apahim:
Ok, agreed with both points. v2 coming soon.

PR #1623, opened 2016-11-21
Preserve test environment for post-mortem debug [v3]

apahim:
v1: #1615 

v2: #1620 
- New option `--keep-tmp (on|off)` for Avocado run command with precedence over the `keep_tmp_files` config option.

v3:
- Drop non-working config option `keep_tmp_files`.
- Decide whether we will keep the tmp dir or not based on the `basedir`. If it's not `None`, then we create the tmp_dir in the `basedir` location and don't remove it when the job finishes.

### DIFF
--- a/avocado/core/data_dir.py
+++ b/avocado/core/data_dir.py
@@ -25,6 +25,7 @@ The general reasoning to find paths is:
 * The next best location is the default system wide one.
 * The next best location is the default user specific one.
 """
+import logging
 import os
 import sys
 import shutil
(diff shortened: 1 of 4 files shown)


apahim:
[Preserve test environment for post-mortem debug](https://trello.com/c/Q6I65OZl/33-preserve-test-environment-for-post-mortem-debug)

avocado/core/data_dir.py
```diff
@@ -212,7 +223,7 @@ def __del__(self):
 _tmp_tracker = _TmpDirTracker()
 
 
-def get_tmp_dir():
+def get_tmp_dir(basedir=None, keep_tmp=None):
```
ldoktor: You forgot the `keep_tmp` here...

apahim: o>

ldoktor:
Actually as you already need to update the code, would you please improve the commit heading of the first patch? It should include the feature name...

apahim:
@ldoktor, updated.

ldoktor:
Thanks, works well...

PR #1624, opened 2016-11-21
fix unicode handling in journal plugin

apahim:
Sending this one-line fix while the PR that would include this is still under discussion.


apahim:
[BUG [regression]: journal plugin cannot handle unicode in test name](https://trello.com/c/SoFLGQVQ/871-bug-regression-journal-plugin-cannot-handle-unicode-in-test-name)

PR #1642, opened 2016-11-25
[wip] do not merge - Only commit 1

apahim:


apahim:
@ldoktor, no infinite recursion here with commit 1 only.

PR #1652, opened 2016-12-01
Fix issue "RuntimeError: maximum recursion depth exceeded" [V5] 

apahim:
v5:
- Create `LoggingFile.rm_handler()`.
- Adjust `_stop_logging()` to avoid any possibility of infinite recursion error.

v4: #1649 
- Adjust the `_stop_logging()` sequence.

v3:  #1647 
- Drop the commit with the protection to the `LoggingFile`
- Remove the handlers in `_stop_logging()`.

v2: #1644 

v1: #1631 

### DIFF
--- a/avocado/core/test.py
+++ b/avocado/core/test.py
@@ -30,6 +30,7 @@ from . import data_dir
 from . import exceptions
 from . import multiplexer
 from . import sysinfo
+from . import output
 from ..utils import asset
 from ..utils import astring
 from ..utils import data_structures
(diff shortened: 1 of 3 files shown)


apahim:

avocado/core/test.py
```diff
@@ -373,22 +375,34 @@ def _start_logging(self):
...
-                                        stream_formatter,
-                                        self._stdout_file)
-        self._register_log_file_handler(logging.getLogger("avocado.test.stderr"),
-                                        stream_formatter,
-                                        self._stderr_file)
-        self._ssh_fh = self._register_log_file_handler(logging.getLogger('paramiko'),
-                                                       formatter,
-                                                       self._ssh_logfile)
+        self._register_log_file_handler(
+                                      logging.getLogger("avocado.test.stdout"),
```
ldoktor: Hmm, I don't really like this style, nor I saw it before. I'm used to either using shorter variable names, assign the necessary variable to another variable, wrap it and start with one tab or keep the long line. But I never saw decreasing the length of one tab in order to fit in. Are you sure this was intended?

PR #1655, opened 2016-12-03
Support SUSE and openSUSE

coolo:
The package manager for both is zypper. Both have an /etc/os-release
file, but the content of it is quite different.

Some examples:

NAME="openSUSE Leap"
VERSION="42.2"
ID=opensuse

NAME="SLES"
VERSION="12-SP2"
VERSION_ID="12.2"

NAME="openSUSE Tumbleweed"
ID=opensuse
VERSION_ID="20161125"

So I took the easy path - what they all have in common is the SUSE
string. And as no other probe looks at /etc/os-release, this is very
safe for now.

### DIFF
--- a/avocado/utils/distro.py
+++ b/avocado/utils/distro.py
@@ -338,6 +338,48 @@ class DebianProbe(Probe):
     CHECK_FILE_DISTRO_NAME = 'debian'
 
 
+class SUSEProbe(Probe):
+
+    """
+    Simple probe for SUSE systems in general
+    """
+
+    CHECK_FILE = '/etc/os-release'
+    CHECK_FILE_CONTAINS = 'SUSE'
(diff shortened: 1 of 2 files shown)


coolo:
hmm, I spotted https://github.com/avocado-framework/avocado-misc-tests/blob/master/generic/gdb.py#L36 - so was there a SuSE once?

coolo:
it's quite puzzling actually. I setup a docker image of 42.2 with avocado (from git master) and it supports SuSE. But on my tumbleweed system not. This problem is inherited from platform.dist().

So I should mimic this - as I meanwhile found more tests that look for 'SuSE' (which is much outdated spelling of SUSE).

coolo:
extended the probe

coolo:
this is now working on tumbleweed and leap.

PR #1671, opened 2016-12-12
[RFC] tests temporary directory

apahim:
This docstring-less prototype is intended for discussion.

I don't think the job is the correct place to create such resource, since we can break the expectations when the multi-host is in place.

On the other hand, the test runner is always expected to be executed entirely in the same host. In this POC, I'm creating the temporary directory in the `run_suite()` and exposing it via both Test API and environment variable.

Btw, I first implemented it as a job pre/post plugin, but from the feedback on list, that approach can "bring inconsistencies or clashes in the very near future".

https://www.redhat.com/archives/avocado-devel/2016-October/msg00015.html

Once we agree in the architecture, the documentation will follow.

### DIFF
--- a/avocado/core/runner.py
+++ b/avocado/core/runner.py
@@ -31,6 +31,7 @@ from . import output
 from . import status
 from .loader import loader
 from .status import mapping
+from .teststmpdir import teststmpdir
 from ..utils import wait
 from ..utils import runtime
 from ..utils import process
@@ -513,11 +514,13 @@ class TestRunner(object):
         no_digits = len(str(test_result_total))
         self.result.tests_total = test_result_total
         self.result.start_tests()
+        tests_tmpdir = teststmpdir.create()
         index = -1
         try:
             for test_template in test_suite:
                 test_template[1]['base_logdir'] = self.job.logdir
                 test_template[1]['job'] = self.job
+                test_template[1]['tests_tmpdir'] = tests_tmpdir
                 break_loop = False
                 for test_factory, variant in self._iter_variants(test_template,
                                                                  mux):
@@ -556,6 +559,7 @@ class TestRunner(object):
         if self.job.sysinfo is not None:
             self.job.sysinfo.end_job_hook()
         self.result.end_tests()
+        teststmpdir.destroy()
         self.job._result_events_dispatcher.map_method('post_tests', self.job)
         self.job.funcatexit.run()
         signal.signal(signal.SIGTSTP, signal.SIG_IGN)
--- a/avocado/core/test.py
+++ b/avocado/core/test.py
@@ -129,7 +129,8 @@ class Test(unittest.TestCase):
     default_params = {}
 
     def __init__(self, methodName='test', name=None, params=None,
-                 base_logdir=None, job=None, runner_queue=None):
+                 base_logdir=None, job=None, runner_queue=None,
+                 tests_tmpdir=None):
         """
         Initializes the test.
 
@@ -145,6 +146,8 @@ class Test(unittest.TestCase):
                             provided, it'll use
                             :func:`avocado.data_dir.create_job_logs_dir`.
         :param job: The job that this test is part of.
+        :param tests_tmpdir: Temporary directory created to share resources
+                             among tests.
         :raises: :class:`avocado.core.test.NameNotTestNameError`
         """
         def record_and_warn(*args, **kwargs):
@@ -238,6 +241,10 @@ class Test(unittest.TestCase):
 
         self.time_elapsed = -1
         unittest.TestCase.__init__(self, methodName=methodName)
+
+        self.teststmpdir = None
+        if tests_tmpdir is not None:
+            self.teststmpdir = tests_tmpdir
 
     @property
     def basedir(self):
--- a/avocado/core/teststmpdir.py
+++ b/avocado/core/teststmpdir.py
@@ -1,0 +1,31 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; specifically version 2 of the License.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+
+
+import os
+import shutil
+import tempfile
+
+
+class TestsTmpDir(object):
+
+    def __init__(self, varname='XXX_TESTS_COMMON_TMPDIR'):
+        self._varname = varname
+
+    def create(self):
+        tmpdir = tempfile.mkdtemp(prefix='avocado_')
+        os.environ[self._varname] = tmpdir
+        return tmpdir
+
+    def destroy(self):
+        if os.environ.get(self._varname) is not None:
+            shutil.rmtree(os.environ.get(self._varname))
+
+teststmpdir = TestsTmpDir()


apahim:
`$ cat ~/avocado/tests/shared_dir.py `
```python
from avocado import Test
import os

class MyTest(Test):
    def setUp(self):
        print 'env', os.environ.get('XXX_TESTS_COMMON_TMPDIR')
        print 'api', self.teststmpdir

    def test1(self):
        pass
```

```
$ avocado --show test run ~/avocado/tests/shared_dir.py
...
START 1-/home/apahim/avocado/tests/shared_dir.py:MyTest.test1
env /var/tmp/avocado_dY8dms
api /var/tmp/avocado_dY8dms
PASS 1-/home/apahim/avocado/tests/shared_dir.py:MyTest.test1
...

```

apahim:
[Per-job tmpdir for direct use from tests](https://trello.com/c/P1JOtVY7/863-per-job-tmpdir-for-direct-use-from-tests)

avocado/core/test.py
```diff
@@ -129,7 +129,8 @@ class Test(unittest.TestCase):
     default_params = {}
 
     def __init__(self, methodName='test', name=None, params=None,
-                 base_logdir=None, job=None, runner_queue=None):
+                 base_logdir=None, job=None, runner_queue=None,
+                 tests_tmpdir=None):
```
ldoktor: Well I'd prefer not to change the test parameters as this would break all the custom test types and even user's custom test types (eg. avocado-vt, avocado-virt). We ought to add either `**kwargs` or custom args dictionary with extensible format for such changes in the next LTS... https://trello.com/c/9kz2ph6n/885-rework-avocado-test-init-arguments

Anyway for now I'd suggest using just one communication channel, therefor something like:

```
@property
def tests_tmpdir(self):
    return os.environ.get("XXX_TESTS_COMMON_TMPDIR")
```

avocado/core/test.py
```diff
@@ -239,6 +242,10 @@ def record_and_warn(*args, **kwargs):
         self.time_elapsed = -1
         unittest.TestCase.__init__(self, methodName=methodName)
 
+        self.teststmpdir = None
+        if tests_tmpdir is not None:
+            self.teststmpdir = tests_tmpdir
```
ldoktor: One question here. Do we expect people to check, whether `teststmpdir` is available or not? From my understanding of the usage I'd expect the tests which need it to depend on it blindly. If that's true it'd be better to raise exception when teststmpdir is not available (which could be implemented in the suggested property).

If we think many users wanted to write tests which treats non-existing `teststmpdir`, then we should keep returning None so they can check `self.workdir if self.teststmpdir is None else self.teststmpdir`. But I doubt many users would want to write such tests.

There could be also another "optimal" variant, which would assume that if `teststmpdir` is not provided by runner, `teststmpdir` is actually `self.workdir`. That would allow tests to always work, only sometimes without sharing the data. For some tests this would be the simplest solution for test writers, but I consider it magic and ugly.

avocado/core/teststmpdir.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+import os
+import shutil
+import tempfile
+
+
+class TestsTmpDir(object):
+
+    def __init__(self, varname='XXX_TESTS_COMMON_TMPDIR'):
```
ldoktor: I think `name` should be compulsory.

avocado/core/teststmpdir.py
```diff
@@ -0,0 +1,31 @@
...
+        self._varname = varname
+
+    def create(self):
```
ldoktor: Is there a reason to not create the file during `__init__`?

avocado/core/teststmpdir.py
```diff
@@ -0,0 +1,31 @@
...
+        tmpdir = tempfile.mkdtemp(prefix='avocado_')
+        os.environ[self._varname] = tmpdir
+        return tmpdir
+
+    def destroy(self):
+        if os.environ.get(self._varname) is not None:
```
ldoktor: Maybe I'm paranoid, but I'd prefer double-checking the value hasn't changed.

avocado/core/teststmpdir.py
```diff
@@ -0,0 +1,31 @@
...
+            shutil.rmtree(os.environ.get(self._varname))
+
+teststmpdir = TestsTmpDir()
```
ldoktor: I don't think we need generic `teststmpdir` always initialized on import. Each `runner` instance can create it's own instance, right?

avocado/core/runner.py
```diff
@@ -513,11 +514,13 @@ def run_suite(self, test_suite, mux, timeout=0, replay_map=None):
         no_digits = len(str(test_result_total))
         self.result.tests_total = test_result_total
         self.result.start_tests()
+        tests_tmpdir = teststmpdir.create()
```
ldoktor: Viz: https://github.com/avocado-framework/avocado/pull/1671/commits/b93c02b7d81c99a10526bba15f115ffa0d9836d1#r92342965 I don't think we should share `teststmpdir` across all runners in one avocado execution. I'd simply use:

```
from . import teststmpdir
tests_tmpdir = teststmpdir.teststmpdir("AVOCADO_TESTS_TMPDIR")
tests_tmpdir.create()
```

avocado/core/teststmpdir.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class TestsTmpDir(object):
+
+    def __init__(self, varname='XXX_TESTS_COMMON_TMPDIR'):
+        self._varname = varname
+
+    def create(self):
+        tmpdir = tempfile.mkdtemp(prefix='avocado_')
+        os.environ[self._varname] = tmpdir
```
ldoktor: Question, do we want to check whether the `self._varname` is not already in environment to avoid overriding previously set values? It could help identify issues in the future.

ldoktor:
Hello @apahim, thank you for the RFC, it's raising more questions than plain email. Generally I understand the point in having it implemented in `runner`, it's more likely to be consistent when we have support form multi-host tests, but I still believe we should start by using `pre/post` job hooks, because:

1. pre/post plugin is a plugin. It's easier to change/replace/turn into core a plugin, than a core feature
2. plugin is less intrusive, it does not require that many changes in core (this PR would break all custom runners and test types)
3. even if multi-host tests had changed the job vs. runner relation, the user perspective of `tmpdir` would stay the same. What I mean is that before `multi-host` there would be no difference (test-writer-wise) between pre/post or runner implementation and after the `mulit-host` we'd have to probably move to `runner` implementation. This question is still hypothetical, because it depends on the `multi-host` implementation so maybe pre/post would suffice, maybe it'd not. On the other hand turning a plugin into core feature is simple, but removing or significantly changing core feature is hard.

For those reasons I think we should first implement it as pre/post plugin, because this interface is available to us and from test-writer perspective there is no difference between those two implementations. Later, when multi-host is implemented (or rather when it's drafted) we should double-check whether it's implementation breaks the concept, or not, because it's possible that the implementation would not break anything. But trying to prepare for something we don't even have a draft for seems invalid to me (given the consequences where even this version could be incompatible with multi-host and then we'd have to break core feature, rather than adjust a plugin).

Apart from this I mentioned several little issues and some conceptual questions in code.

avocado/core/teststmpdir.py
ldoktor: Also if we are to make some default, I'd chose `AVOCAOD_` prefix as we already do create some `AVOCADO_` prefixed values.

avocado/core/test.py
apahim: Ok

avocado/core/test.py
apahim: Ok, makes sense. I will raise an exception here.

avocado/core/teststmpdir.py
```diff
@@ -0,0 +1,31 @@
...
+#
+# See LICENSE for more details.
+
+
+import os
+import shutil
+import tempfile
+
+
+class TestsTmpDir(object):
```
apahim: If we are going to use pre/post plugin, no need to have additional code outside the plugin.

avocado/core/teststmpdir.py
apahim: This was discussed in the mail thread. I started with `AVOCADO_` and I was told to not use it.

avocado/core/teststmpdir.py
apahim: Nope

avocado/core/teststmpdir.py
apahim: ok

avocado/core/teststmpdir.py
apahim: No, no need. As we are changing to pre/post plugin, it will handle this.

avocado/core/teststmpdir.py
apahim: ok

PR #1692, opened 2017-01-11
Remove argument support for SIMPLE tests

apahim:
To provide support for passing arguments for simple tests in command
line, we were creating some cases where the test reference could not be
properly handled. Also, for test references with white-spaces, we have
an inconsistent behavior. See reference.

This patch removes the support for arguments in test references,
restoring the proper behavior, which is expected to be shell-like.

As consequence, we now handle correctly white-spaces in tests references
as well as Unicode characters.

Reference: https://www.redhat.com/archives/avocado-devel/2016-November/msg00011.html
Reference: https://trello.com/c/bfE9NBbl

### DIFF
--- a/avocado/core/test.py
+++ b/avocado/core/test.py
@@ -21,6 +21,7 @@ framework tests.
 import inspect
 import logging
 import os
+import pipes
 import re
 import shutil
 import sys
@@ -702,7 +703,9 @@ class SimpleTest(Test):
     def __init__(self, name, params=None, base_logdir=None, job=None):
         super(SimpleTest, self).__init__(name=name, params=params,
                                          base_logdir=base_logdir, job=job)
-        self._command = self.filename
+        self._command = None
+        if self.filename is not None:
+            self._command = pipes.quote(self.filename)
 
     @property
     def filename(self):
--- a/selftests/functional/test_basic.py
+++ b/selftests/functional/test_basic.py
@@ -1,3 +1,4 @@
+# This Python file uses the following encoding: utf-8
 import aexpect
 import glob
 import json
@@ -86,6 +87,11 @@ class MyTest(Test):
     def test(self):
          self.runner_queue.put({"running": False})
          time.sleep(60)
+'''
+
+SLEEP_60 = '''#!/bin/sh
+
+sleep 60
 '''
 
 DIE_WITHOUT_REPORTING_STATUS = '''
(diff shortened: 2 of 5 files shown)


apahim:

avocado/core/test.py
```diff
@@ -702,7 +703,9 @@ class SimpleTest(Test):
     def __init__(self, name, params=None, base_logdir=None, job=None):
         super(SimpleTest, self).__init__(name=name, params=params,
                                          base_logdir=base_logdir, job=job)
-        self._command = self.filename
+        self._command = None
+        if self.filename is not None:
+            self._command = pipes.quote(self.filename)
```
ldoktor: Yep, this looks quite good, but there is one uncovered aspect and that is external runner with spaces. Currently it supports arguments, but in an odd way so it always fails (when using spaces in binary name). I'd suggest this version:

```diff
diff --git a/avocado/core/loader.py b/avocado/core/loader.py
index c380aa9..5e26237 100644
--- a/avocado/core/loader.py
+++ b/avocado/core/loader.py
@@ -790,7 +790,7 @@ class ExternalLoader(TestLoader):
             if len(external_runner_and_args) > 1:
                 executable = external_runner_and_args[0]
             else:
-                executable = runner
+                executable = external_runner_and_args[0]
             if not os.path.exists(executable):
                 msg = ('Could not find the external runner executable "%s"'
                        % executable)
```

Which unlike normal test name allows arguments. Alternative acceptable solution would be to remove the `shlex.split` and only accept path to the binary (similarly to test-reference). Also please add a selftest for this example.

^^ I tested this by using echo binary copied to `/tmp/a b`, the first variant allows one to run:

```
avocado --show all run --external-runner "/tmp/a\ b aaa" asdf
```

The later would only allow the usage of:

```
avocado --show all run --external-runner "/tmp/a b" asdf
```
while specification of the path would be friendlier. I don't have strong opinion about which version to use. I'm slightly inclined towards the first way (that's why I mentioned it as first) because it allows to group shared arguments to the external runner ("./run -v -s -r"). The second way is on the other hand easier to follow as it's the same behavior as test reference. The reason for me preferring the first approach is that I see external-runner as advanced feature and advanced features should be powerful.

selftests/functional/test_basic.py
```diff
@@ -713,10 +697,9 @@ def test_non_absolute_path(self):
 
     @unittest.skipIf(not SLEEP_BINARY, 'sleep binary not available')
     def test_kill_stopped_sleep(self):
-        sleep = "'%s 60'" % SLEEP_BINARY
         proc = aexpect.Expect("./scripts/avocado run %s --job-results-dir %s "
                               "--sysinfo=off --job-timeout 3"
-                              % (sleep, self.tmpdir))
+                              % (self.sleep_60, self.tmpdir))
```
ldoktor: Btw instead of script you could use external runner here 😉

selftests/functional/test_basic.py
```diff
@@ -1137,18 +1121,6 @@ def test_json_plugin_errortest(self):
         self.run_and_check('errortest.py', exit_codes.AVOCADO_TESTS_FAIL,
                            1, 1, 0, 0)
 
-    @unittest.skipIf(not ECHO_BINARY, 'echo binary not available')
-    def test_ugly_echo_cmd(self):
-        data = self.run_and_check('"/bin/echo -ne foo\\\\\\n\\\'\\\\\\"\\\\\\'
```
ldoktor: I'd prefer keeping this test, because IIRC it was based on Avocado not being able to even execute this command. Please use `--external-runner /bin/echo '-ne foo\\\\\\n\\\'\\\\\\"\\\\\\...` to run this test.

avocado/core/test.py
apahim: Yes, let's keep at least the external runner with args. As you said, advanced features should be powerful. We can always re-evaluate that if it becomes a problem.

selftests/functional/test_basic.py
apahim: Sure, np.

selftests/functional/test_basic.py
apahim: Ok

PR #1693, opened 2017-01-11
Remove argument support for SIMPLE tests [V2]

apahim:
v2:
- Keep selftests `test_ugly_echo_cmd`.
- Allow `external-runner` to use arguments.
- Instead of script, use external runner in test `test_kill_stopped_sleep`.

v1: #1692

### DIFF
--- a/avocado/core/loader.py
+++ b/avocado/core/loader.py
@@ -23,7 +23,6 @@ import imp
 import inspect
 import os
 import re
-import pipes
 import shlex
 import sys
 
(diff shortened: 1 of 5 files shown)


avocado/core/loader.py
```diff
@@ -794,10 +787,7 @@ def _process_external_runner(args, runner):
 
         if runner:
             external_runner_and_args = shlex.split(runner)
-            if len(external_runner_and_args) > 1:
-                executable = external_runner_and_args[0]
-            else:
-                executable = runner
+            executable = external_runner_and_args[0]
```
ldoktor: Thanks, so it's the more powerful way for advanced users 😄

apahim:

PR #1697, opened 2017-01-12

ggopinat:
newline if needed.

ignore_bg_processes

### DIFF
--- a/avocado/utils/process.py
+++ b/avocado/utils/process.py
@@ -1097,57 +1097,9 @@ def run(cmd, timeout=None, verbose=True, ignore_status=False,
     return cmd_result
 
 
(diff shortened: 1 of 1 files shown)


ggopinat:
[Shouldn't `process.system_output` strip the last line?](https://trello.com/c/jpESQses/785-shouldn-t-process-system-output-strip-the-last-line)

clebergnu:
@ggopinat as you probably noticed by now, you deleted the `system()` function, probably unintentionally.

ggopinat:
Yep. I corrected that but I still seem to be running into a 
signed-off check fail.

I am not sure what the error is, since my PR was on the latest master.

----- Original Message -----

PR #1704, opened 2017-01-15

ggopinat:
newline if needed.

ignore_bg_processes

### DIFF
--- a/avocado/utils/process.py
+++ b/avocado/utils/process.py
@@ -1147,7 +1147,7 @@ def system(cmd, timeout=None, verbose=True, ignore_status=False,
 
 def system_output(cmd, timeout=None, verbose=True, ignore_status=False,
                   allow_output_check='all', shell=False, env=None, sudo=False,
-                  ignore_bg_processes=False):
+                  ignore_bg_processes=False, strip_trail_nl=True):
     """
     Run a subprocess, returning its output.
 
@@ -1181,6 +1181,11 @@ def system_output(cmd, timeout=None, verbose=True, ignore_status=False,
                  has a sudo configuration such that a password won't be
                  prompted. If that's not the case, the command will
                  straight out fail.
+    :type sudo: bool
+    :param ignore_bg_processes: Whether to ignore background processes
+    :type ignore_bg_processes: bool
+    :param strip_trail_nl: Whether to strip the trailing newline
+    :type strip_trail_nl: bool
 
     :return: Command output.
     :rtype: str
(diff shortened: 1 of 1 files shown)


avocado/utils/process.py
```diff
@@ -1189,4 +1194,6 @@ def system_output(cmd, timeout=None, verbose=True, ignore_status=False,
     cmd_result = run(cmd=cmd, timeout=timeout, verbose=verbose, ignore_status=ignore_status,
                      allow_output_check=allow_output_check, shell=shell, env=env,
                      sudo=sudo, ignore_bg_processes=ignore_bg_processes)
+    if strip_trail_nl:
```
clebergnu: This code is consistent with the commit description.  The original card, though, has the intention to actually change the current behavior and align it with the other similar utility functions given.

I suggest you change this parameter to something like `dont_strip_nl` with the opposite behavior.  IMO, I'd even consider such a parameter to be a candidate for removal in a future release, say, before the next LTS release.

avocado/utils/process.py
```diff
@@ -1181,6 +1181,11 @@ def system_output(cmd, timeout=None, verbose=True, ignore_status=False,
                  has a sudo configuration such that a password won't be
                  prompted. If that's not the case, the command will
                  straight out fail.
+    :type sudo: bool
```
clebergnu: Thanks for adding these missing docstrings!  While you're at it, a couple of hints and requests:
 * If possible, also add the `param sudo` docstring
 * Ideally, as a good practice, this docstring should be in a separate commit.  To check if it's better to have a given piece of code in a separate commit or not, just ask: "does this hunk contributes to what's described on the commit message?"

avocado/utils/process.py
clebergnu: @ggopinat my mistake, I did not notice that `strip_trail_nl` is set to `True` by default.

clebergnu: @ggopinat I'm going to get back on this after the upcoming release.

clebergnu:
Since the code delivers what the original card requested, and the added docstring is a bonus (and separating is a good practice, but nothing major), I'm approving this PR.

clebergnu:
[Shouldn't `process.system_output` strip the last line?](https://trello.com/c/jpESQses/785-shouldn-t-process-system-output-strip-the-last-line)

clebergnu:
[Shouldn't `process.system_output` strip the last line?](https://trello.com/c/jpESQses/785-shouldn-t-process-system-output-strip-the-last-line)

PR #1710, opened 2017-01-19
avocado.core.runner: add test directory to sys.path

apahim:
available in sys.path.

the test module when needed.


apahim:

PR #1713, opened 2017-01-20
avocado.core.runner: add test directory to sys.path

apahim:
available in sys.path.

the test module when needed.

### DIFF
--- a/avocado/core/runner.py
+++ b/avocado/core/runner.py
@@ -339,6 +339,10 @@ class TestRunner(object):
         """
         proc = None
         sigtstp = multiprocessing.Lock()
+        if 'modulePath' in test_factory[1]:
+            test_path = test_factory[1]['modulePath']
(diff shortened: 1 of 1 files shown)


lmr:
I have a workaround for that on `scylla-cluster-tests` - I ask users to run `python setup.py develop` beforehand. This would be handy to be ported to 36.X.

apahim:
@lmr in that case, could you verify this pr? :)

lmr:
@apahim `scylla-cluster-tests` is compatible with the 36.X branch - I really didn't have enough time to keep up with the API changes in avocado during the last year. I can write a functional test for this Pull Request, let me see...

apahim:
Oh, ok, no worries. Just if it was handy to verify. Tks anyway,

lmr:
Closing in favor of #1717.

PR #1717, opened 2017-01-24
[V2] avocado.core.runner: add test directory to sys.path

lmr:
available in sys.path.

the test module when needed.

Changelog:
V1 - V2: Add a functional test

### DIFF
--- a/avocado/core/runner.py
+++ b/avocado/core/runner.py
@@ -339,6 +339,10 @@ class TestRunner(object):
         """
         proc = None
         sigtstp = multiprocessing.Lock()
+        if 'modulePath' in test_factory[1]:
+            test_path = test_factory[1]['modulePath']
+            test_module_dir = os.path.abspath(os.path.dirname(test_path))
+            sys.path.insert(0, test_module_dir)
 
         def sigtstp_handler(signum, frame):     # pylint: disable=W0613
             """ SIGSTOP all test processes on SIGTSTP """
--- a/selftests/functional/test_runner_queue.py
+++ b/selftests/functional/test_runner_queue.py
@@ -1,0 +1,75 @@
+import os
+import sys
+import shutil
+import tempfile
+
+if sys.version_info[:2] == (2, 6):
+    import unittest2 as unittest
+else:
+    import unittest
+
+from avocado.core import exit_codes
+from avocado.utils import script
+from avocado.utils import process
+
+basedir = os.path.join(os.path.dirname(os.path.abspath(__file__)), '..', '..')
+basedir = os.path.abspath(basedir)
+
+AVOCADO_EXTERNAL_LIB = """
+def cleanup_function(arg1):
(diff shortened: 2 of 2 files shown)


selftests/functional/test_runner_queue.py
```diff
@@ -0,0 +1,67 @@
...
+                    '--sysinfo=off',
+                    '--job-results-dir',
+                    "%s" % self.tmpdir,
+                    "%s" % mytest]
+        result = process.run(' '.join(cmd_line))
+        self.assertEqual(result.exit_status, exit_codes.AVOCADO_ALL_OK)
+        self.assertIn('Bogus cleanup function. arg1: Hello Avocado Test Queue!',
+                      result.stdout,
+                      'Cleanup function message not found in stdout:\n%s' %
+                      result.stdout)
```
apahim: Missing `teadDown()` with cleanup.

lmr: Sure, my bad. I'm running some errands, so I'll update the PR later today.

lmr: Solved, now the functional test has a `tearDown()` method with cleanup actions.

apahim: @lmr do you mind including?
```python
if __name__ == '__main__':
    unittest.main()
```

lmr: Just updated the unittest with that snippet. Sorry about my forgetfulness.

apahim: Thank you. Let's wait someone else to review now, since this is a combined work.

apahim:
@clebergnu, @ldoktor, could you take a look here?

avocado/core/runner.py
```diff
@@ -339,6 +339,10 @@ def run_test(self, test_factory, queue, summary, job_deadline=0):
         """
         proc = None
         sigtstp = multiprocessing.Lock()
+        if 'modulePath' in test_factory[1]:
+            test_path = test_factory[1]['modulePath']
+            test_module_dir = os.path.abspath(os.path.dirname(test_path))
+            sys.path.insert(0, test_module_dir)
```
ldoktor: Actually I'm quite afraid this potentially could break the main Avocado process. I'd prefer if we just store the modulePath along with the pickled object in `func_at_exit` (if provided by the test) and we'd only insert it during the `func_at_exit` and remove after we executed the code.

Imagine that the test adds a path and later in code the main thread tries to import something but receives the test's module instead of the system module.

lmr: Naming the test modules after system libraries would be a bad idea in the first place.

Now there is merit to wanting to isolate the test process from the main avocado process, I agree with the concern from that perspective. I will take a look at this later then.

apahim:
@lmr, I'm working on this.

apahim:

PR #1718, opened 2017-01-24
[RFC] Test: add skipIf and skipUnless decorators

apahim:
Avocado does not support self.skip() to be called from inside a test.
This is a design decision that users have to comply with. On the other
hand, users keep requesting ways to skip tests.

This patch creates the decorators Test.skipIf and Test.skipUnless to be
used in test functions. Corresponding exceptions will be raised in those
cases.

Reference: https://trello.com/c/JK5Z5dql

### DIFF
--- a/avocado/core/test.py
+++ b/avocado/core/test.py
@@ -478,6 +478,9 @@ class Test(unittest.TestCase):
             raise exceptions.TestSetupFail(details)
         try:
             testMethod()
+        except exceptions.TestSkipIfError as details:
+            stacktrace.log_exc_info(sys.exc_info(), logger='avocado.test')
+            raise exceptions.TestSkipError(details)
         except exceptions.TestSkipError as details:
             stacktrace.log_exc_info(sys.exc_info(), logger='avocado.test')
             skip_illegal_msg = ('Calling skip() in places other than '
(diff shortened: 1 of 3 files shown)


avocado/core/test.py
```diff
@@ -691,6 +694,30 @@ def fetch_asset(self, name, asset_hash=None, algorithm='sha1',
         return asset.Asset(name, asset_hash, algorithm, locations,
                            self.cache_dirs, expire).fetch()
 
+    @staticmethod
+    def skipIf(condition, reason):
```
lmr: My concern here is placement. In the `unittest` module, the `skipIf` decorator is part of the main `unittest` namespace:

    from unittest import skipIf

Also, structurally, the decorator doesn't seem to belong to the `Test` class. I'd suggest that those are moved to some library, and make it also directly available under `avocado.skipIf`.

selftests/functional/test_skiptests.py
```diff
@@ -0,0 +1,64 @@
...
+class AvocadoSkipTests(Test):
+
+    @Test.skipIf(True, 'Skipped due to the True condition')
+    def test1(self):
+        pass
+
+    @Test.skipUnless(False, 'Skipped due to the False condition')
+    def test2(self):
+        pass
+"""
```
lmr: Maybe this test could also include conditions that are evaluated using user written libraries in the same directory as the test. This seems to be always a tricky subject, given the way we handle module imports.

avocado/core/test.py
apahim: Yes, I was not sure about this location. Problem is that a common avocado usage starts with `from avocado import Test`. While in unittest world, the common usage is `import unittest`. Using `import avocado` is less common, so users will need to add the `from avocado import skipIf` or stick to `import avocado`. Anyway, you're right. Let's not attach this to the Test class itself. Thank you.

selftests/functional/test_skiptests.py
apahim: Good point.

PR #1728, opened 2017-02-01
Improve yaml_to_mux error messages

mengalong:
When we use args:--mux-yaml and the yaml file not existing it
show more detail error messages

Reference: https://trello.com/c/hYCrS1oI/900-improve-yaml-to-mux-error-messages
Signed-off-by: mengalong <alongmeng@gmail.com>

### DIFF
--- a/avocado/plugins/yaml_to_mux.py
+++ b/avocado/plugins/yaml_to_mux.py
@@ -310,7 +310,7 @@ class YamlToMux(CLI):
             try:
                 data.merge(create_from_yaml(multiplex_files, debug))
             except IOError as details:
-                logging.getLogger("avocado.app").error(details.strerror)
+                logging.getLogger("avocado.app").error(details)
                 sys.exit(exit_codes.AVOCADO_JOB_FAIL)
 
         # Deprecated --multiplex option


avocado/plugins/yaml_to_mux.py
```diff
@@ -310,7 +310,7 @@ def run(self, args):
             try:
                 data.merge(create_from_yaml(multiplex_files, debug))
             except IOError as details:
-                logging.getLogger("avocado.app").error(details.strerror)
+                logging.getLogger("avocado.app").error(details)
```
apahim: You're exposing internal information:
```
[Errno 2] No such file or directory: 'mux.yaml'
```
This `[Errno 2]` is the internal exception error code. No need for the user to receive that information.

avocado/plugins/yaml_to_mux.py
```diff
@@ -310,7 +310,7 @@ def run(self, args):
...
                 sys.exit(exit_codes.AVOCADO_JOB_FAIL)
 
         # Deprecated --multiplex option
```
apahim: This deprecated option should be coved by this fix as well:
```
$ avocado run passtest.py --multiplex mux.yaml
The use of '--multiplex' is deprecated, please use '--mux-yaml' instead
No such file or directory
```

mengalong: Thanks for your advice, let me fix it

PR #1741, opened 2017-02-03
Add skip, skipIf and skipUnless decorators [V2]

apahim:
v2:
- Use `avocado.utils.script` in selftests.
- Include doccumentation.
- Better naming and organization for skip exceptions.

v1: #1725 
- Improve selftests.
- Move decorators from `avocado.Test` to `avocado`.

RFC: #1718

### DIFF
--- a/avocado/core/decorators.py
+++ b/avocado/core/decorators.py
@@ -1,0 +1,53 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: Red Hat Inc. 2017
+# Author: Amador Pahim <apahim@redhat.com>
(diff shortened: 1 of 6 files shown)


clebergnu:
[Add a decorator to skip tests, similar to unittest.skip*](https://trello.com/c/JK5Z5dql/526-add-a-decorator-to-skip-tests-similar-to-unittest-skip)

avocado/core/decorators.py
```diff
@@ -0,0 +1,53 @@
...
+
+def skip(message=None):
+    """
+    Decorator to skip a test.
+    """
+    def decorator(function):
+        if not isinstance(function, type):
+            @wraps(function)
+            def wrapper(*args, **kwargs):
+                raise exceptions.TestDecoratorSkip(message)
```
ldoktor: I'm wondering what is the benefit of strictly separating unittest and our exceptions. If we just say that `unittest.SkipTest` is allowed from everywhere including the `test` all `unittest.skip*` methods would work, people would not have to use new APIs while still the `self.skip` and such would produce test errors.

I mean when someone looks for our exceptions he can still use `raise avocado.core.exceptions.TestDecoratorSkip` to skip inside the test's body and there will be people using it. Why not to just allow to use `unittest.SkipTest` and just say that people should not use it as it's not a good practice and keep failing in case `self.skip` is used. I think that would be sufficient and the usage would be way easier (and we would not have to implement any of these `skip` decorators as they already exists)

PR #1750, opened 2017-02-06
Adds warning to UI & job.log if remote avocado version mismatch

msaldivar:
For debugging purposes it's nice to know if the remote version is
different from the local version of avocado. Code has been added to
job.py and remote/runner.py to support this functionality.
remote/runner.py will gather the remote and local versions, and print a
warning to the UI if they are different. remote/runner.py also adds the
version variables to the summary set. job.py checks the length of the
summary set variable if it equals exactly two, it'll compare the
versions and add a warning to job.log. The versions are popped from
summary returning summary to it's previous state, and code execution
continues as before.

### DIFF
--- a/avocado/core/job.py
+++ b/avocado/core/job.py
@@ -467,6 +467,19 @@ class Job(object):
                                              variant,
                                              self.timeout,
                                              replay_map)
+        # if the length of summary is 2, check if it came from remote/runner.py
+        # the summary set holds both versions of avocado local and remote
+        # if the values are not from remote/runner.py add them back to the set and continue code execution
+        if len(summary) == 2:
+            local_version = summary.pop()
+            remote_version = summary.pop()
+            if (local_version.isdigit() and remote_version.isdigit()) and (remote_version != local_version):
+                _TEST_LOGGER.warn("WARNING: avocado version on the remote system is %s, local is %s",
+                                  remote_version, local_version)
+            else:
+                summary.add(local_version)
+                summary.add(remote_version)
+
         # If it's all good so far, set job status to 'PASS'
         if self.status == 'RUNNING':
             self.status = 'PASS'
--- a/avocado/core/remote/runner.py
+++ b/avocado/core/remote/runner.py
@@ -22,6 +22,7 @@ import logging
 from fabric.exceptions import CommandTimeout
 
 from .test import RemoteTest
+from .. import version
 from .. import output
 from .. import remoter
 from .. import virt
@@ -47,6 +48,8 @@ class RemoteTestRunner(TestRunner):
         super(RemoteTestRunner, self).__init__(job, result)
         #: remoter connection to the remote machine
         self.remote = None
+        self.remote_version = ""
+        self.local_version = ""
 
     def setup(self):
         """ Setup remote environment and copy test directories """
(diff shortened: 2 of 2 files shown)


avocado/core/remote/runner.py
```diff
@@ -270,6 +279,9 @@ def run_suite(self, test_suite, variants, timeout=0, replay_map=None):
                 raise exceptions.JobError(details)
             sys.stdout = stdout_backup
             sys.stderr = stderr_backup
+        # returning the remote and local avocado version to be added to job.log as a possible warning
+        summary.add(self.remote_version.split('.')[0])
+        summary.add(self.local_version.split('.')[0])
```
apahim: Let's not abuse `summary`. It's intended to contain types of test failures. Also, no need for the job to know about the version mismatch. At this point you can send the warning directly to the log.

avocado/core/remote/runner.py
```diff
@@ -86,7 +89,13 @@ def check_remote_avocado(self):
         if result.exit_status == 127:
             return (False, None)
 
+        # Get the remote and local versions of avocado, and compare the major versions only
```
apahim: `check_remote_avocado()` already returns the remote avocado version. You can keep it as is and handle the comparison in `run_suite()`.

avocado/core/remote/runner.py
```diff
@@ -86,7 +89,13 @@ def check_remote_avocado(self):
...
         match = self.remote_version_re.findall(result.stdout)
+        self.remote_version = '.'.join(list(match[0]))
+        self.local_version = version.VERSION
+        if self.remote_version.split('.')[0] != self.local_version.split('.')[0]:
```
apahim: Why to join `self.remote_version` to then split? If you join it only for the log message (after this `if`) you don't need to split.

avocado/core/job.py
```diff
@@ -467,6 +467,19 @@ def run_tests(self):
...
+        if len(summary) == 2:
+            local_version = summary.pop()
+            remote_version = summary.pop()
+            if (local_version.isdigit() and remote_version.isdigit()) and (remote_version != local_version):
+                _TEST_LOGGER.warn("WARNING: avocado version on the remote system is %s, local is %s",
+                                  remote_version, local_version)
+            else:
+                summary.add(local_version)
+                summary.add(remote_version)
+
```
apahim: No need for the job to know about this. You can handle everything in the runner.

apahim:
Also, a rebase should be enough to make travis happy.

avocado/core/remote/runner.py
msaldivar: Hi @apahim  My logic to do the join is to make sure it's in the same format as what  `version.VERSION` returns. Since `match` is returned as `(36, 2)`, I do a join so it's in a string format, and split both strings to only compare the major versions

avocado/core/job.py
msaldivar: @apahim I was having trouble with the logger in `runner` to write to job.log, that's why I originally sent the information to `job`. I'll refactor the code and test again. Thanks for the input!

avocado/core/remote/runner.py
apahim: `if str(match[0]) != version.VERSION.split('.')[0]` might do the trick.

avocado/core/job.py
apahim: Problem is that job.log is overwritten by `archive.uncompress` with the remote job.log. So you will only be able to write in it after that.

avocado/core/remote/runner.py
apahim: Or, better: `if str(match[0]) != version.MAJOR`

PR #1752, opened 2017-02-06
Adds warning to UI & job.log if remote avocado version mismatch

msaldivar:
Refactored code, and kept everything inside remote/runner.py

Previous pull request #1750 

Reference: https://trello.com/c/ZcV36ncd

Signed-off-by: Maurice Saldivar <maurice.a.saldivar@hpe.com>

PR #1753, opened 2017-02-06

msaldivar:
Fixed indent error that travis-ci picked up. The error was produced
by switching between vim and pycharm. Code has been refactored, and
all logic is self contained inside remote/runner.py. The warning appears in both
the UI and job.log

Previous pull requests: #1750 & #1752 

### DIFF
--- a/avocado/core/remote/runner.py
+++ b/avocado/core/remote/runner.py
@@ -22,6 +22,7 @@ import logging
 from fabric.exceptions import CommandTimeout
 
 from .test import RemoteTest
+from .. import version
 from .. import output
 from .. import remoter
 from .. import virt
@@ -217,7 +218,7 @@ class RemoteTestRunner(TestRunner):
         try:
             try:
                 self.setup()
-                avocado_installed, _ = self.check_remote_avocado()
+                avocado_installed, remote_version = self.check_remote_avocado()
                 if not avocado_installed:
                     raise exceptions.JobError('Remote machine does not seem to'
                                               ' have avocado installed')
@@ -264,6 +265,13 @@ class RemoteTestRunner(TestRunner):
             self.result.end_tests()
             self.job._result_events_dispatcher.map_method('post_tests',
                                                           self.job)
+            # compare the major version of remote and local version of avocado. If different write to UI and job.log
+            if str(remote_version[0]) != version.MAJOR:
+                log = logging.getLogger("avocado.job")
+                self.job.log.info("WARNING    : avocado version on the remote system is %s, local is %s",
+                                  '.'.join(map(str, remote_version)), version.VERSION)
+                log.warn("WARNING    : avocado version on the remote system is %s, local is %s",
+                         '.'.join(map(str, remote_version)), version.VERSION)
         finally:
             try:
                 self.tear_down()


avocado/core/remote/runner.py
```diff
@@ -264,6 +265,13 @@ def run_suite(self, test_suite, variants, timeout=0, replay_map=None):
             self.result.end_tests()
             self.job._result_events_dispatcher.map_method('post_tests',
                                                           self.job)
+            # compare the major version of remote and local version of avocado. If different write to UI and job.log
+            if str(remote_version[0]) != version.MAJOR:
+                log = logging.getLogger("avocado.job")
+                self.job.log.info("WARNING    : avocado version on the remote system is %s, local is %s",
```
apahim: Btw, using `APP_LOG.warn()`, you don't need the string `WARNING :`. The yellow color is enough to say it's a warning. So you can make a better use of that space. I'd use a message like `VERSIONS: Local (%s) and remote (%s) are not the same.`

avocado/core/remote/runner.py
```diff
@@ -264,6 +265,13 @@ def run_suite(self, test_suite, variants, timeout=0, replay_map=None):
...
+                                  '.'.join(map(str, remote_version)), version.VERSION)
+                log.warn("WARNING    : avocado version on the remote system is %s, local is %s",
+                         '.'.join(map(str, remote_version)), version.VERSION)
```
apahim: Last, something to think about: what is the best moment to show the message in the UI? After or before the tests? The way it is now, the message shows up after the tests, so you have to wait for the tests to finish to then see the message. Think in a long test (like `examples/tests/sleeptenmin.py`). Would it be good to have the information about version mismatch before the tests? I see some value in having the message in advance. In that case, the `APP_LOG.warn()` should be called before the line:
```python
            ...
            results = self.run_test(self.job.references, timeout)
            ...
```
On the other hand, as you know, the `TEST_LOG.warn()` have to wait for the `archive.uncompress(zip_path_filename, local_log_dir)`.

Your call. I have no strong opinion here.

avocado/core/remote/runner.py
```diff
@@ -264,6 +265,13 @@ def run_suite(self, test_suite, variants, timeout=0, replay_map=None):
             self.result.end_tests()
             self.job._result_events_dispatcher.map_method('post_tests',
                                                           self.job)
+            # compare the major version of remote and local version of avocado. If different write to UI and job.log
+            if str(remote_version[0]) != version.MAJOR:
+                log = logging.getLogger("avocado.job")
```
msaldivar: I agree with using `APP_LOG` & `TEST_LOG` it adds consistency between the two `runner.py` files. I'll change that in my v4 branch

avocado/core/remote/runner.py
msaldivar: It would be a warning. Using the yellow color would make sure the user doesn't miss the message as it stands at against the normal UI messages.

avocado/core/remote/runner.py
msaldivar: I see value in displaying the warning before the the tests I'm just not sure if it matters to much. At least in my use case I wouldn't kill a test just because the versions mismatch, but if a test fails the warning message would be nice. Thanks for suggestions, I'll make a new branch with the changes and submit another pull reques! t

PR #1759, opened 2017-02-07

msaldivar:
Added TEST_LOG & APP_LOG to add consistency between the runner.py files.
Changed .info(..) to .warn(..) to add yellow color to terminal screen.
Added a variable to hold the warning message.

Previous pull requests: #1753 

Signed-off-by: Maurice Saldivar maurice.a.saldivar@hpe.com

### DIFF
--- a/avocado/core/remote/runner.py
+++ b/avocado/core/remote/runner.py
@@ -22,6 +22,7 @@ import logging
 from fabric.exceptions import CommandTimeout
 
 from .test import RemoteTest
+from .. import version
 from .. import output
 from .. import remoter
 from .. import virt
@@ -32,6 +33,9 @@ from ..test import TestName
 from ...utils import astring
 from ...utils import archive
 from ...utils import stacktrace
+
+TEST_LOG = logging.getLogger("avocado.test")
+APP_LOG = logging.getLogger("avocado.app")
 
 
 class RemoteTestRunner(TestRunner):
@@ -192,6 +196,7 @@ class RemoteTestRunner(TestRunner):
         if not timeout:     # avoid timeout = 0
             timeout = None
         summary = set()
+        warning_message = "VERSIONS   : Local %s and remote %s are not the same."
 
         stdout_backup = sys.stdout
         stderr_backup = sys.stderr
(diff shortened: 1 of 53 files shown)


avocado/core/remote/runner.py
```diff
@@ -264,6 +269,10 @@ def run_suite(self, test_suite, variants, timeout=0, replay_map=None):
             self.result.end_tests()
             self.job._result_events_dispatcher.map_method('post_tests',
                                                           self.job)
+            # compare the major version of remote and local version of avocado. If different write to UI and job.log
+            if str(remote_version[0]) != version.MAJOR:
+                APP_LOG.warn(warning_message, '.'.join(map(str, remote_version)), version.VERSION)
+                TEST_LOG.warn(warning_message, '.'.join(map(str, remote_version)), version.VERSION)
```
apahim: Hmm what troubles me here is the double `join`/`map`/attribution. You can create the `warning_message` inside this `if`,  but there's no need to create it twice from scratch. Do you mind in fixing this? You can force-update then so we will be ready to merge.

msaldivar: Yup I can do that, I want to make sure I understand though. Instead of calling `join/map` twice just make one call to `join/map` that will update the `warning_message` variable and write that out, is that correct?

apahim: What I have in mind is:
```python
            if str(remote_version[0]) != version.MAJOR:
                msg = "VERSIONS   : Local %s and remote %s are not the same." %
                    ('.'.join(map(str, remote_version)), version.VERSION)
                APP_LOG.warn(msg)
                TEST_LOG.warn(msg)
```

apahim:
Btw, while you're into it, please squash the two commits.

avocado/core/remote/runner.py
msaldivar: Okay, that's what I was thinking. Thanks for the clarification.

msaldivar:
@apahim Okay I just did the force update and did a rebase to squash the commits. Let me know if there's something else I should/need to do. Thanks for all the help!

apahim:
Still, there are 2 commits.

msaldivar:
@apahim sorry, I must've did the rebase command incorrectly. I'll close this branch and make a new branch so the commits are fixed.

PR #1761, opened 2017-02-07

msaldivar:
Reduced the number of calls to join/map

Previous pull request: #1759 

### DIFF
--- a/avocado/core/remote/runner.py
+++ b/avocado/core/remote/runner.py
@@ -22,6 +22,7 @@ import logging
 from fabric.exceptions import CommandTimeout
 
 from .test import RemoteTest
+from .. import version
 from .. import output
 from .. import remoter
 from .. import virt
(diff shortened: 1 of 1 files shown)


ldoktor:
[Warn in case remote avocado version mismatches](https://trello.com/c/ZcV36ncd/679-warn-in-case-remote-avocado-version-mismatches)

avocado/core/remote/runner.py
```diff
@@ -192,6 +196,7 @@ def run_suite(self, test_suite, variants, timeout=0, replay_map=None):
         if not timeout:     # avoid timeout = 0
             timeout = None
         summary = set()
+        warning_message = "VERSIONS   : Local %s and remote %s are not the same."
```
ldoktor: Just for the record, this is not C. We don't have to define our variables at the beginning so we can only produce this `warning_message` where it's needed and only when it's needed. Anyway it's just a note for the future 😉.

msaldivar: haha very true @ldoktor! I'll be sure to keep that in mind when I submit future pull requests.

PR #1766, opened 2017-02-09
Restore start_tests() for result plugins

apahim:
With commit 2e6ecce6, the start_tests() method from results was lost.
But the runner was using that method to report the number of tests to
the UI.

This patch re-creates the start_tests() method, now using the new plugin
interface, fixing the tests number information in the UI.

### DIFF
--- a/avocado/core/plugin_interfaces.py
+++ b/avocado/core/plugin_interfaces.py
@@ -185,6 +185,12 @@ class ResultEvents(JobPreTests, JobPostTests):
     """
 
     @abc.abstractmethod
+    def start_tests(self, result):
+        """
+        Event triggered when the tests are about to start
+        """
+
+    @abc.abstractmethod
     def start_test(self, result, state):
         """
         Event triggered when a test starts running
--- a/avocado/plugins/tap.py
+++ b/avocado/plugins/tap.py
@@ -89,6 +89,9 @@ class TAPResult(ResultEvents):
         if tests > 0:
             self.__write("1..%d", tests)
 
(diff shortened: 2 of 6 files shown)


apahim:
[BUG: [Regression] The `TESTS` UI shows the number of tests, not the NO tests * NO variants](https://trello.com/c/62I3ABZ0/894-bug-regression-the-tests-ui-shows-the-number-of-tests-not-the-no-tests-no-variants)

avocado/core/plugin_interfaces.py
```diff
@@ -185,6 +185,12 @@ class ResultEvents(JobPreTests, JobPostTests):
     """
 
     @abc.abstractmethod
+    def start_tests(self, result):
+        """
+        Event triggered when the tests are about to start
+        """
```
ldoktor: OK it looks like we don't have a method where `result` would be already initialized and tests were not yet executed and this in-deed requires it. @clebergnu you should probably also check this out, whether you had something else in mind with your job API rfc which would cover this...

avocado/plugins/tap.py
```diff
@@ -89,6 +89,9 @@ def pre_tests(self, job):
         if tests > 0:
             self.__write("1..%d", tests)
 
+    def start_tests(self, result):
```
ldoktor: You should remove the `self._write("1..%d", tests)` from `pre_tests` and move it here with the proper number of tests...

apahim: good catch. thanks.

PR #1776, opened 2017-02-10
avocado.core.runner: add test directory to sys.path [V3]

apahim:
available in sys.path.

the test module when needed.

---

v3:
- Restoring sys.path to the original state after the test finishes.

v2: #1717
- Adding selftest.

V1: #1710

### DIFF
--- a/avocado/core/runner.py
+++ b/avocado/core/runner.py
@@ -359,6 +359,16 @@ class TestRunner(object):
 
         signal.signal(signal.SIGTSTP, sigtstp_handler)
 
+        # Saving the original sys.path state
+        _syspath = sys.path[:]
+
+        # Injecting the test module into the sys.path, so user can define
+        # a method from the test module to be used in func_at_exit
+        if 'modulePath' in test_factory[1]:
+            test_path = test_factory[1]['modulePath']
+            test_module_dir = os.path.abspath(os.path.dirname(test_path))
+            sys.path.insert(0, test_module_dir)
+
         proc = multiprocessing.Process(target=self._run_test,
                                        args=(test_factory, queue,))
         test_status = TestStatus(self.job, queue)
@@ -436,6 +446,9 @@ class TestRunner(object):
         # Get/update the test status
         test_state = test_status.finish(proc, time_started, cycle_timeout,
                                         step)
+
+        # Restoring the sys.path to the original state
+        sys.path = _syspath
 
         # Try to log the timeout reason to test's results and update test_state
         if abort_reason:
--- a/selftests/functional/test_runner_queue.py
+++ b/selftests/functional/test_runner_queue.py
@@ -1,0 +1,75 @@
+import os
+import sys
+import shutil
+import tempfile
+
+if sys.version_info[:2] == (2, 6):
+    import unittest2 as unittest
+else:
+    import unittest
+
+from avocado.core import exit_codes
+from avocado.utils import script
+from avocado.utils import process
+
+basedir = os.path.join(os.path.dirname(os.path.abspath(__file__)), '..', '..')
+basedir = os.path.abspath(basedir)
+
+AVOCADO_EXTERNAL_LIB = """
+def cleanup_function(arg1):
+    print('Bogus cleanup function. arg1: %s' % arg1)
+"""
+
+AVOCADO_TEST_RUNNER_QUEUE_EXTERNAL_LIB = """#!/usr/bin/env python
+from avocado import Test
+from avocado.utils import runtime as avocado_runtime
+from avocado_queue_testlib import cleanup_function
+
+
+class AvocadoQueueTests(Test):
+    def test_stuff_add_cleanup(self):
+        avocado_runtime.CURRENT_TEST.runner_queue.put({'func_at_exit': cleanup_function,
+                                                       'args': ('Hello Avocado Test Queue!',),
+                                                       'once': True})
+        self.assertEqual(1, 1)
+"""
+
+
+class TestRunnerQueue(unittest.TestCase):
+    def setUp(self):
+        os.chdir(basedir)
+        self.tmpdir = tempfile.mkdtemp(prefix='avocado_' + __name__)
+
+    def test_runner_queue_external_lib(self):
+        mylib = script.TemporaryScript(
+            'avocado_queue_testlib.py',
+            AVOCADO_EXTERNAL_LIB,
+            'avocado_runner_queue_functional',
+            0644)
+        mylib.save()
+        mytest = script.Script(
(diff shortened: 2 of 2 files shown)


apahim:

avocado/core/runner.py
```diff
@@ -359,6 +359,16 @@ def sigtstp_handler(signum, frame):     # pylint: disable=W0613
...
 
+        # Saving the original sys.path state
+        _syspath = sys.path[:]
+
+        # Injecting the test module into the sys.path, so user can define
+        # a method from the test module to be used in func_at_exit
+        if 'modulePath' in test_factory[1]:
+            test_path = test_factory[1]['modulePath']
+            test_module_dir = os.path.abspath(os.path.dirname(test_path))
+            sys.path.insert(0, test_module_dir)
```
ldoktor: Well this only works when the module is in the modulePath and it'll always update the `sys.path`, not just when `func_at_exit` is used. I have to say I had something else in mind, which should be IMO more scalable, so please let me know what do you think.

My approach would be to use `pickle.dumps` while pickling the `func_at_exit` and adding optional `sysPath` as an entry. Therefor the test code would do:

```python
self.runner_queue.put({"func_at_exit": pickle.dumps(cleanup_function),
                                               "args": ("Hello Avocado Test Queue!",),
                                               "once": True,
                                               "sysPath": [os.path.dirname(avocado_queue_testlib.__FILE__)]}
```

The execution won't be affected and only during the `func_at_exit.run` it'd do:

```python
func, args, kwargs = item
pth = sys.path[:]
try:
    sys.path.extend(self.path)
    func = pickle.loads(func)
    ...
finally:
    sys.path = pth
```

Which would be easy to move into a different process as we only change the `sys` for the cleanup. Ideally we'd also create `self.queue.func_at_exit` which would do the `pickle` for us, but for now we could simply leave it as is and support pickled and unpickled entries until we add the API to handle it from tests.

Anyway what do you say, would this work for you? I'd prefer this solution because:

1. It modifies the `sys` only for the cleanup (easy separation, lower impact)
2. It allows injecting custom paths (which is quite useful for complex tests)
3. It's not modifying anything unless explicitly asked for (therefor safe and fast)

The cons I see are:

1. It requires `pickle` from inside tests (could be addressed by `API` to register `func_at_exit` which  we keep postponing)

selftests/functional/test_runner_queue.py
```diff
@@ -0,0 +1,75 @@
...
+
+AVOCADO_TEST_RUNNER_QUEUE_EXTERNAL_LIB = """#!/usr/bin/env python
+from avocado import Test
+from avocado.utils import runtime as avocado_runtime
+from avocado_queue_testlib import cleanup_function
+
+
+class AvocadoQueueTests(Test):
+    def test_stuff_add_cleanup(self):
+        avocado_runtime.CURRENT_TEST.runner_queue.put({'func_at_exit': cleanup_function,
```
ldoktor: How about `self.runner_queue`? I hope we remove `CURRENT_TEST` from the runtime as I really hate it.

selftests/functional/test_runner_queue.py
```diff
@@ -0,0 +1,75 @@
...
+        mytest.save()
+        # job should be able to finish under 5 seconds. If this fails, it's
+        # possible that we hit the "simple test fork bomb" bug
+        cmd_line = ['./scripts/avocado',
+                    'run',
+                    '--sysinfo=off',
+                    '--job-results-dir',
+                    "%s" % self.tmpdir,
+                    "%s" % mytest]
+        result = process.run(' '.join(cmd_line))
```
ldoktor: Any reasons not to simply create it as a string?

selftests/functional/test_runner_queue.py
```diff
@@ -0,0 +1,75 @@
...
+    def setUp(self):
+        os.chdir(basedir)
+        self.tmpdir = tempfile.mkdtemp(prefix='avocado_' + __name__)
+
+    def test_runner_queue_external_lib(self):
+        mylib = script.TemporaryScript(
+            'avocado_queue_testlib.py',
+            AVOCADO_EXTERNAL_LIB,
+            'avocado_runner_queue_functional',
+            0644)
```
ldoktor: Please use `0o644` which works well in py2 and py3

avocado/core/runner.py
apahim: I'd like to avoid telling test writers to use pickle for this to work. Is way easier for then just to set the PYTHONPATH before running this test. Let me try something else. I get back to you soon.

selftests/functional/test_runner_queue.py
apahim: ok

selftests/functional/test_runner_queue.py
apahim: Ok

selftests/functional/test_runner_queue.py
apahim: Well, it was created by @lmr, but I like the organization. No need to change it tough.

avocado/core/runner.py
apahim: Also, the moment when the sys.path should be ready is in the `queue.get()`, not in `func_at_exit.run()`, what makes this tricky. Anyway, working on a new version.

selftests/functional/test_runner_queue.py
ldoktor: Well you can still use:
```python
("foo"
 "bar %s" % "bbb"
 "baz %s" % "qqq")
```

to keep the organization but avoid the unnecessary code. I consider this as ugly...

avocado/core/runner.py
ldoktor: Well the important thing for me is to avoid modifying `sys.path` while running the test. Other solutions are welcome and as I said, the target usage is `self.runner_queue.func_at_exit(...)` which would be a method of `runner_queue` object, which would simply pickle the stuff and send it via itself to the runner. The thing is that `runner_queue` is still only a `SimpleQueue` because we haven't created this API yet. But more and more I think it'd be that simple that we should go ahead and do it...

apahim:
@ldoktor, please have a look at #1778.

avocado/core/runner.py
lmr: Pickling is cumbersome to say the least. My 2 cents, don't go that route.

ldoktor: @lmr yes, pickling is cumbersome, but the SimpleQueue already pickles the object so if the object is unpickable, we can't send it through in the first place. Therefor adding a second pickle to avoid direct unpickling from the main avocado runner is not a big difference, but it's a huge security/safety benefit, don't you think?

PR #1786, opened 2017-02-15
Remove the TESTS line from the UI

apahim:
This information is not accurate since it does not consider the number
of variants. Also, if it was accurate, it would be redundant.

Let's just remove it.  

### DIFF
--- a/avocado/plugins/human.py
+++ b/avocado/plugins/human.py
@@ -51,7 +51,6 @@ class Human(ResultEvents):
         if replay_source_job:
             self.log.info("SRC JOB ID : %s", replay_source_job)
         self.log.info("JOB LOG    : %s", job.logfile)
-        self.log.info("TESTS      : %s", len(job.test_suite))
 
     def start_test(self, result, state):
         if not self.owns_stdout:


apahim:
[BUG: Remove the duplicate "TESTS" line from the UI](https://trello.com/c/62I3ABZ0/894-bug-remove-the-duplicate-tests-line-from-the-ui)

avocado/plugins/human.py
```diff
@@ -51,7 +51,6 @@ def pre_tests(self, job):
         if replay_source_job:
             self.log.info("SRC JOB ID : %s", replay_source_job)
         self.log.info("JOB LOG    : %s", job.logfile)
-        self.log.info("TESTS      : %s", len(job.test_suite))
```
ldoktor: In every line there could be at least one bug. Let's hope there is none here 😄.

ldoktor:
Actually to be pedantic, we should update the documentation. Anyway I think it's safe to rely on people brain and avoid extensive patches with meaningless content.

apahim:
@ldoktor I'm preparing a *cleanup* PR. I will include the docs changes in that PR then.

ldoktor:
Well I hate changes like that. It's understandable without the change so I'd leave it till we do some important update. Anyway if it bothers you that much, feel free to do it.

PR #1788, opened 2017-02-15
Runner fixes and cleanup

apahim:
- Remove unused `Remote._run()` method.
- Properly separate stderr from stdout from fabric remote executions.
- Raise a `JobError` exception when the remote job fails.


apahim:
[BUG: journal plugin crashes on failed remote test execution](https://trello.com/c/kn9Wpdwu/922-bug-journal-plugin-crashes-on-failed-remote-test-execution)

optional_plugins/runner_remote/avocado_runner_remote/__init__.py
```diff
@@ -254,51 +254,6 @@ def run(self, command, ignore_status=False, quiet=True, timeout=60):
...
-            try:
-                fabric_result = fabric.operations.run(command=command,
-                                                      quiet=quiet,
-                                                      warn_only=True,
-                                                      timeout=timeout,
-                                                      pty=False)
-                break
-            except fabric.network.NetworkError as details:
-                fabric_exception = details
-                if 'not found in known_hosts' in details.message:
```
clebergnu: Right.  `run()` at the module top level is the other one.  But, I spotted a difference at least in this line, that is, it's not present in the top level `run()` function.  Can you double check why?

optional_plugins/runner_remote/avocado_runner_remote/__init__.py
```diff
@@ -114,8 +116,8 @@ def run(command, ignore_status=False, quiet=True, timeout=60):
     end_time = time.time()
     duration = end_time - start_time
     result.command = command
-    result.stdout = str(fabric_result)
-    result.stderr = fabric_result.stderr
+    result.stdout = str(fabric_result.stdout)
+    result.stderr = str(fabric_result.stderr)
```
clebergnu: Would also be helpful to double check if fabric makes stderr available in the packaged versions of the distros we support.

optional_plugins/runner_remote/avocado_runner_remote/__init__.py
```diff
@@ -424,6 +424,9 @@ def run_test(self, references, timeout):
         try:
             result = self.remote.run(avocado_cmd, ignore_status=True,
                                      timeout=timeout)
+            if result.exit_status & exit_codes.AVOCADO_JOB_FAIL:
```
clebergnu: There's a typo in the commit message: "raise **ax** exception".

apahim: :/

optional_plugins/runner_remote/avocado_runner_remote/__init__.py
apahim: Ok. I can do it tomorrow.

optional_plugins/runner_remote/avocado_runner_remote/__init__.py
apahim: That check was lost with the commit that moved the method to the module: https://github.com/avocado-framework/avocado/commit/c06aea0f985cb9eea92bc5fde135aae0be36aa2f
But we already gave the 'unknown hosts' a good treatment after that commit, so I think we're safe in that regard.

optional_plugins/runner_remote/avocado_runner_remote/__init__.py
apahim: Yep, it's pretty much everywhere.

optional_plugins/runner_remote/avocado_runner_remote/__init__.py
clebergnu: Good, thanks for double checking that.

optional_plugins/runner_remote/avocado_runner_remote/__init__.py
clebergnu: Good.  Fabric can be surprisingly different from version to version (just found yet another example while packaging version 46.0).

apahim:
@clebergnu ack

apahim:
@clebergnu, updated.

PR #1791, opened 2017-02-16
selftests: suppress missing directory message

apahim:
checkall runs `ls` twice in the job-results dir. Once before the
selftests and once again after the tests, comparing the outputs to make
sure there's no new results there after the selftests.

When the job-results does not exist in advance (the case in travis
environments), the message 'ls: cannot access /home/travis/avocado/job-results:
No such file or directory' is shown in the logs. The message is harmless
and does not affect the check. Let's just suppress the stderr from the
`ls` command in that check.

Reference: https://trello.com/c/IBoLAJjf

### DIFF
--- a/selftests/checkall
+++ b/selftests/checkall
@@ -78,7 +78,7 @@ results_dir_content() {
     fi
 }
 
-[ "$SKIP_RESULTSDIR_CHECK" ] || RESULTS_DIR_CONTENT="$(ls $RESULTS_DIR)"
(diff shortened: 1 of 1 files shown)


apahim:
[BUG: there's no guarantee that ~/avocado/job-results exists](https://trello.com/c/IBoLAJjf/912-bug-there-s-no-guarantee-that-avocado-job-results-exists)

ldoktor:
Btw the `ls` is called twice, once from the `results_dir_content` and that one is not treated...

apahim:
But the second one is called after the selftests, when the job-results is expected to be there. No need to handle that one.

ldoktor:
Which is a bit odd, but I know why it is like that...

PR #1807, opened 2017-02-20
Fix names of runners' packages in 46.0 release notes.

mateush:
Signed-off-by: Mateusz Wagner <mateuszwag@gmail.com>
Fixes issue a with sign in a previous pull req: https://github.com/avocado-framework/avocado/pull/1804


apahim:
Notice that the fix will not be available in the version of the release notes matching the version of the released Avocado. The fix is applied on master, meaning you can see the fix here:
http://avocado-framework.readthedocs.io/en/latest/release_notes/46_0.html
But you cannot see the fix here:
http://avocado-framework.readthedocs.io/en/46.0/release_notes/46_0.html

PR #1813, opened 2017-02-23
SIGTERM handler

apahim:
We already handle the SIGINT, but when we receive a SIGTERM, we just
exit the main Avocado process and leave the subprocesses behind.

This patch creates a SIGTERM proper handler, to kill the children
processes before exiting the main process.

Reference: https://trello.com/c/6qBdaSM8

### DIFF
--- a/avocado/core/app.py
+++ b/avocado/core/app.py
@@ -17,6 +17,7 @@ The core Avocado application.
 """
 
 import os
+import psutil
 import signal
 import sys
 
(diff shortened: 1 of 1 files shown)


apahim:
[Handle SIGTERM](https://trello.com/c/6qBdaSM8/648-handle-sigterm)

avocado/core/app.py
```diff
@@ -17,6 +17,7 @@
 """
 
 import os
+import psutil
```
ldoktor: The psutil is not an avocado dependency (It is for selftests, but not for core avocado). Would you mind using our `avocado.utils.process.kill_process_tree` instead? Both functions suffer the issue where processes created at the time of discovery are not recognized/killed and ours doesn't require any additional dependency.

apahim:
Sure, thanks for pointing it out.

PR #1814, opened 2017-02-23
Handle sigterm [v2]

apahim:
v2:
- Add to `avocado.utils.process.get_children_pids()` the ability to return all levels of sub-processes (using the argument `recursive=True`).
- Use `avocado.utils.process` instead os `psutil`.

v1: #1813

### DIFF
--- a/avocado/core/app.py
+++ b/avocado/core/app.py
@@ -25,6 +25,7 @@ from .dispatcher import CLICmdDispatcher
 from .dispatcher import CLIDispatcher
 from .output import STD_OUTPUT
 from .parser import Parser
+from ..utils import process
 
 
 class AvocadoApp(object):
@@ -38,6 +39,16 @@ class AvocadoApp(object):
         # Catch all libc runtime errors to STDERR
         os.environ['LIBC_FATAL_STDERR_'] = '1'
 
+        def sigterm_handler(signum, frame):     # pylint: disable=W0613
+            children = process.get_children_pids(os.getpid(), recursive=True)
+            for child in children:
+                try:
+                    os.kill(int(child), signal.SIGTERM)
+                except OSError:
+                    pass
+            raise SystemExit('Terminated')
+
+        signal.signal(signal.SIGTERM, sigterm_handler)
         signal.signal(signal.SIGTSTP, signal.SIG_IGN)   # ignore ctrl+z
         self.parser = Parser()
         output.early_start()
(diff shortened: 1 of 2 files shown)


apahim:
[Handle SIGTERM](https://trello.com/c/6qBdaSM8/648-handle-sigterm)

ldoktor:
Do you have any reason no to use the `process.kill_process_tree` which is designed for this task and results in more-processes-killed on a fork-bomb (therefor better effort than getting all kids and then killing them)?

avocado/core/app.py
```diff
@@ -38,6 +39,16 @@ def __init__(self):
         # Catch all libc runtime errors to STDERR
         os.environ['LIBC_FATAL_STDERR_'] = '1'
 
+        def sigterm_handler(signum, frame):     # pylint: disable=W0613
+            children = process.get_children_pids(os.getpid(), recursive=True)
+            for child in children:
+                try:
+                    os.kill(int(child), signal.SIGTERM)
```
ldoktor: One more thing, I'd recommend using `SIGKILL` as this is basically an emergency and we want to destroy as much processes as possible.

apahim:
@ldoktor. I'm handling a `SIGTERM`, not  a `SIGKILL`. So, I don't intend to pass a `SIGKILL` to the children when receiving a `SIGTERM` in the parent.
With that said, the `kill_process_tree(pid, sig=signal.SIGTERM)` will not terminate the entire process tree, just the first level bellow the parent, leaving orphans from the second level down. The recursion mechanism I created in `get_children_pids()` seems a good fit for what I intend to do.
Handling `SIGTERM` is not about imploding the whole tree. It's about giving the children a chance to die with dignity.

ldoktor:
The `kill_process_tree` is recursive and it kills the childrens at the time it reaches the top level, not after the processing of the recursive tree, therefor it's more efficient on fork-bombs (or simply processes spawning processes).

As for the `SIGKILL` this is a matter of how we look at it. It's not a strong opinion but I'd still prefer that. My view is that you send `SIGTERM` to `avocado` which means "hey Avocado, would you please finish ASAP?". Now the question is what that means to the subprocess. I think avocado should say "hey subprocesses just die, I'm finishing and don't want to leave anything behind". With SIGTERM it just says `hey, subprocess, would you be so kind and terminate yourself? I'd like to finish but wel.. no rush...".

My point is you send SIGTERM to Avocado so sending SIGKILL to tests and other background processes seems like a good fit. Anyway (once again) this was a suggestion, not a requirement.

apahim:
Ok, I was approaching `kill_process_tree` differently, informing the process ID of the avocado main process. That would stop the main process and cause unexpected behavior. Instead, if we inform the PID of the first level children, it will work as expected. Sending a V3.

PR #1831, opened 2017-03-03
yaml_to_mux: fix deprecation message

apahim:
--filter-only/--filter-out are being deprecated by
--mux-filter-only/--mux-filter-out, but the deprecation
messages are pointing users to --mux-only/--mux-out, which
does not exist.

This patch fixes the deprecation messages, referencing the
correct new options.


apahim:
[BUG: man page should contain all command line options](https://trello.com/c/UB6hG2mD/901-bug-man-page-should-contain-all-command-line-options)

PR #1832, opened 2017-03-03
Add checkpoint plugin and allow resuming tests when replaying

benzea:
The idea here is to allow resuming a test in case avocado or the machine
dies (for example running a local test that might cause a kernel panic).
When a test run is recovered, the tests that were already run will be
skipped (and their old result sets copied if requested). On any test
marked as INTERRUPTED the original test will be run but the
recover_failure method will be called prior to any operation. If it is
not implemented, then the default implementation is to FAIL the test.

Example, interrupting avocado manually during Sleeptest:
$ ./scripts/avocado run ./examples/tests/sleeptest.py ./examples/tests/sleeptest.py
JOB ID     : 16e4e4e33256a6844cdbe87972031859c22c08c3
JOB LOG    : /home/benjamin/avocado/job-results/job-2017-03-03T17.10-16e4e4e/job.log
 (1/2) ./examples/tests/sleeptest.py:SleepTest.test: PASS (1.02 s)
 (2/2) ./examples/tests/sleeptest.py:SleepTest.test:  ^\Quit (core dumped)

Here SleepTest is patched to return SKIP from recover_failure.
$ ./scripts/avocado run --replay 16e4e4e --resume
Using src job Mux data only, use `--replay-ignore variants` to override them.
JOB ID     : 601a3aa20c63f59f19b207dcb75672c7a66e7d4e
SRC JOB ID : 16e4e4e33256a6844cdbe87972031859c22c08c3
JOB LOG    : /home/benjamin/avocado/job-results/job-2017-03-03T17.11-601a3aa/job.log
 (1/2) ./examples/tests/sleeptest.py:SleepTest.test: SKIP
 (2/2) ./examples/tests/sleeptest.py:SleepTest.test: SKIP
RESULTS    : PASS 0 | ERROR 0 | FAIL 0 | SKIP 2 | WARN 0 | INTERRUPT 0
TESTS TIME : 0.00 s
JOB HTML   : /home/benjamin/avocado/job-results/job-2017-03-03T17.11-601a3aa/html/results.html

With --fail-interrupted the test fails:
$ ./scripts/avocado run --replay 16e4e4e --resume --fail-interrupted
Using src job Mux data only, use `--replay-ignore variants` to override them.
JOB ID     : 030432144585754a746512806f377a3772fdc745
SRC JOB ID : 16e4e4e33256a6844cdbe87972031859c22c08c3
JOB LOG    : /home/benjamin/avocado/job-results/job-2017-03-03T17.12-0304321/job.log
 (1/2) ./examples/tests/sleeptest.py:SleepTest.test: SKIP
 (2/2) ./examples/tests/sleeptest.py:SleepTest.test: ERROR (0.02 s)
RESULTS    : PASS 0 | ERROR 1 | FAIL 0 | SKIP 1 | WARN 0 | INTERRUPT 0
TESTS TIME : 0.02 s
JOB HTML   : /home/benjamin/avocado/job-results/job-2017-03-03T17.12-0304321/html/results.html

Adding --replay-results causes the first test passes:
$ ./scripts/avocado run --replay 16e4e4e --resume --fail-interrupted --replay-results
Using src job Mux data only, use `--replay-ignore variants` to override them.
JOB ID     : 11767a55901b8fac8ea9aafcc2604047110cecad
SRC JOB ID : 16e4e4e33256a6844cdbe87972031859c22c08c3
JOB LOG    : /home/benjamin/avocado/job-results/job-2017-03-03T17.14-11767a5/job.log
 (1/2) ./examples/tests/sleeptest.py:SleepTest.test: PASS (1.02 s)
 (2/2) ./examples/tests/sleeptest.py:SleepTest.test: ERROR (0.02 s)
RESULTS    : PASS 1 | ERROR 1 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0
TESTS TIME : 1.04 s
JOB HTML   : /home/benjamin/avocado/job-results/job-2017-03-03T17.14-11767a5/html/results.html

### DIFF
--- a/avocado/core/loader.py
+++ b/avocado/core/loader.py
@@ -275,7 +275,7 @@ class TestLoaderProxy(object):
         :type test_factory: tuple
         :return: an instance of :class:`avocado.core.test.Test`.
         """
-        test_class, test_parameters = test_factory
+        test_class, test_parameters, test_decorator = test_factory
         # discard tags, as they are *not* intended to be parameters
         # for the test, but used previously during filtering
         if 'tags' in test_parameters:
@@ -308,7 +308,14 @@ class TestLoaderProxy(object):
                     if issubclass(obj, test.Test):
                         test_class = obj
                         break
-        test_instance = test_class(**test_parameters)
+        if test_decorator is not None:
+            if isinstance(test_decorator, tuple):
+                test_instance = test_decorator[0](test_class, test_parameters,
+                                                  test_decorator[1])
+            else:
+                test_instance = test_decorator(test_class, test_parameters)
+        else:
+            test_instance = test_class(**test_parameters)
 
         return test_instance
 
--- a/avocado/core/runner.py
+++ b/avocado/core/runner.py
@@ -484,10 +484,10 @@ class TestRunner(object):
                            " present in test factory: %s"
                            % (template[0], template[1]))
                     raise ValueError(msg)
-                factory = [template[0], template[1].copy()]
+                factory = [template[0], template[1].copy(), None]
                 factory[1]["params"] = params
             else:
-                factory = template
+                factory = tuple(template[0], template[1].copy(), None)
             yield factory, variant
 
     def run_suite(self, test_suite, variants, timeout=0, replay_map=None):
@@ -530,16 +530,16 @@ class TestRunner(object):
                         summary.add('INTERRUPTED')
                         if 'methodName' in test_parameters:
                             del test_parameters['methodName']
-                        test_factory = (test.TimeOutSkipTest, test_parameters)
+                        test_factory = (test.TimeOutSkipTest, test_parameters,
+                                        None)
                         break_loop = not self.run_test(test_factory, queue,
                                                        summary)
                         if break_loop:
                             break
                     else:
-                        if (replay_map is not None and
-                                replay_map[index] is not None):
-                            test_parameters["methodName"] = "test"
-                            test_factory = (replay_map[index], test_parameters)
+                        if (replay_map is not None and index < len(replay_map)):
+                            test_factory = (test_factory[0], test_factory[1],
+                                            replay_map[index])
 
                         break_loop = not self.run_test(test_factory, queue,
                                                        summary, deadline)
--- a/avocado/core/test.py
+++ b/avocado/core/test.py
@@ -144,7 +144,8 @@ class Test(unittest.TestCase):
     timeout = None
 
     def __init__(self, methodName='test', name=None, params=None,
-                 base_logdir=None, job=None, runner_queue=None):
+                 base_logdir=None, job=None, runner_queue=None,
+                 recovery_required=False, previous_run=None):
         """
         Initializes the test.
 
@@ -244,6 +245,9 @@ class Test(unittest.TestCase):
         self.paused_msg = ''
 
         self.runner_queue = runner_queue
+
+        self.__recovery_required = recovery_required
+        self.previous_run = previous_run
 
         unittest.TestCase.__init__(self, methodName=methodName)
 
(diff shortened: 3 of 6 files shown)


benzea:
[Create a `resume` option for the job replay feature after a crash](https://trello.com/c/fGgmg8jd/946-create-a-resume-option-for-the-job-replay-feature)

benzea:
[Create a `resume` option for the job replay feature after a crash](https://trello.com/c/fGgmg8jd/946-create-a-resume-option-for-the-job-replay-feature)

avocado/core/loader.py
```diff
@@ -308,7 +308,14 @@ def load_test(self, test_factory):
                     if issubclass(obj, test.Test):
                         test_class = obj
                         break
-        test_instance = test_class(**test_parameters)
+        if test_decorator is not None:
+            if isinstance(test_decorator, tuple):
```
apahim: This is becoming very hackish. `test_factory` should be known format and well documented. Right now, it' s `:param test_factory: a pair of test class and parameters.`. With this change, it can be some other things as well. Or not :)
I don't like the idea of coding the loader to solve problems that are up to the runner. I'd try to keep the loader free of changes.

avocado/core/runner.py
```diff
@@ -530,16 +530,16 @@ def run_suite(self, test_suite, variants, timeout=0, replay_map=None):
...
                         if break_loop:
                             break
                     else:
-                        if (replay_map is not None and
-                                replay_map[index] is not None):
-                            test_parameters["methodName"] = "test"
-                            test_factory = (replay_map[index], test_parameters)
+                        if (replay_map is not None and index < len(replay_map)):
+                            test_factory = (test_factory[0], test_factory[1],
+                                            replay_map[index])
```
apahim: As I said, I don't fancy this third element in the `test_factory`. But I see what you're  trying here. I'd keep all this logic in the runner and keep the `test_factory` as is.

avocado/core/test.py
```diff
@@ -144,7 +144,8 @@ class Test(unittest.TestCase):
     timeout = None
 
     def __init__(self, methodName='test', name=None, params=None,
-                 base_logdir=None, job=None, runner_queue=None):
+                 base_logdir=None, job=None, runner_queue=None,
+                 recovery_required=False, previous_run=None):
```
apahim: Should the test have knowledge about the recovery? I think in this more like a replay functionality. The very test that was under execution when the interruption takes place should be either skipped or executed. I don' t see a reason to raise a `TestError`.

avocado/core/test.py
```diff
@@ -978,6 +1010,52 @@ class ReplaySkipTest(SkipTest):
     _skip_reason = "Test skipped due to a job replay filter!"
 
 
+class ReplayResultCopier(Test):
```
apahim: This is another point of discussion. As a concept, a skipped test is a test that is not executed. We cannot say we skipped the test and still it contains the execution information. If that' s a requirement for you, than we should introduce a new status, like ` RECOVERED`, with proper specification and documentation. But in my view, I'd just skip it and don't populate it's results.

avocado/plugins/checkpoint.py
```diff
@@ -0,0 +1,90 @@
...
+# See LICENSE for more details.
+#
+# Copyright: Red Hat Inc. 2014, 2017
+# Author: Cleber Rosa <cleber@redhat.com>
+# Author: Benjamin Berg <bberg@redhat.com>
+
+"""Checkpoint Plugin"""
+
+import os
+import sqlite3
```
apahim: `sqlite3`

avocado/core/test.py
```diff
@@ -770,6 +789,19 @@ def fetch_asset(self, name, asset_hash=None, algorithm='sha1',
...
+    def recover_failure(self, previous_run):
+        """
+        Method which is called before setUp and other functions to recover if
+        the previous run was interrupted. Avocado does not know about the tests
+        state, for recovery only the incomplete result object is provided
+        which may allow recovering some data from the tests log directory.
+
+        :param previous_run: Stored test status from previous run
+        """
+
```
apahim: Where is the recovery code? Anyway, this could be simply conceptual. A test that was interrupted during it's execution should be, imo, executed again in the `resume` command.

benzea:
Thanks for the review!

 * Agree that the reply results thing doesn't really belong where I put it. I was somewhat experimenting there as something like that could simplify collecting/uploading results later. At the same time, actually modifying the old job ID would be entirely sane in such a case.
 * Probably true that touching the test_factory isn't necessary, in particular when not doing any copying foo.
 * For the recovery function I was considering having the ability for the test to mark itself as FAILED in case it was interrupted earlier and is resumed later. To explain more, what I am considering for example is putting the computer into different sleep states and then mark the test as failed if it is resumed after a reboot instead of simply continuing after the computer is woken again. The test could easily store some persistent data though to detect the error case in the setUp or the test functions.
 * Putting the code to sync everything into the json plugin sounds good to me.

apahim:
Hi @benzea, let me close this PR. Looking forward to the V2.

PR #1852, opened 2017-03-09
docs: tearDown() on test skip

apahim:
Mention the behaviour of different skip forms regarding the tearDown().

Reference: https://trello.com/c/gsdBAmMY


apahim:
[Running `tearDown` when test is skipped by decorator](https://trello.com/c/gsdBAmMY/937-running-teardown-when-test-is-skipped-by-decorator)

PR #1855, opened 2017-03-10
selftests: selfcheck with results in a tempdir

apahim:
To make sure the seltests does not leave job results behind,
we introduced a check which compares the job-results directory
content from before the selftets with the content after the
selftests. But using `AVOCADO_SELF_CHECK=1 make check`, we
expect at least one extra result in the job-results directory,
making the content of the job-results to not be the same after
the tests.

This patch changes the avocado command used for self check
in `make check`, pointing the results to a temporary directory.

Reference: https://trello.com/c/y57xZSUS

### DIFF
--- a/selftests/checkall
+++ b/selftests/checkall
@@ -89,7 +89,7 @@ if [ "$AVOCADO_PARALLEL_CHECK" ]; then
 elif [ -z "$AVOCADO_SELF_CHECK" ]; then
     run_rc selftests selftests/run
 else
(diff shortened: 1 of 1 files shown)


apahim:
[[BUG] `AVOCADO_SELF_CHECK=1 make check` is broken](https://trello.com/c/y57xZSUS/947-bug-avocado-self-check-1-make-check-is-broken)

PR #1866, opened 2017-03-17
Adding requirement plugins

ggopinat:
Similar to make link the Makefile now walks throught the plugins
and install dependencies depending on the makefile type.

https://trello.com/c/R9rxte6o/608-add-make-requirements-plugins

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,10 +140,13 @@ requirements-selftests: requirements
 	- pip install -r requirements-selftests.txt
 
 requirements-plugins: requirements
-	for MAKEFILE in $(AVOCADO_PLUGINS);\
-		do AVOCADO_DIRNAME=$(AVOCADO_DIRNAME) make -C $$MAKEFILE requirements &>/dev/null && echo ">> DONE $$MAKEFILE" || echo ">> SKIP $$MAKEFILE";\
-	done
-
+	for MAKEFILE in $(AVOCADO_PLUGINS);do\
+		if test -f $$MAKEFILE/Makefile -o -f $$MAKEFILE/setup.py -o -f $$MAKEFILE/requirements.txt; then echo ">> REQUIREMENTS $$MAKEFILE";\
+                        if test -f $$MAKEFILE/Makefile; then AVOCADO_DIRNAME=$(AVOCADO_DIRNAME) make -C $$MAKEFILE requirement &>/dev/null;\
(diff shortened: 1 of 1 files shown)


ldoktor:
Hello @ggopinat the Travis failure is due to your old code-base. Please rebase your patch on top of the latest master. To do that simply do:

```
git fetch
git rebase --ignore-date origin/master
```

if `origin` is the upstream repo. If the upstream repo is called differently, use:

```
git fetch $UPSTREAM
git rebase --ignore-date $UPSTREAM/master
```
and feel free to force-update this PR (or create a new one, it's up to you).

ldoktor:
Actually I have a question, why do you try to `develop` the existing `setup.py`? I'd go only with `requirements.txt` or `make requirements` as the `develop` actually prepares it for execution, which might clash with `make install` and is performed by `make link` anyway...

clebergnu:
@ldoktor `develop` will run `setup.py` and find `install_requires` which may not be in a `requirements.txt` file.

ldoktor:
Yes, let me summarize what we came up with on meeting:

Running `make develop`
------------------------------------

(+) will download all requirements
(-) deploys the code which means people using `make install` will still be using the `develop` version, which could lead to hidden issues as they would assume they use the `installed` version, but in fact will be using the currently checked-out version (IMO big issue)
(-) deploys the code therefor people having multiple clones on a single computer might assume they are using the last one they `make link`ed but this link will be overridden by `make requirements-plugins`
(-) runs the `develop` twice when `make requirements-plugins` and `make link` is used

Not running `make develop`
----------------------------

(+) won't change the user environment (code is not deployed)
(-) won't install the "hardcoded" dependencies, they will be only checked on `make install` or `make link`, so for the first time people using `make requirements-plugin` online and then switching offline might end up with not all dependenices (I don't see this as a big issue as the deps will be sorted during deployment)

I'd prefer the `not running make develop` as I dislike the idea of `make requirement-plugins` spoiling my environment (in other way than just installing dependencies) but I can live with both.

clebergnu:
@ggopinat it may be a good idea for you to give your opinion here.

clebergnu:
[Add `make requirements-plugins`](https://trello.com/c/R9rxte6o/608-add-make-requirements-plugins)

clebergnu:
Closing in favor of #2001

PR #1868, opened 2017-03-21
Robot Plugin

apahim:
This patch creates an optional plugin to run Robot Framework tests
inside Avocado.

### DIFF
--- a/optional_plugins/robot/avocado_robot/__init__.py
+++ b/optional_plugins/robot/avocado_robot/__init__.py
@@ -1,0 +1,131 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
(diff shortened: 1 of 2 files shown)


optional_plugins/robot/avocado_robot/__init__.py
```diff
@@ -0,0 +1,131 @@
...
+            robot_binary = utils_path.find_command(ROBOT_BINARY_NAME)
+        except utils_path.CmdNotFoundError:
+            raise exceptions.TestError('"%s" binary not found.' %
+                                       ROBOT_BINARY_NAME)
+
+        suite_name, test_name = self.name.name.split(':')[1].split('.')
+        cmd = '%s --suite "%s" --test "%s" %s' % (robot_binary,
+                                                  suite_name,
+                                                  test_name,
+                                                  self.filename)
```
lmr: You probably have considered using http://robot-framework.readthedocs.io/en/3.0.2/autodoc/robot.html#robot.run instead of running the subprocess above. It would be interesting for the purposes of documentation to explain why you chose the subprocess way instead of the robot framework API, though.

apahim: Oh, you're right. Missed completely that part. Sending a V2.  Thank you.

apahim: Yes, actually I remember that I started with that last week and than changed so we can run as subprocesses. I'm adding some documentation on that. Thank you.

PR #1872, opened 2017-03-22
Adjusts in SKIP and CANCEL

apahim:
- Start the deprecation process for `self.skip()`.
- Don't execute `setUp()` and `tearDown()` when using skip decorators.
- Docs adjustments.

Reference: https://trello.com/c/viBJIEwI

PR #1874, opened 2017-03-22
Adjusts in SKIP and CANCEL [v2]

apahim:
**Goals**:

SKIP

- Skip decorators can decorate only test methods (not `setUp()`/`tearDown()`).
- `self.skip()` can be used only in `setUp()`

CANCEL:

DOCS:

---

v2:

v1: #1872 
- Docs adjustments.

### DIFF
--- a/avocado/core/decorators.py
+++ b/avocado/core/decorators.py
@@ -65,6 +65,7 @@ def skip(message=None):
             def wrapper(*args, **kwargs):
                 raise core_exceptions.TestDecoratorSkip(message)
             function = wrapper
+        function.__skip__ = True
         return function
     return decorator
 
--- a/avocado/core/test.py
+++ b/avocado/core/test.py
@@ -560,7 +560,8 @@ class Test(unittest.TestCase):
         stdout_check_exception = None
         stderr_check_exception = None
         try:
-            self.setUp()
+            if not getattr(testMethod, "__skip__", False):
+                self.setUp()
         except (exceptions.TestSetupSkip,
                 exceptions.TestDecoratorSkip,
                 exceptions.TestTimeoutSkip,
@@ -605,7 +606,8 @@ class Test(unittest.TestCase):
                 self.log.debug(' -> %s %s: %s', key, type(value), value)
         finally:
             try:
-                self.tearDown()
+                if not getattr(testMethod, "__skip__", False):
+                    self.tearDown()
             except exceptions.TestSetupSkip as details:
                 stacktrace.log_exc_info(sys.exc_info(), logger='avocado.test')
                 skip_illegal_msg = ('Calling skip() in places other than '
--- a/avocado/core/test.py
+++ b/avocado/core/test.py
@@ -569,11 +569,7 @@ class Test(unittest.TestCase):
             raise exceptions.TestSkipError(details)
         except exceptions.TestCancel as details:
             stacktrace.log_exc_info(sys.exc_info(), logger='avocado.test')
-            skip_illegal_msg = ('Calling cancel() in setUp() '
-                                'is not allowed in avocado, you '
-                                'must fix your test. Original cancel exception: '
-                                '%s' % details)
-            raise exceptions.TestError(skip_illegal_msg)
+            raise
         except exceptions.TestDecoratorSkip as details:
             stacktrace.log_exc_info(sys.exc_info(), logger='avocado.test')
             skip_illegal_msg = ('Using skip decorators in places other than '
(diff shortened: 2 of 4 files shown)


apahim:

avocado/core/test.py
```diff
@@ -782,7 +782,11 @@ def skip(self, message=None):
         :param message: an optional message that will be recorded in the logs
         :type message: str
         """
-        raise exceptions.TestSetupSkip(message)
+        dep_msg = '[WARNING: self.skip() will be deprecated, please ' \
+                  'use self.cancel()] '
+        if message is not None:
+            dep_msg += message
```
ldoktor: Well message could be anything, I'd suggest using `"%s %s" % (dep_msg, message)`.

docs/source/WritingTests.rst
```diff
@@ -957,6 +957,9 @@ Avocado offers some options for the test writers to skip a test:
 Test ``skip()`` Method
 ----------------------
 
+.. warning:: `self.skip()` is under deprecation process. Please adjust
+   your tests to use the `self.cancel()` or the skip decorators instead.
```
ldoktor: Well at this point `self.cancel` is not supported during `setUp`, therefor people can't adjust their scripts. I know you are changing this later in this PR but I'd suggest changing the order of commits to first make `self.cancel` as powerful as `sefl.skip` and then add the deprecation message.

ldoktor: Also it'd be nice to add a date/version where we are going to deprecate the `skip`

avocado/core/decorators.py
```diff
@@ -65,6 +65,7 @@ def decorator(function):
             def wrapper(*args, **kwargs):
                 raise core_exceptions.TestDecoratorSkip(message)
             function = wrapper
+        function.__skip__ = True
```
ldoktor: Well this should be unique enough, but how about using something even less prone to accidental clash like `__skip_this_test_due_decorator__` or so?

avocado/core/test.py
```diff
@@ -560,7 +560,8 @@ def _run_avocado(self):
         stdout_check_exception = None
         stderr_check_exception = None
         try:
-            self.setUp()
+            if not getattr(testMethod, "__skip__", False):
```
ldoktor: Again to avoid accidental clashes, let's use `if getattr() is True`

avocado/core/decorators.py
ldoktor: Btw good idea 😉

avocado/core/test.py
```diff
@@ -569,11 +569,7 @@ def _run_avocado(self):
             raise exceptions.TestSkipError(details)
         except exceptions.TestCancel as details:
             stacktrace.log_exc_info(sys.exc_info(), logger='avocado.test')
-            skip_illegal_msg = ('Calling cancel() in setUp() '
-                                'is not allowed in avocado, you '
-                                'must fix your test. Original cancel exception: '
-                                '%s' % details)
-            raise exceptions.TestError(skip_illegal_msg)
+            raise
```
ldoktor: This means `tearDown` won't be executed, I thought we changed it (in https://trello.com/c/viBJIEwI/941-add-support-to-cancel-test-execution there is this sentence - `teardown() is always run when using self.cancel()`)

anyway as we don't run `tearDown` on `setUp` failure I'm fine with that. (more-like I'd like to change that, I'll try to write an RFC)

ldoktor: Also why do you treat `TestCancel` differently than plain exception? I'd just remove this block and it'd get treated by `except` including `avocado.test` traceback, which is useful when looking for where it got actually skipped (there can be many locations with the same message)

avocado/core/decorators.py
apahim: sure

avocado/core/test.py
apahim: ok

avocado/core/test.py
apahim: ok, np

docs/source/WritingTests.rst
apahim: yes, good catch.

avocado/core/test.py
apahim: tearDown() should be executed here. Fixing.

avocado/core/test.py
```diff
@@ -782,7 +782,11 @@ def skip(self, message=None):
         :param message: an optional message that will be recorded in the logs
         :type message: str
         """
-        raise exceptions.TestSetupSkip(message)
+        dep_msg = '[WARNING: self.skip() will be deprecated, please ' \
+                  'use self.cancel()] '
```
ldoktor: One more thing, you should probably change the `examples/tests/skiponsetup.py` in the same commit to demonstrate the required change.

PR #1878, opened 2017-03-23
Adjusts in SKIP and CANCEL [V3]

apahim:
**Goals**:

SKIP

- Skip decorators can decorate only test methods (not `setUp()`/`tearDown()`).
- `self.skip()` can still be used only in `setUp()`

CANCEL:

DOCS:

---
v3:
- Reordered commits so the deprecation `self.skip()` deprecation message reflects a working behavior.
- Make `self.cancel()` called in `setUp()` to run `tearDown()`.
- Marge the two commits handling `SKIP` in one commit only.
- Renamed `function.__skip__` to `function.__skip_test_decorator__`.
- Be explicit in the `True`/`False` test of `function.__skip_test_decorator__`.
- Selftests improvements to cover the changes in this PR.
- Change example test `skiponsetup.py` to reflect the deprecation process.

v2: #1874 

v1: #1872 
- Docs adjustments.


selftests/functional/test_skiptests.py
```diff
@@ -80,9 +114,40 @@ def test_skip_decorators(self):
...
         result = process.run(' '.join(cmd_line), ignore_status=True)
         json_results = json.loads(result.stdout)
+        debuglog = json_results['debuglog']
+
         self.assertEqual(result.exit_status, exit_codes.AVOCADO_ALL_OK)
-        self.assertEquals(json_results['pass'], 2)
-        self.assertEquals(json_results['skip'], 4)
+        self.assertEquals(json_results['skip'], 3)
+        self.assertFalse('setup executed' in open(debuglog, 'r').read())
+        self.assertFalse('teardown executed' in open(debuglog, 'r').read())
```
ldoktor: Nice touch 😄

ldoktor: Btw while on it how about also adding check that the `test` method is not executed? (I know right now it does not make sense, but in the future we might introduce different layout and it might actually happen...)

apahim: Sure. Are you done with the review in this version?

clebergnu:

PR #1880, opened 2017-03-23
Robot Plugin [V4]

apahim:
v4:
- Make robot loader always registered along with default loaders.
- Fix discovery recursion.
- Fix `avocado list` results.
- Accept extra regex in the test reference the same way we accept in builtin loaders (`avocado list ~/Downloads/WebDemo/login_tests/:invalid_login.*Empty`)

v3: #1873 
- Handle `robot.run()` `stderr`.
- Handle `robot.run()` return code.

v2: #1869 
- Use `robot.run()` instead of robot command.

v1: #1868


clebergnu:
[Run Robot tests in Avocado](https://trello.com/c/p6eFlEmL/974-run-robot-tests-in-avocado)

PR #1882, opened 2017-03-24
Adjusts in SKIP and CANCEL [V4]

apahim:
**Goals**:

SKIP

- `self.skip()` can still be used only in `setUp()`

CANCEL:

DOCS:

---
v4:
- Removed the code that would not accept skip decorators to be used in `setUp()`. This topic will be internally discussed before come to github.

v3: #1878 
- Reordered commits so the deprecation `self.skip()` deprecation message reflects a working behavior.
- Make `self.cancel()` called in `setUp()` to run `tearDown()`.
- Marge the two commits handling `SKIP` in one commit only.
- Renamed `function.__skip__` to `function.__skip_test_decorator__`.
- Be explicit in the `True`/`False` test of `function.__skip_test_decorator__`.
- Selftests improvements to cover the changes in this PR.
- Change example test `skiponsetup.py` to reflect the deprecation process.

v2: #1874 

v1: #1872 
- Docs adjustments.


apahim:

avocado/core/test.py
```diff
@@ -572,17 +573,18 @@ def _run_avocado(self):
...
-            raise exceptions.TestError(skip_illegal_msg)
+            cancel_test = details
         except:  # Old-style exceptions are not inherited from Exception()
             stacktrace.log_exc_info(sys.exc_info(), logger='avocado.test')
             details = sys.exc_info()[1]
             raise exceptions.TestSetupFail(details)
         try:
+            #TODO: This should be removed when this try/testMethod()
+            #      becomes an else/testMethod(). Then the TestCance
+            #      exception has to be raised in the setUp() `except`.
```
ldoktor: Well this is not yet decided so we don't actually know what happens... I'd just leave it on the one who takes this task to decide how is he going to implement it...

ldoktor: I mean there is a big room for optimization including the `__skip_test_decorator__` which can be tackled at the beginning, not polluting the main execution...

ldoktor: ... which means no actions are required, just leaving a message that the comment was probably not really essential...

docs/source/WritingTests.rst
```diff
@@ -960,6 +960,10 @@ Avocado offers some options for the test writers to skip a test:
 Test ``skip()`` Method
 ----------------------
 
+.. warning:: `self.skip()` will be deprecated by the end of 2017.
```
ldoktor: Nitpick: on English lessons we learned that `by ...` means sometimes before but close to the $DATE and `on ...` means exactly on the $DATE. Shouldn't this be `at the end of 2017`? (not that it'd matters, I'm just wondering whether my English is up to date and how it works in the real life...

examples/tests/cancelonsetup.py
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env python
```
ldoktor: Well moving the `skip` to `cancel` would make it clearer, but this is also understandable...

ldoktor: Also you could probably remove the `skip_outside_setup` as it is a negative test which is not used anywhere...

docs/source/WritingTests.rst
apahim: I meant close to the end of the year. Like a Christmas gift :)
But I can change it. What would you prefer?

avocado/core/test.py
apahim: Sure, I can remove it.

docs/source/WritingTests.rst
ldoktor: I don't mind the current, I'm curious whether I remember the rules correctly and what the native say about it as international English is sometimes different from the Oxford one...

apahim:
@ldoktor, sure, tranks. V5 on the way.

PR #1899, opened 2017-03-31
Linux modules modinfo enhancements

narasimhan-v:


avocado/utils/linux_modules.py
```diff
@@ -119,7 +119,12 @@ def loaded_module_info(module_name):
                 elif key == 'depends':
                     value = items[1].split(',')
                 elif key == 'parm':
-                    param_list.append(items[1].split(':')[0])
+                    param_dic = {'type': None}
+                    param_dic['name'] = items[1].split(':')[0]
+                    param_type = re.search(r"\((\w+)\)", items[-1])
+                    if param_type is not None:
+                        param_dic['type'] = param_type.group(1)
+                    param_list.append(param_dic)
```
ldoktor: Well the only problem I have here is that it again changes the format of the returned info:

Old:
```
{'depends': ['drm_kms_helper', 'drm', 'video', 'i2c-algo-bit'],
 'filename': '/lib/modules/4.9.13-201.fc25.x86_64/kernel/drivers/gpu/drm/i915/i915.ko.xz',
 'name': 'i915',
 'params': ['modeset',
            'panel_ignore_lid',
            'semaphores',
            'enable_rc6',
            'enable_dc',
            'enable_fbc',
            'lvds_channel_mode',
...
```

New
```
{'depends': ['drm_kms_helper', 'drm', 'video', 'i2c-algo-bit'],
 'filename': '/lib/modules/4.9.13-201.fc25.x86_64/kernel/drivers/gpu/drm/i915/i915.ko.xz',
 'name': 'i915',
 'params': [{'name': 'modeset', 'type': 'int'},
            {'name': 'panel_ignore_lid', 'type': 'int'},
            {'name': 'semaphores', 'type': 'int'},
            {'name': 'enable_rc6', 'type': 'int'},
            {'name': 'enable_dc', 'type': 'int'},
            {'name': 'enable_fbc', 'type': 'int'},
            {'name': 'lvds_channel_mode', 'type': 'int'},
            {'name': 'lvds_use_ssc', 'type': 'int'},
...
```
And this change is bigger then the previous one. Anyway I hope you are the only user of this library and you do try to stabilize the results soon as when we have people depending on those we'd have to start dealing with the backward compatibility issues. Please do consider that when designing future utils...

avocado/utils/linux_modules.py
```diff
@@ -189,11 +190,14 @@ def module_is_loaded(module_name):
...
     :param module_name: Name of module to search for
     :type module_name: str
-    :return: True is module is loaded
+    :return: True if module is loaded
     :rtype: bool
     """
     module_name = module_name.replace('-', '_')
-    return bool(loaded_module_info(module_name))
+    for line in genio.read_file("/proc/modules").splitlines():
+        if line.split()[0] == module_name:
```
ldoktor: minor optimization, you are only interested in the first part, so how about using `split(None, 1)` to avoid splitting the line after first split.

avocado/utils/linux_modules.py
narasimhan-v: @ldoktor Agreed. Will make sure to keep that in mind in future, though I could not work out a format which is backward compatible for this one.

avocado/utils/linux_modules.py
narasimhan-v: Thanks, will make this change

narasimhan-v:
@ldoktor Any idea why travis fails here for selftests ?

ldoktor:
The travis failure is related to insufficient cleanup (which is odd but on our side). Anyway please rebase the commit to the latest master and it'll only run a single test...

```
# something like...
git fetch
git rebase --ignore-date origin/master
```

narasimhan-v:
@ldoktor Thanks, done the changes and a rebase too

ldoktor:
[Report argument type in `avocado.utils.linux_modules.loaded_module_info`.](https://trello.com/c/UJNqgy3Z/993-report-argument-type-in-avocado-utils-linux-modules-loaded-module-info)

PR #1913, opened 2017-04-07
support multiple docstring directives

apahim:
Using more than one docstrig directive makes Avocado to use only the
first one. This patch makes the safeloader to handle multiple docstring
directives (i.e. `:avocado: enable` and `:avocado: tags=foo`).

Reference: https://trello.com/c/6FpySXuC

### DIFF
--- a/avocado/core/safeloader.py
+++ b/avocado/core/safeloader.py
@@ -76,10 +76,10 @@ def get_docstring_directive(docstring):
     :type docstring: str
     """
     if docstring is None:
-        return None
-    result = AVOCADO_DOCSTRING_DIRECTIVE_RE.search(docstring)
+        return []
+    result = AVOCADO_DOCSTRING_DIRECTIVE_RE.findall(docstring)
     if result is not None:
-        return result.groups()[0]
+        return result
 
 
 def is_docstring_directive_enable(docstring):
@@ -89,7 +89,10 @@ def is_docstring_directive_enable(docstring):
     :rtype: bool
     """
     result = get_docstring_directive(docstring)
-    return result == 'enable'
+    if 'enable' in result:
+        return True
+
+    return False
 
 
 def is_docstring_directive_disable(docstring):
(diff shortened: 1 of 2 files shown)


apahim:
[BUG: Make it possible to define 'enable' and 'tags' docstring](https://trello.com/c/6FpySXuC/958-bug-make-it-possible-to-define-enable-and-tags-docstring)

avocado/core/safeloader.py
```diff
@@ -76,10 +76,10 @@ def get_docstring_directive(docstring):
     :type docstring: str
     """
     if docstring is None:
-        return None
-    result = AVOCADO_DOCSTRING_DIRECTIVE_RE.search(docstring)
+        return []
```
clebergnu: I actually suggest we change the method name to `get_docstring_directives` since it now returns many directives.  Bonus points for `:rtype: list`

avocado/core/safeloader.py
```diff
@@ -89,7 +89,10 @@ def is_docstring_directive_enable(docstring):
     :rtype: bool
     """
     result = get_docstring_directive(docstring)
-    return result == 'enable'
+    if 'enable' in result:
```
clebergnu: The names begin to make less sense with this change.  I can certainly live with the current ones, but reading this code would be eaiser on **my mind** (YMMV) if this method was now called `docstring_directives_has_enable()`.

The implementation (suggestion) can also look like:

```python 
return 'enable' in get_docstring_directives(docstring)
```

selftests/unit/test_safeloader.py
```diff
@@ -80,6 +82,7 @@ def test_newlines(self):
     def test_enabled(self):
         self.assertTrue(safeloader.is_docstring_directive_enable(":avocado: enable"))
         self.assertTrue(safeloader.is_docstring_directive_enable(":avocado:\tenable"))
+        self.assertTrue(safeloader.is_docstring_directive_enable(":avocado: enable\n:avocado: tags=fast"))
```
clebergnu: Suggestion for test (and possibly documentation):  what happens if this is given:

```
   :avocado: enable
   :avocado: disable
   :avocado: tags=fast
```

avocado/core/safeloader.py
apahim: Well, as you don't mind about the functions names, let me propose something I believe is more efficient.

selftests/unit/test_safeloader.py
apahim: ok

PR #1923, opened 2017-04-13
fix output on unknown tests

apahim:
When the tests references cannot be resolved to a test_suite, the
results plugins should not output anything.

Reference: https://trello.com/c/oMQsQHC6

### DIFF
--- a/selftests/functional/test_output.py
+++ b/selftests/functional/test_output.py
@@ -459,6 +459,15 @@ class OutputPluginTest(unittest.TestCase):
         self.assertIn("not found", result.stderr)
         self.assertNotIn("Avocado crashed", result.stderr)
 
+    def test_results_plugins_no_tests(self):
+        os.chdir(basedir)
+        for output in ['json', 'xunit', 'tap']:
+            cmd_line = ("%s run UNEXISTING --job-results-dir %s --%s -"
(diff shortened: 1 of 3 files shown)


apahim:
[BUG: Result plugins generating output when job does not have a test suite](https://trello.com/c/oMQsQHC6/960-bug-result-plugins-generating-output-when-job-does-not-have-a-test-suite)

selftests/functional/test_output.py
```diff
@@ -459,6 +459,15 @@ def test_broken_pipe(self):
         self.assertIn("not found", result.stderr)
         self.assertNotIn("Avocado crashed", result.stderr)
 
+    def test_results_plugins_no_tests(self):
+        os.chdir(basedir)
+        for output in ['json', 'xunit', 'tap']:
+            cmd_line = ("%s run UNEXISTING --job-results-dir %s --%s -"
+                        % (AVOCADO, self.tmpdir, output))
+            result = process.run(cmd_line, ignore_status=True)
```
ldoktor: We do generate all of the results by default, how about running the command just once and then checking all results?

apahim: I'm testing stdout here. If you want me to test the files (non)creation, I can do that for sure.

PR #1927, opened 2017-04-18
fix output on unknown tests [V2]

apahim:
v2:
- Test results files (non)creation instead of the avocado command (non)stdout.

v1: #1923


apahim:
[BUG: Result plugins generating output when job does not have a test suite](https://trello.com/c/oMQsQHC6/960-bug-result-plugins-generating-output-when-job-does-not-have-a-test-suite)

PR #1937, opened 2017-04-20
resultsdb plugin

apahim:
Propagate tests results to resultsdb.

Reference: https://trello.com/c/AcruJdgS/906-result-plugin-for-resultsdb
Signed-off-by: Amador Pahim <apahim@redhat.com>

PR #1938, opened 2017-04-20
resultsdb plugin [v2]

apahim:
v2:
- Docs.
- Config option.

v1: #1937


apahim:
[Result plugin for ResultsDB](https://trello.com/c/AcruJdgS/906-result-plugin-for-resultsdb)

etc/avocado/avocado.conf
```diff
@@ -83,3 +83,10 @@ skip_broken_plugin_notification = []
 # options or test types).
 # The keyword "@DEFAULT" will be replaced with all available unused loaders.
 loaders = ['file', '@DEFAULT']
+
+# Options that will only have effect when the
+# corresponding optional plugin is installed
+[optional.plugins]
+# Plugin: resultsdb
+# URL of the resultsdb API
+# resultsdb_api = http://resultsdb.example.com/api/v2.0/
```
clebergnu: Ideally, there will be no references to "optional" things inside the main (core) avocado configuration file.  That's why the drop-in (avocado.conf.d) exist.  It'll certainly make the deployment of the optional plugin a lit bit more complex, but I think it's the way to go.  It also matches the approach used with vt, vt-joblock, etc.

clebergnu: Anyway, I'm just giving quick comments here (maybe you want to wait until I can do full review).

apahim: Ah, sure, makes sense.

apahim: > Anyway, I'm just giving quick comments here (maybe you want to wait until I can do full review).

Sorry, I will hold myself this time :D

clebergnu: I was trying to avoid unecessary work from you (sending new versions with smaller changes) :)

docs/source/optional_plugins/index.rst
```diff
@@ -0,0 +1,12 @@
+.. index file for optional plugins
+
+================
+Optional Plugins
+================
+
+The following pages are the doccumentation for the optinal plugins:
```
clebergnu: Typo: doccumentation

optional_plugins/resultsdb/avocado_resultsdb/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+        Returns the resultsdb corresponding status to Avocado status.
+        Valid statuses are:
+         PASSED
+         FAILED
+         INFO
+         NEEDS_INSPECTION
+        """
+        mapping = {'PASS': 'PASSED',
+                   'FAIL': 'FAILED',
+                   'ERROR': 'FAILED',
```
ldoktor: Shouldn't ERROR be NEEDS_INSPECTION?

optional_plugins/resultsdb/avocado_resultsdb/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+                   'INTERRUPTED': 'FAILED',
```
ldoktor: The same here

optional_plugins/resultsdb/avocado_resultsdb/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+                   'SKIP': 'INFO',
+                   'CANCEL': 'INFO',
```
ldoktor: ABORTED?

optional_plugins/resultsdb/avocado_resultsdb/__init__.py
ldoktor: OK, There are multiple versions and the 2nd version does not support ERROR/INTERRUPTED/... Then I'm fine with your mapping except of the `ERROR` and `INTERRUPTED` which should be `NEEDS_INSPECTION`.

Actually I'd just cover everything but `NEEDS_INSPECTION` and leave the rest to the default.

ldoktor:
@apahim, very nice, only if I understand it correctly, restultdb allows updating the results and I think it's make sense to have this as a streamline plugin (without it one can simply publish xunit/tap and publish it), what do you think about this? I think it shouldn't be that hard and people would benefit from it (eg. on host crash)

optional_plugins/resultsdb/avocado_resultsdb/__init__.py
apahim: Sure. Indeed ERROR/INTERRUPTED should map to NEEDS_INSPECTION.

docs/source/optional_plugins/index.rst
apahim: ok. also 'optinal' :)

apahim:
@ldoktor yep, streamline plugin is totally feasible.  `resultdb` does not have `update_results`, but we can create the `group`(job) in  `pre_tests()` and create the result in `end_test()`. Sending a V3,

PR #1941, opened 2017-04-21
resultsdb plugin [v3]

apahim:
v3:
- Move config option to a new file inside conf.d
- Make the plugin a streamline plugin.
- Introduce new option to publish the link to the logs (--resultsdb-logs).
- Adjust the status mappings.

v2: #1938 
- Docs.
- Config option.

v1: #1937


apahim:
[Result plugin for ResultsDB](https://trello.com/c/AcruJdgS/906-result-plugin-for-resultsdb)

optional_plugins/resultsdb/avocado_resultsdb/__init__.py
```diff
@@ -0,0 +1,172 @@
...
+        note = None
+        if state['fail_reason'] is not None:
+            note = str(state['fail_reason'])
+
+        ref_url = None
+        if self.rdblogs is not None:
+            logdir = os.path.basename(state['logdir'])
+            ref_url = '%s/%s/test-results/%s' % (self.rdblogs,
+                                                 self.job_logdir,
+                                                 logdir)
```
ldoktor: OK, so the `rdblogs` will be appended with `$job_logdir/test-results/$logdir`. This is quite important information and it'd be nice to have it in the documentation. (I was wondering how resultsdb assigns logs based on the id or so...)

optional_plugins/resultsdb/avocado_resultsdb/__init__.py
```diff
@@ -0,0 +1,172 @@
...
+
+    @staticmethod
+    def _status_map(status):
+        """
+        Returns the resultsdb corresponding status to Avocado status.
+        Valid statuses are:
+         PASSED
+         FAILED
+         INFO
+         NEEDS_INSPECTION
```
ldoktor: This is hideously long, how about putting it into a single line? `PASSED/FAILED/INFO/NEEDSS_INTROSPECTION`. Also it might be useful to mention on several places that this is based on resultsdb-2.0 as resultsdb-1.0 supported other statuses and the future versions might also be more-less incompatible...

apahim: I like it long, but I will improve this docstring.

optional_plugins/resultsdb/avocado_resultsdb/__init__.py
apahim: Indeed. Adding to the docs.

optional_plugins/resultsdb/avocado_resultsdb/__init__.py
clebergnu: @ldoktor I know this was a suggestion, but as another kind of *suggestion*, let me say: we should try to avoid bringing up issues that are solely related to style preferences (and that have no standard previously discussed and defined).  IMO, it acts against the optimal flow of development and review.

optional_plugins/resultsdb/avocado_resultsdb/__init__.py
```diff
@@ -0,0 +1,172 @@
...
+            return
+
+        self.job_id = job.unique_id
+        self.job_logdir = os.path.basename(job.logdir)
+
+        ref_url = None
+        if self.rdblogs is not None:
+            ref_url = '%s/%s' % (self.rdblogs, self.job_logdir)
+
+        self.rdbapi.create_group(self.job_id, ref_url)
```
ldoktor: The group supports a description, what would you say about using the `sys.argv` or at least the urls? I think it could help improving the experience https://taskotron.fedoraproject.org/resultsdb/groups

apahim: No, I don't think we should do that. The command line or the references are not enough to describe a job properly. I prefer to keep this as is.

ldoktor: Sure, no problem with that and we can always tweak it later.

PR #1956, opened 2017-04-25
Added a sleep time after starting multipath services in multipath.py

Naresh-ibm:
Test which are using this utility are failing as they start using
multipath as soon as services are re-satrted. But multipath need
some time to load the services completely. So to load the
multipaths properly we need to give some sleep time which resolve
the issue

### DIFF
--- a/avocado/utils/multipath.py
+++ b/avocado/utils/multipath.py
@@ -58,6 +58,7 @@ def form_conf_mpath_file(blacklist="", defaults_extra=""):
     # multipath.conf file to take effect.
     time.sleep(5)
     service.SpecificServiceManager(get_svc_name()).restart()
+    time.sleep(4)
 
 
 def device_exists(path):


avocado/utils/multipath.py
```diff
@@ -58,6 +58,7 @@ def form_conf_mpath_file(blacklist="", defaults_extra=""):
     # multipath.conf file to take effect.
     time.sleep(5)
     service.SpecificServiceManager(get_svc_name()).restart()
+    time.sleep(4)
```
apahim: So, in the end you need a sleep **after** the service restart. Maybe you can remove the sleep `5` before?

Naresh-ibm: @apahim we need the sleep() at both the places, first one is for multipath.conf file( as it is commented)and second sleep() is for multipath services.

apahim: That's annoying :)

PR #1962, opened 2017-04-26
skip/cancel: internal improvements

apahim:

- The test.SkipTest class was renamed to test.FakeTest to be even more
  generic, intending to be overridden by subclasses that make the test
  test.FakeTest class will not SKIP the test if used directly anymore.

  setUp().

- The skip exceptions were simplified and the only missing piece is to
  remove the exceptions.TestSetupSkip, when the time to drop the
  self.skip() comes.

### DIFF
--- a/avocado/core/test.py
+++ b/avocado/core/test.py
@@ -40,6 +40,7 @@ from ..utils import genio
 from ..utils import path as utils_path
 from ..utils import process
 from ..utils import stacktrace
+from .decorators import skip
 from .settings import settings
 from .version import VERSION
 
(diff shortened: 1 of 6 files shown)


apahim:
[Rename `SkipTimeout` to `CancelTimeout`](https://trello.com/c/7Xwcpl8V/990-rename-skiptimeout-to-canceltimeout)

avocado/core/test.py
```diff
@@ -961,15 +957,14 @@ def test(self):
         raise exceptions.NotATestError(e_msg)
 
 
-class SkipTest(Test):
+class FakeTest(Test):
```
ldoktor: I thought we wanted to completely avoid using `Test` class as it's inappropriate. How about creating `BaseTest` class with only the essential stuff (without `data_dir` and such) and use it as a base for `Test` and other dummy classes?

ldoktor: Btw the `MissingTest` and such can safely be simply inherited from `object` as they are never executed. Anyway I'd prefer if you did not touch those (MissingTest and NotATest) as I'll be moving them (and inheriting them from object) in loaders cleanup.

ldoktor: As for the name, have you considered `MockTest`? `Fake` for me means something like fraud, but `Mock` is more about replacing self instead of something. Anyway I'm not native, so don't know...

ldoktor: Actually these are actual tests, so please ignore the first comment. Also no need to change anything in the second comment as I'll be tackling this in my PR so just let me know about the naming suggestion, anyway it's really minor as I'm not native and my perceiving can be wrong.

apahim: Yup, makes sense. Mock > Fake, in this case.

apahim:
Sure, sending in a couple of minutes.

PR #1966, opened 2017-04-27
Skip internal improvements [v3]

apahim:

- The test.SkipTest class was renamed to test.MockingTest to be even more
  generic, intending to be overridden by sub-classes that make the test
  test.MockingTest class will not SKIP the test if used directly anymore.

  setUp().

- The skip exceptions were cleaned-up the avocado.test exception
  handler was simplified. The only missing piece is to  remove the
  exceptions.TestSetupSkip, when the time to drop the self.skip()
  comes.

---

v1: #1962 

v2: #1964 
- Rename `FakeTest` to `MockingTest`.

v3:
- Don't rename exceptions.TestSkipError.
- Split the change in two commits.

### DIFF
--- a/avocado/core/decorators.py
+++ b/avocado/core/decorators.py
@@ -63,7 +63,7 @@ def skip(message=None):
         if not isinstance(function, type):
             @wraps(function)
             def wrapper(*args, **kwargs):
-                raise core_exceptions.TestDecoratorSkip(message)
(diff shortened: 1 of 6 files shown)


apahim:
[Rename `SkipTimeout` to `CancelTimeout`](https://trello.com/c/7Xwcpl8V/990-rename-skiptimeout-to-canceltimeout)

PR #1967, opened 2017-04-27
html report: include params and split Test ID

apahim:
- Include `Params` as a test name tooltip.
- Split Test ID in UID, Test Name and Variant, so users can order by any of them individually.

### DIFF
--- a/optional_plugins/html/avocado_result_html/__init__.py
+++ b/optional_plugins/html/avocado_result_html/__init__.py
@@ -120,6 +120,10 @@ class ReportModel(object):
         for tst in self.result.tests:
             formatted = {}
             formatted['name'] = tst['name']
+            params = 'Params:\n'
+            for path, key, value in tst['params'].iteritems():
+                params += '  %s:%s => %s\n' % (path, key, value)
+            formatted['params'] = params
             formatted['status'] = tst['status']
             logdir = os.path.join(results_dir, 'test-results', tst['logdir'])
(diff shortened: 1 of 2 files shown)


apahim:
[Separate test-name and variant-id in html results](https://trello.com/c/qz3pHIZq/944-separate-test-name-and-variant-id-in-html-results)

optional_plugins/html/avocado_result_html/__init__.py
```diff
@@ -119,11 +119,13 @@ def tests(self):
         results_dir = self.results_dir(False)
         for tst in self.result.tests:
             formatted = {}
-            formatted['name'] = tst['name']
+            formatted['uid'] = tst['name'].uid
```
clebergnu: One difference I spotted here is: one the test ID shown on the UI (and on the previous HTML report) we have a "0 padded" number, such as `01`.  With this change, we now have `1`.

Not sure if you caught that, and I'm also not sure if that is a problem.

optional_plugins/html/avocado_result_html/resources/templates/report.mustache
```diff
@@ -68,7 +68,7 @@
             {{#tests}}
                 <tr class="{{row_class}}">
                     <td>{{time_start}}</td>
-                    <td><a href="{{logdir}}">{{name}}</a></td>
+                    <td><a href="{{logdir}}" title="{{params}}">{{name}}</a></td>
```
clebergnu: This is super useful! Well done!

PR #1972, opened 2017-05-03
Support tags in test methods

apahim:
- Refactor `loader._find_avocado_tests()` to reduce code repetition.
- Add support for tags in test methods:
  - Tags are now related to the method and not to the class.
  - Methods will inherit the tags from the class.

### DIFF
--- a/avocado/core/loader.py
+++ b/avocado/core/loader.py
@@ -32,7 +32,6 @@ from . import test
 from . import safeloader
 from ..utils import path
 from ..utils import stacktrace
-from ..utils import data_structures
 from .settings import settings
 
 #: Show default tests (for execution)
@@ -629,12 +628,12 @@ class FileLoader(TestLoader):
                 if safeloader.check_docstring_directive(docstring, 'disable'):
                     continue
 
-                tags = safeloader.get_docstring_directives_tags(docstring)
+                cl_tags = safeloader.get_docstring_directives_tags(docstring)
 
                 if safeloader.check_docstring_directive(docstring, 'enable'):
-                    methods = self._get_methods(statement.body)
-                    result[statement.name] = {'methods': methods,
-                                              'tags': tags}
+                    info = self._get_methods_info(statement.body,
+                                                  cl_tags)
+                    result[statement.name] = info
                     continue
 
                 if test_import:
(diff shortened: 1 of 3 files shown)


apahim:
[Support docstring directives in test methods](https://trello.com/c/DYRJOIkO/957-support-docstring-directives-in-test-methods)

avocado/core/loader.py
```diff
@@ -658,16 +653,20 @@ def _find_avocado_tests(self, path):
...
         return result
 
+    @staticmethod
+    def _get_methods(statement_body):
+        methods = [st.name for st in statement_body if
+                   isinstance(st, ast.FunctionDef) and
+                   st.name.startswith('test')]
+        methods = data_structures.ordered_list_unique(methods)
+        return methods
+
```
clebergnu: OK!

avocado/core/loader.py
```diff
@@ -677,7 +681,7 @@ def _make_avocado_tests(self, test_path, make_broken, subtests_filter,
                 test_factories = []
                 for test_class, info in tests.items():
                     if isinstance(test_class, str):
-                        for test_method in info['methods']:
+                        for test_method in sorted(info):
```
clebergnu: There's a big problem here.  We don't want the test methods ordered.  We want to keep the same order that was found in the source code.  For the following class:

```python
from avocado import Test

class MyClass(Test):
    def test2(self):
        pass

    def testA(self):
        pass
    
    def test1(self):
        pass

    def testZZZ(self):
        pass
    
    def test(self):
        pass

```

The original behavior is:

```
avocado list foo.py
INSTRUMENTED /tmp/foo.py:MyClass.test2
INSTRUMENTED /tmp/foo.py:MyClass.testA
INSTRUMENTED /tmp/foo.py:MyClass.test1
INSTRUMENTED /tmp/foo.py:MyClass.testZZZ
INSTRUMENTED /tmp/foo.py:MyClass.test
```

With this, it becomes:

```
INSTRUMENTED /tmp/foo.py:MyClass.test
INSTRUMENTED /tmp/foo.py:MyClass.test1
INSTRUMENTED /tmp/foo.py:MyClass.test2
INSTRUMENTED /tmp/foo.py:MyClass.testA
INSTRUMENTED /tmp/foo.py:MyClass.testZZZ
```

apahim: oh, ok, missed that. V2 on the way.

PR #1974, opened 2017-05-03
mux: show total number of variants

apahim:
Now that the variant IDs are not a sequential number, we need this extra
information so users can easily know how many variants were generated.

Reference: https://trello.com/c/N7pB9E4O


apahim:
[Add a total count to `avocado multiplex`](https://trello.com/c/N7pB9E4O/1015-add-a-total-count-to-avocado-multiplex)

PR #1979, opened 2017-05-04
 Support tags in test methods [v2]

apahim:
V1: #1972
- Refactor `loader._find_avocado_tests()` to reduce code repetition.
- Add support for tags in test methods:
  - Tags are now related to the method and not to the class.
  - Methods will inherit the tags from the class.

V2:
- Use an ordered structure to keep the methods in the order that they are discovered in.
- Include a selftest to make sure the order is maintained.

### DIFF
--- a/selftests/unit/test_loader.py
+++ b/selftests/unit/test_loader.py
@@ -61,12 +61,18 @@ class DisabledTest(Test):
 
 class FastTest(Test):
     '''
-    :avocado: tags=fast,net
+    :avocado: tags=fast
     '''
     def test_fast(self):
+        '''
+        :avocado: tags=net
+        '''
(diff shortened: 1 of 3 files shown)


apahim:
[Support docstring directives in test methods](https://trello.com/c/DYRJOIkO/957-support-docstring-directives-in-test-methods)

selftests/unit/test_loader.py
```diff
@@ -463,6 +489,17 @@ def test_filter_tags(self):
                                                 'does,not,exist'])
             self.assertEqual(len(filtered), 0)
 
+    def test_methods_order(self):
```
clebergnu: Nice!

PR #1983, opened 2017-05-06
Extended vendors_map dict for Power8

vrbagalkote:
Added a new item for vendors_map dic to def get_cpu_vendor_name():

Signed-off-by: Venkat R B <vrbagal1@linux.vnet.ibm.com>

### DIFF
--- a/avocado/utils/cpu.py
+++ b/avocado/utils/cpu.py
@@ -78,7 +78,8 @@ def get_cpu_vendor_name():
     vendors_map = {
         'intel': ("GenuineIntel", ),
         'amd': ("AMD", ),
-        'power7': ("POWER7", )
+        'power7': ("POWER7", ),
+        'power8': ("POWER8", )
     }
 
     cpu_info = _get_cpu_info()


vrbagalkote:
Smaple O/P:

>>> ven = cpu.get_cpu_vendor_name()
>>> print ven
power8

avocado/utils/cpu.py
```diff
@@ -78,7 +78,8 @@ def get_cpu_vendor_name():
     vendors_map = {
         'intel': ("GenuineIntel", ),
         'amd': ("AMD", ),
-        'power7': ("POWER7", )
+        'power7': ("POWER7", ),
+        'power8': ("POWER8", )
```
ldoktor: Yep, works well even for `POWER8E`...

vrbagalkote:
Actually, I realised now that it is appropriate to add "ibm" as vendor than power7, power8 etc. For all power arch, IBM is the vendor. I just did a force push, but then I saw, its already merged. Can I send other patch changing it correctly. Because, for power6 and below this function will return none.

vrbagalkote:
This will work for everything, I guess.

'ibm': ("POWER", )

ldoktor:
This is hard to decide and it needs a bit more discussion (@sathnaga, @PandaWei, @Hao-Liu, @luckyh).

First issue is that some existing code can depend on the returned value being `power*`, so guys, please let us know whether you actually use this function in your projects and whether changing that value won't break everything (fgrep didn't found anything in avocado, avocado-vt and avocado-misc-tests).

Second issue is that for GenuineIntel and AMD it's pretty clear who is the vendor, but is it true for POWER? I'm not entirely sure, but there is the openpower foundation. Anyway I don't know the details so anyone who knows more please share the links...

Last question, is this function in use? Perhaps we can simply remove it....

ldoktor:
@vrbagalkote can you please send this suggestion to avocado-devel mailing list? It makes sense to me but apparently GH is not the right place to discuss it...

ldoktor:
[Do we want to rework the purpose of `avocado.utils.cpu.get_cpu_vendor_name()`?](https://trello.com/c/uOVhvv5o/1040-do-we-want-to-rework-the-purpose-of-avocado-utils-cpu-get-cpu-vendor-name)

luckyh:
Um, I didn't find anything depended on this in avocado-vt or tp-qemu. To me, I'd like to request a more powerful interface that not only reports vendor name but also gives the rest information of `/proc/cpuinfo`.

For example:
```python
cpuinfo = get_cpu_info()
print cpuinfo.arch
print cpuinfo.vendor_id
print cpuinfo.model_name
print cpuinfo.nr_processor
...
```

ldoktor:
Well the problem with this is that some of your info strictly relates to CPU and on heterogeneous systems hardly depend on which CPU you pick.

As for the `nr_processor` one should use: `import multiprocessing; multiprocessing.cpu_count()` or recently another method was added in `cpu.total_cpus_count`.

Anyway I'm wondering what do you need those info for? In most cases I only need `arch`, `nr_processors` or full listing of `/proc/cpuinfo` to see where the test ran. (I'm just wondering whether you are not trying to over-engineer it)

PR #2011, opened 2017-05-17
test: always execute tearDown()

apahim:
Currently, when something goes wrong in setUp(), tearDown() is not
executed. This patch changes the behaviour of our Test class, making the
tearDown() to be always executed, regardless what happens in setUp().

Reference: https://trello.com/c/zSZ4xFwp

### DIFF
--- a/avocado/core/test.py
+++ b/avocado/core/test.py
@@ -563,7 +563,6 @@ class Test(unittest.TestCase):
         stdout_check_exception = None
         stderr_check_exception = None
         skip_test = getattr(testMethod, '__skip_test_decorator__', False)
-        cancel_test = False
         try:
             if skip_test is False:
                 self.setUp()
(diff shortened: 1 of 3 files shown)


apahim:
[Run `tearDown` (almost) always when `setUp` was executed.](https://trello.com/c/zSZ4xFwp/970-run-teardown-almost-always-when-setup-was-executed)

docs/source/WritingTests.rst
```diff
@@ -520,6 +520,11 @@ If you need to perform setup actions before/after your test, you may do so
 in the ``setUp`` and ``tearDown`` methods, respectively. We'll give examples
 in the following section.
 
+The ``teadDown`` method is always executed (unless the test is skipped).
```
ldoktor: typo `tearDown`

selftests/functional/test_test.py
```diff
@@ -0,0 +1,69 @@
...
+AVOCADO = os.environ.get("UNITTEST_AVOCADO_CMD", "./scripts/avocado")
+
+TEST_SETUP_EXCEPTION = """
+import avocado
+
+class AvocadoTest(avocado.Test):
+
+    def setUp(self):
+        self.log.info('setup code before')
+        raise
```
ldoktor: well this works, but is not really a valid way to raise exceptions... (I know, it's just a selftest dummy test, but still people are learning wrong patterns...)

ldoktor: Also this is very similar to existing test defined in `functional/test_canceltests.py`, how about making this test generic, putting it directly in `selftests/.data/` and reusing it everywhere? Also this selftest does not cover all possible situation and if we are (and I think we should) to test all variants, we'd benefit from a generic test which raises/cancels on various places logging messages all around.

I imagine at least:

```yaml
location: !mux
    import:
    setUp:
    test:
    tearDown:
exception: !mux
    self.skip:
    self.cancel:
    self.fail:
    exception:
```

And this test can replace some of the `test_basic` ones as it checks all the statuses (if we include `self.log.warn(...)` which would also make sense). and some of the `skip` and `cancel` tests. Putting it all to one place would IMO make sense and there is no need to create the test file for every variant, simply use `exc = self.params.get("setUpFailure"); if exc: ...`.

avocado/core/test.py
```diff
@@ -563,47 +563,45 @@ def _run_avocado(self):
...
+            except:  # Old-style exceptions are not inherited from Exception()
+                stacktrace.log_exc_info(sys.exc_info(), logger=LOG_JOB)
+                details = sys.exc_info()[1]
+                if not isinstance(details, Exception):  # Avoid passing nasty exc
+                    details = exceptions.TestError("%r: %s" % (details, details))
+                test_exception = details
+                self.log.debug("Local variables:")
+                local_vars = inspect.trace()[1][0].f_locals
+                for key, value in local_vars.iteritems():
+                    self.log.debug(' -> %s %s: %s', key, type(value), value)
```
ldoktor: The code looks good ...

docs/source/WritingTests.rst
ldoktor: Actually I'd rather re-phrase it in order to make sense as it talks about next section and then you add another sentence which makes at least me a bit confused. How about:

_To perform setup actions before/after your test, you may use ``setUp`` and ``tearDown`` methods. The ``tearDown`` method is always executed even on ``setUp`` failure so don't forget to initialize your variables early in the ``setUp``. Example of usage is in the next section :doc:`Running third party...`_

apahim: ok, sure

avocado/core/test.py
apahim: cool

selftests/functional/test_test.py
apahim: You mean something like this (WiP)?
https://github.com/apahim/avocado/commit/ff23ccb2b6ad0d53067249a9379698ebd7fa910e

ldoktor: exactly

PR #2012, opened 2017-05-17
Be lenient with test references

apahim:
Introducing the --force-execution to create the Job regardless the fact
that some test references might not be resolved to tests.

Reference: https://trello.com/c/bDUJjhq0

### DIFF
--- a/avocado/core/loader.py
+++ b/avocado/core/loader.py
@@ -233,7 +233,7 @@ class TestLoaderProxy(object):
     def get_decorator_mapping(self):
         return self._decorator_mapping
 
-    def discover(self, references, which_tests=DEFAULT):
+    def discover(self, references, which_tests=DEFAULT, force=None):
         """
         Discover (possible) tests from test references.
 
@@ -278,8 +278,12 @@ class TestLoaderProxy(object):
                 tests.extend([(test.MissingTest, {'name': reference})
                               for reference in unhandled_references])
             else:
-                raise LoaderUnhandledReferenceError(unhandled_references,
-                                                    self._initialized_plugins)
+                if force == 'on':
+                    LOG_UI.error(LoaderUnhandledReferenceError(unhandled_references,
+                                                               self._initialized_plugins))
+                else:
+                    raise LoaderUnhandledReferenceError(unhandled_references,
+                                                        self._initialized_plugins)
         return tests
 
     def load_test(self, test_factory):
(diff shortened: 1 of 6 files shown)


apahim:
[Be lenient with test references](https://trello.com/c/bDUJjhq0/1006-be-lenient-with-test-references)

avocado/core/loader.py
```diff
@@ -233,7 +233,7 @@ def get_type_label_mapping(self):
     def get_decorator_mapping(self):
         return self._decorator_mapping
 
-    def discover(self, references, which_tests=DEFAULT):
+    def discover(self, references, which_tests=DEFAULT, force=None):
```
clebergnu: Since the other parameters are documented, it's a good idea to also document this one.

avocado/plugins/run.py
```diff
@@ -85,6 +85,10 @@ def configure(self, parser):
                             default='off', help='Keep job temporary files '
                             '(useful for avocado debugging). Defaults to off.')
 
+        parser.add_argument('--force-execution', choices=('on', 'off'),
```
clebergnu: IMO the description is accurate, but the option name is too generic.  We may want to force execution on other different conditions later.  I'd go with something like `--ignore-resolution-errors`.

avocado/core/loader.py
apahim: Yes, sure.

avocado/plugins/run.py
apahim: What about `--ignore-resolver-errors` or `--ignore-loader-errors`?

ldoktor: From my side it's `skip-missing` or `ignore-missing` or even more verbose `ignore-missing-references` or `ignore-missing-tests`.

selftests/functional/test_basic.py
```diff
@@ -190,6 +190,18 @@ def test_runner_failfast(self):
         self.assertEqual(result.exit_status, expected_rc,
                          "Avocado did not return rc %d:\n%s" % (expected_rc, result))
 
+    def test_runner_force_execution(self):
+        os.chdir(basedir)
+        cmd_line = ('%s run --sysinfo=off --job-results-dir %s '
+                    'passtest.py badtest.py --force-execution on'
```
ldoktor: I'd also suggest adding test where all tests are bad to see we don't change the behavior (should not be executed, right?)

avocado/plugins/run.py
apahim: I'd not like to use 'skip' here because we have the `SKIP` status and it might confuse people. I think `ignore-missing-references` makes a lot of sense though.

selftests/functional/test_basic.py

PR #2029, opened 2017-05-23
 test: always execute tearDown() [v2]

apahim:
v2:
- Improved docs.
- Improved selftets.

v1: #2011
Currently, when something goes wrong in setUp(), tearDown() is not
executed. This patch changes the behaviour of our Test class, making the
tearDown() to be always executed, regardless what happens in setUp().


apahim:
@ldoktor, @clebergnu, I don't see many tests that can be replaced by this new one. Anyway, this one is increasing the coverage a lot.
@ldoktor, about including code to execute at `import`, I'm not convinced about the benefits. All the commands we have are going to break the test on import, so I don't see a reason to test that. Anyway, with this new test we have a good foundation for further improvements.

apahim:
[Run `tearDown` (almost) always when `setUp` was executed.](https://trello.com/c/zSZ4xFwp/970-run-teardown-almost-always-when-setup-was-executed)

selftests/functional/test_statuses.py
```diff
@@ -0,0 +1,291 @@
...
+basedir = os.path.abspath(basedir)
+
+AVOCADO = os.environ.get("UNITTEST_AVOCADO_CMD", "./scripts/avocado")
+
+TEST_CONTENT = """
+import os
+import sys
+from avocado import Test
+
+class StatusTest(Test):
```
ldoktor: Would you please re-consider putting this into `selftests/.data`? It is really useful test for development and that way one would be able to execute it while working on some feature.

apahim: Sure, I put here since no one else is using it. But I can put there for sure.

selftests/functional/test_statuses.py
```diff
@@ -0,0 +1,291 @@
...
+        command: self.cancel()
+    fail:
+        command: self.fail()
+    warn:
+        command: self.log.warn('')
+    exception:
+        command: raise Exception
+    exit:
+        command: sys.exit(-1)
+    kill:
```
ldoktor: I think this is quite an extreme and it adds 10s, so how about using:

```
        !filter-only : /run/locations/test
```

to get the coverage, but only check the `test` variant? (so 10s vs. 30s)

selftests/functional/test_statuses.py
```diff
@@ -0,0 +1,291 @@
...
+EXPECTED_RESULTS = {'skip-setup-d304': ('SKIP',
+                                        ['setup pre',
+                                         'teardown pre',
+                                         'teardown post',
+                                         "[WARNING: self.skip() will be "
+                                         "deprecated. Use 'self.cancel()' "
+                                         "or the skip decorators]"],
+                                        ['setup post',
+                                         'test pre',
+                                         'test post']),
```
ldoktor: well this could be deducted from `all - expected`, but it is only a minor improvement so just consider it...

ldoktor: actually most things can be deducted from the `(\w+)-(\w+).*`:

```python
if match.group(1) == "setup":
    exp = ["setup pre", "teardown pre", "teardown post"]
    if match.group(0) == "skip":
        exp.append("[WARNING: ...]")
elif ...
```
anyway this is easy to read so I'm fine either way.

selftests/functional/test_statuses.py
```diff
@@ -0,0 +1,291 @@
...
+                         (expected[0], test['status'], variant))
+        debug_log = open(test['logfile'], 'r').read()
+        for msg in expected[1]:
+            self.assertIn(msg, debug_log,
+                          "Message '%s' should be in the log (%s)" %
+                          (msg, variant))
+        for msg in expected[2]:
+            self.assertNotIn(msg, debug_log,
+                             "Message '%s' should not be in the log (%s)" %
+                             (msg, variant))
```
ldoktor: would you please provide all necessary details in case of failure? I'd expect the json output and job.log to be able to tell what was actually going on.

selftests/functional/test_statuses.py
apahim: The second comment feels like http://devopsreactions.tumblr.com/post/150439088410/overengineering :-)
I will follow the suggestion in the first comment though. It will remove some code keeping the readability. Thank you.

selftests/functional/test_statuses.py
apahim: sure

apahim: Do you think the test debug.log and the test section from the json results are enough? They seem enough to me. That way we don't need to create extra files, just append those into the message.

ldoktor: yep, that's the idea.

selftests/functional/test_statuses.py
ldoktor: yep, the first one will be welcome, the second one is a bit overengineered...

selftests/functional/test_statuses.py
ldoktor: thank you, I promise I'll use it (at least once 😄)

PR #2053, opened 2017-06-02
 Recursive Test Discovery [V2]

apahim:
v2:
- Improved docs addressing v1 comments.

v1: #2052 


apahim:
[Inherit `test*` methods from parents](https://trello.com/c/GtFlhcOj/1011-inherit-%60test*%60-methods-from-parents)

avocado/core/loader.py
```diff
@@ -613,7 +613,7 @@ def _find_avocado_tests(self, path):
         # The name used, in case of 'import avocado as avocadolib'
         mod_import_name = None
         # The resulting test classes
-        result = {}
+        result = collections.OrderedDict()
```
ldoktor: I don't think we should define the order of discovered methods and if we do want to do it, I'd go with alphabetical order, which is used by unittest.

avocado/core/loader.py
```diff
@@ -708,16 +711,19 @@ def _make_avocado_tests(self, test_path, make_broken, subtests_filter,
             tests = self._find_avocado_tests(test_path)
             if tests:
                 test_factories = []
-                for test_class, info in tests.items():
+                for test_class in tests:
```
ldoktor: well `iteritems()` would still be better here to avoid the need to look for `tests[test_class]` inside the branches... Alternatively you can get the path once on line 716 before the for loop.

avocado/core/loader.py
```diff
@@ -645,18 +645,24 @@ def _find_avocado_tests(self, path, class_name=None):
 
             # Looking for a 'class Anything(anything):'
             elif isinstance(statement, ast.ClassDef):
+
```
ldoktor: I think the commit message is wrong (at least there is a typo, but I think the `disable` behavior changed...)

avocado/core/loader.py
```diff
@@ -684,6 +690,77 @@ def _find_avocado_tests(self, path, class_name=None):
                                                                       cl_tags),
                                     'path': path}
                             result[statement.name] = info
+                            continue
+
+                # Looking for the 'recursive' docstring or a class_name (what
+                # means we are under recursion)
+                if (safeloader.check_docstring_directive(docstring, 'recursive')
```
ldoktor: The recursive should take precedence to `{mod,test}_import`, because you can have multiple inheritance and still inherit from Test.

ldoktor:
```diff
diff --git a/avocado/core/loader.py b/avocado/core/loader.py
index 88eb138..d475862 100644
--- a/avocado/core/loader.py
+++ b/avocado/core/loader.py
@@ -617,6 +617,8 @@ class FileLoader(TestLoader):
         # The resulting test classes
         result = collections.OrderedDict()
 
+        if os.path.isdir(path):
+            path = os.path.join(path, "__init__.py")
         mod = ast.parse(open(path).read(), path)
 
         for statement in mod.body:
@@ -669,29 +671,6 @@ class FileLoader(TestLoader):
                     result[statement.name] = info
                     continue
 
-                if test_import:
-                    base_ids = [base.id for base in statement.bases
-                                if hasattr(base, 'id')]
-                    # Looking for a 'class FooTest(Test):'
-                    if test_import_name in base_ids:
-                        info = {'methods': self._get_methods_info(statement.body,
-                                                                  cl_tags),
-                                'path': path}
-                        result[statement.name] = info
-                        continue
-
-                # Looking for a 'class FooTest(avocado.Test):'
-                if mod_import:
-                    for base in statement.bases:
-                        module = base.value.id
-                        klass = base.attr
-                        if module == mod_import_name and klass == 'Test':
-                            info = {'methods': self._get_methods_info(statement.body,
-                                                                      cl_tags),
-                                    'path': path}
-                            result[statement.name] = info
-                            continue
-
                 # Looking for the 'recursive' docstring or a class_name (what
                 # means we are under recursion)
                 if (safeloader.check_docstring_directive(docstring, 'recursive')
@@ -762,6 +741,29 @@ class FileLoader(TestLoader):
                                             result.update(res)
                     continue
 
+                if test_import:
+                    base_ids = [base.id for base in statement.bases
+                                if hasattr(base, 'id')]
+                    # Looking for a 'class FooTest(Test):'
+                    if test_import_name in base_ids:
+                        info = {'methods': self._get_methods_info(statement.body,
+                                                                  cl_tags),
+                                'path': path}
+                        result[statement.name] = info
+                        continue
+
+                # Looking for a 'class FooTest(avocado.Test):'
+                if mod_import:
+                    for base in statement.bases:
+                        module = base.value.id
+                        klass = base.attr
+                        if module == mod_import_name and klass == 'Test':
+                            info = {'methods': self._get_methods_info(statement.body,
+                                                                      cl_tags),
+                                    'path': path}
+                            result[statement.name] = info
+                            continue
+
         return result
 
     @staticmethod
diff --git a/selftests/unit/test_loader.py b/selftests/unit/test_loader.py
index 1cd93d7..a73886a 100644
--- a/selftests/unit/test_loader.py
+++ b/selftests/unit/test_loader.py
@@ -224,9 +224,10 @@ class SecondChild(FirstChild):
 """
 
 RECURSIVE_DISCOVERY_TEST2 = """
+from avocado import Test
 from recursive_discovery_test1 import SecondChild
 
-class ThirdChild(SecondChild):
+class ThirdChild(Test, SecondChild):
     '''
     :avocado: recursive
     '''
```

avocado/core/loader.py
apahim: Any reason to not use the discovery order, as we do for the class methods?

avocado/core/loader.py
apahim: yes, indeed.

avocado/core/loader.py
apahim: matter of preference, but I'm ok with both.

avocado/core/loader.py
apahim: ok, this will make the recursion to keep going after it reaches the 'base' class, but indeed solves the case you pointed out. thank you.

avocado/core/loader.py
```diff
@@ -684,6 +690,77 @@ def _find_avocado_tests(self, path, class_name=None):
...
+                        # Looking for a 'class FooTest(module.Parent)'
+                        if isinstance(parent, ast.Attribute):
+                            parent_class = parent.attr
+                        # Looking for a 'class FooTest(Parent)'
+                        else:
+                            parent_class = parent.id
+
+                        res = self._find_avocado_tests(path, parent_class)
+                        if res:
+                            parents.pop(0)
```
ldoktor: You don't want to remove 0-st item, but the current item, therefor `parents.remove(parent)`.

avocado/core/loader.py
ldoktor: First reason is people should not depend on the order of tests (they sometimes do and that is really dangerous). Second reason is it's not compatible with unittest, which can be confusing for users.

apahim: There's an order already. So the mentioned danger is already in place. But that order is not easily predictable without listing the tests, since it depends on the python dictionary ordering based on the hash of the elements. I'm proposing to take control of that order. First method that came to my mind was the same method used to order the test methods from a test class: discovery order. I'd be ok with alphabetical order, but that would not change the current state that much. So, let me drop this commit as it's not popular and not really relevant for this PR. We can discuss that later.

apahim:
HI @ldoktor, thanks for pointing the issues. Working on a V3.

PR #2068, opened 2017-06-14
docs: improve the docs regarding the test references

apahim:
Let's add some extra information and some examples about the test
references.

Reference: https://trello.com/c/iJlNcBaf

### DIFF
--- a/docs/source/GetStartedGuide.rst
+++ b/docs/source/GetStartedGuide.rst
@@ -211,6 +211,75 @@ keep reading.
           positional arguments, therefore test names (in case of
           ``avocado run``)
 
+Test References
+---------------
+
+A Test Reference is a string that can be resolved into (interpreted as)
+one or more tests by the Avocado Test Resolver. It is usually a path for
+a test file and it can be followed by a `:` and then a regular expression
+that will be used to filter the Test IDs that you're interested in.
+
+For instance, if you want to list all tests that are present in the
+`gdbtest.py` file, you can use the list command below::
+
+    $ avocado list /usr/share/avocado/tests/gdbtest.py
+    INSTRUMENTED /usr/share/avocado/tests/gdbtest.py:GdbTest.test_start_exit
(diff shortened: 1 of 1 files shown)


apahim:
[improve the docs regarding the test references](https://trello.com/c/iJlNcBaf/930-improve-the-docs-regarding-the-test-references)

docs/source/GetStartedGuide.rst
```diff
@@ -211,6 +211,75 @@ keep reading.
           positional arguments, therefore test names (in case of
           ``avocado run``)
 
+Test References
+---------------
```
ldoktor: Well this is quite detailed explanation, I'd not bother someone starting with Avocado with it. What would you think about moving this to `ReferenceGuide` section?

ldoktor: Actually it's already described there and the detailed explanation here only applies to `FileLoader`. So how about tweak it a bit and create a `FileLoader discovery` section in `Loaders` with all these neat details?

docs/source/GetStartedGuide.rst
```diff
@@ -211,6 +211,75 @@ keep reading.
...
+
+A Test Reference is a string that can be resolved into (interpreted as)
+one or more tests by the Avocado Test Resolver. It is usually a path for
+a test file and it can be followed by a `:` and then a regular expression
```
ldoktor: Well this is up to the loader plugin, `avocado-vt`, `external` does not use `path` and `SIMPLE` test also does not support regexps.

docs/source/GetStartedGuide.rst
apahim: Yes, makes sense.

docs/source/GetStartedGuide.rst
apahim: agreed

apahim:
Sure.

PR #2067, opened 2017-06-14
docs: add `code review` section

apahim:
Let's document the aspects of a code review in Avocado.

Signed-off-by: Amador Pahim <apahim@redhat.com>

PR #2071, opened 2017-06-15
docs: add `code review` section [V2]

apahim:
v2:
- Improve the sentence about PR's failing in the CI.

v1: #2067

### DIFF
--- a/docs/source/ContributionGuide.rst
+++ b/docs/source/ContributionGuide.rst
@@ -242,6 +242,38 @@ approved by your employer) and you grant a full, complete, irrevocable
 copyright license to all users and developers of the Avocado project,
 present and future, pursuant to the license of the project.
 
+Code Review
+-----------
+
+Every single Pull Request in Avocado has to be reviewed by at least one
+other developer. All members of the core team have permission to merge
+a Pull Request, but there are some conditions that have to be fulfilled
+before merging the code:
+
+- Pull Request has to pass the CI tests.
+- One 'Approved' code review should be given.
+- No explicit disapproval should be present.
+
+Pull Requests failing in CI will not be merged, and reviews won't be
+given to them until all the problems are sorted out. If you want help
+to debug the issues with the CI, please reach the developers and/or the
+community using the available communication channels.
+
+While reviewing the code, one should:
+
+- Verify that the code is sound and clean.
+- Run the highest level of selftests (see `make`).
+- Verify that code works to its purpose.
+- Make sure the commits organization is proper (i.e. code is well
+  organized in atomic commits, there's no extra/unwanted commits, ...).
+- Provide an in-line feedback with explicit questions and/or requests of
+  improvements.
+- Provide a general feedback in the review message, being explicit about
+  what's expected for the next Pull Request version, if that's the case.
+
+When the Pull Request is approved, the reviewer will merge the code or
+wait for someone with merge permission to merge it.
+
 Tests Repositories
 ==================
 


docs/source/ContributionGuide.rst
```diff
@@ -242,6 +242,38 @@ approved by your employer) and you grant a full, complete, irrevocable
...
+-----------
+
+Every single Pull Request in Avocado has to be reviewed by at least one
+other developer. All members of the core team have permission to merge
+a Pull Request, but there are some conditions that have to be fulfilled
+before merging the code:
+
+- Pull Request has to pass the CI tests.
+- One 'Approved' code review should be given.
+- No explicit disapproval should be present.
```
ldoktor: Well is it `No` or is it that the ratio has to be +1? Because we already merged PRs with 1 NACK and 2 ACKS. Anyway this is not that important and can be tweaked later...

docs/source/ContributionGuide.rst
```diff
@@ -242,6 +242,38 @@ approved by your employer) and you grant a full, complete, irrevocable
...
+
+Pull Requests failing in CI will not be merged, and reviews won't be
+given to them until all the problems are sorted out. If you want help
+to debug the issues with the CI, please reach the developers and/or the
+community using the available communication channels.
+
+While reviewing the code, one should:
+
+- Verify that the code is sound and clean.
+- Run the highest level of selftests (see `make`).
```
ldoktor: is this a requirement? I've never done this (except of release)

docs/source/ContributionGuide.rst
apahim: Is it not? I was told that it is a requirement: an explicit NACK will make a given PR not mergeable.

docs/source/ContributionGuide.rst
apahim: Not sure it's a requirement. I see it as a good practice and it's part of my workflow.

docs/source/ContributionGuide.rst
ldoktor: I don't really know, anyway it's always up to the situation, we are but humans...

docs/source/ContributionGuide.rst
ldoktor: to me it seems unnecessary as it won't work with bisect and we do run full check on release...

docs/source/ContributionGuide.rst
clebergnu: I agree with "No explicit disapproval".  A NACK should be given when something is clearly broken and will disrupt the project.  It should never be given for situations where one sees an alternative implementation, if it has some benefits.  For instance, optmizing code for better (less) CPU usage in non-critical areas should *not* be a reson for a NACK.

docs/source/ContributionGuide.rst
clebergnu: The highest level is what we currently call `make check-full`.  Since we're *not* running that under CI, I think we should indeed be running it locally.  TBH, I have been running `make check`, but I support changing that.

docs/source/ContributionGuide.rst
lmr: Explicit NACK is different than requesting changes, and it's up for discussion, one can help a person that NACKed a given Pull Request to change his/her mind later. It seems reasonable to me.

clebergnu: @lmr right, then the NACK would be lifted/removed.

ldoktor: ack

docs/source/ContributionGuide.rst
ldoktor: well, I don't like that, but I can learn to live with it.... (which means add it to mine `merge_pr` script).

ldoktor:
[Add review process documentation](https://trello.com/c/ZcGtTAUp/1031-add-review-process-documentation)

PR #2104, opened 2017-06-26
Fix replay `--execution-order`

apahim:
Some adjusts are needed in order to have `--execution-order` properly working with `replay`.

The commits here were cherry-picked from PR #2099.


clebergnu:
[BUG: Replay does not reflects `--execution-order`](https://trello.com/c/0RnLwbs8/1052-bug-replay-does-not-reflects-execution-order)

PR #2114, opened 2017-07-11
utils/memory.py : functions to memory hotplug and hotunplug

abdhaleegit:
hotpluggable, check if kernel supports memory hotplug. also functions for
swap free and swap total

### DIFF
--- a/avocado/utils/memory.py
+++ b/avocado/utils/memory.py
@@ -27,8 +27,75 @@ from . import process
 from . import genio
 
 
-# Returns total memory in kb
+def _check_memory_state(mem_block):
+    """
+    Check the given memory block is online or offline
 
+    :param mem_block: memory block id.
+    :type string: like 198
+    :return: 'True' if online or 'False' if offline
+    :rtype: bool
+    """
+    if 'online' in open('/sys/devices/system/memory/memory%s/state' % mem_block).read():
+        return True
+    return False
+
+
+def _check_hot_plug():
+    """
+    Check kernel support for memory hotplug
+
+    :return: True if hotplug supported,  else False
+    :rtype: 'bool'
+    """
+    if glob.glob('/sys/devices/system/memory/memory*'):
+        return True
+    return False
+
+
+def _is_hot_pluggable(mem_block):
+    """
+    Check if the given memory block is hotpluggable
+
+    :param mem_block: memory block id.
+    :type string: like 198
+    :retrun: True if hotpluggable, else False
+    :rtype: 'bool'
+    """
+    if os.path.exists('/sys/devices/system/memory/memory%s/removable' % mem_block):
+        return True
+    return False
+
+
+def memory_hot_plug(mem_block):
+    """
+    Online the memory for the given block id.
+
+    :param mem_block: memory block id.
+    :type string: like 198
+    """
+    if _is_hot_pluggable(mem_block):
+        with open('/sys/devices/system/memory/memory%s/state' % mem_block, 'w') as fd:
+            fd.write('online')
+        if not _check_memory_state(mem_block):
+            raise ValueError("unable to online memory block, device busy?")
+
+
+def memory_hot_unplug(mem_block):
+    """
+    Offline the memory for the given block id.
+
+    :param mem_block: memory block id.
+    :type string: like 198
+    """
+    if _is_hot_pluggable(mem_block):
+        with open('/sys/devices/system/memory/memory%s/state' % mem_block, 'w') as fd:
+            fd.write('offline')
+        if _check_memory_state(mem_block):
+            raise ValueError("unable to offline memory block, device busy?")
+
+
+# Returns total memory in kb
 
 def read_from_meminfo(key):
     """
@@ -39,6 +106,20 @@ def read_from_meminfo(key):
     cmd_result = process.run('grep %s /proc/meminfo' % key, verbose=False)
     meminfo = cmd_result.stdout
     return int(re.search(r'\d+', meminfo).group(0))
+
+
+def swaptotal():
+    """
+    Read ``SwapTotal`` from meminfo.
+    """
+    return read_from_meminfo('SwapTotal')
+
+
+def freeswap():
+    """
+    Read ``SwapFree`` from meminfo.
+    """
+    return read_from_meminfo('SwapFree')
 
 
 def memtotal():


avocado/utils/memory.py
```diff
@@ -27,8 +27,75 @@
...
+def _is_hot_pluggable(mem_block):
+    """
+    Check if the given memory block is hotpluggable
+
+    :param mem_block: memory block id.
+    :type string: like 198
+    :retrun: True if hotpluggable, else False
+    :rtype: 'bool'
+    """
+    if os.path.exists('/sys/devices/system/memory/memory%s/removable' % mem_block):
```
apahim: It's not the presence/absence of this file that represents the availability of the hot-plug feature. You have to check its content. 1 for hot-plug and 0 for non-hot-plug. If you know for a fact that in some systems this file can be missing for non-hot-plug blocks, than you have to test both the presence and the content of the file.

avocado/utils/memory.py
```diff
@@ -27,8 +27,75 @@
...
+        return True
+    return False
+
+
+def memory_hot_plug(mem_block):
```
apahim: We are in `memory` module, so I'd call this function `def hotplug(block):`

avocado/utils/memory.py
```diff
@@ -27,8 +27,75 @@
...
+    """
+    Online the memory for the given block id.
 
+    :param mem_block: memory block id.
+    :type string: like 198
+    """
+    if _is_hot_pluggable(mem_block):
```
apahim: We should do something when the block is not hot-pluggable. Raising an exception seems reasonable for me.

avocado/utils/memory.py
```diff
@@ -27,8 +27,75 @@
...
+        with open('/sys/devices/system/memory/memory%s/state' % mem_block, 'w') as fd:
+            fd.write('online')
+        if not _check_memory_state(mem_block):
+            raise ValueError("unable to online memory block, device busy?")
```
apahim: According to the documentation: "exception ValueError: Raised when a built-in operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as IndexError."
I don't see a better builtin exception though. You can create your own exceptions here, like `MemoryStateError`, `MemoryHotPlugError`, ..., or even only one `MemoryError` that you can call providing the error message, like: 

```python
class MemoryError(Exception):
    """
    Called when some memory related operation fails
    """
    pass

...
...
def hotplug(block):
    if not _is_hot_pluggable(block):
        raise MemoryError("Memory block %s does not support hot-plug." % block)

    with open('/sys/devices/system/memory/memory%s/state' % mem_block, 'w') as state_file:
        state_file.write('online')

    if not _check_memory_state(block):
        raise MemoryError("Unable to hot-plug memory block %s. Device busy?" % block)
```

avocado/utils/memory.py
```diff
@@ -27,8 +27,75 @@
...
+
+
+def memory_hot_unplug(mem_block):
+    """
+    Offline the memory for the given block id.
+
+    :param mem_block: memory block id.
+    :type string: like 198
+    """
+    if _is_hot_pluggable(mem_block):
```
apahim: The same comment applies here.

avocado/utils/memory.py
```diff
@@ -27,8 +27,75 @@
...
+        with open('/sys/devices/system/memory/memory%s/state' % mem_block, 'w') as fd:
+            fd.write('online')
+        if not _check_memory_state(mem_block):
+            raise ValueError("unable to online memory block, device busy?")
+
+
+def memory_hot_unplug(mem_block):
```
apahim: Suggestion: `def hotunplug(block):`

avocado/utils/memory.py
```diff
@@ -41,6 +108,20 @@ def read_from_meminfo(key):
     return int(re.search(r'\d+', meminfo).group(0))
 
 
+def swaptotal():
+    """
+    Read ``SwapTotal`` from meminfo.
+    """
+    return read_from_meminfo('SwapTotal')
```
apahim: I have very few appreciation for functions that don't really add value to the library. I mean, what's the value of calling `swaptotal()` instead of calling `read_from_meminfo('SwapTotal')`?

avocado/utils/memory.py
```diff
@@ -41,6 +108,20 @@ def read_from_meminfo(key):
...
+
+
+def freeswap():
+    """
+    Read ``SwapFree`` from meminfo.
+    """
+    return read_from_meminfo('SwapFree')
```
apahim: Ditto. Also, this should not be in this commit. But I'd not propose these 2 functions anyway.

avocado/utils/memory.py
```diff
@@ -27,8 +27,75 @@
...
+    :type string: like 198
+    """
+    if _is_hot_pluggable(mem_block):
+        with open('/sys/devices/system/memory/memory%s/state' % mem_block, 'w') as fd:
+            fd.write('offline')
+        if _check_memory_state(mem_block):
+            raise ValueError("unable to offline memory block, device busy?")
+
+
+# Returns total memory in kb
```
apahim: This commend should be removed at once.

avocado/utils/memory.py
```diff
@@ -41,6 +108,20 @@ def read_from_meminfo(key):
     return int(re.search(r'\d+', meminfo).group(0))
```
apahim: This is not related to your commit, but your commit made me read this function, so here we go.
IMO, this function is should be improved. It's using `grep` and a regex search, and both are not really needed.
I'd suggest to parse the `/proc/meminfo` into a dictionary and, accepting `None` key, return the whole dictionary. If some key is provided, than only that key value is returned. It will increase the flexibility and the usability.
Something like:

```python
def read_from_meminfo(key=None):
    meminfo = {}
    with open('/proc/meminfo', 'r') as f:
        for line in f.readlines():
            item, value = line.strip().split(None, 1)
            meminfo[item.strip(':')] = value.strip(' kB')
    if key is not None:
        return meminfo[key]
    return meminfo
```

Usage would be:
```python

>>> read_from_meminfo()
{'WritebackTmp': '0', 'SwapTotal': '7930876', 'Active(anon)': '937468', 'SwapFree': '7930876', 'DirectMap4k': '374784', 'KernelStack': '7312', 'MemFree': '9732968', 'HugePages_Rsvd': '0', 'Committed_AS': '5107800', 'Active(file)': '1240388', 'NFS_Unstable': '0', 'VmallocChunk': '0', 'CmaFree': '0', 'Writeback': '0', 'Inactive(file)': '1698816', 'MemTotal': '15853856', 'VmallocUsed': '0', 'DirectMap1G': '6291456', 'HugePages_Free': '0', 'AnonHugePages': '0', 'AnonPages': '1748280', 'Active': '2177856', 'Inactive(anon)': '940808', 'CommitLimit': '15857804', 'Hugepagesize': '2048', 'Cached': '3067080', 'SwapCached': '0', 'VmallocTotal': '34359738367', 'CmaTotal': '0', 'Shmem': '130140', 'Mapped': '375208', 'ShmemPmdMapped': '0', 'SUnreclaim': '198616', 'Unevictable': '32', 'SReclaimable': '535092', 'MemAvailable': '12868908', 'Mlocked': '32', 'DirectMap2M': '10575872', 'HugePages_Surp': '0', 'Bounce': '0', 'Inactive': '2639624', 'PageTables': '26068', 'HardwareCorrupted': '0', 'ShmemHugePages': '0', 'HugePages_Total': '0', 'Slab': '733708', 'Buffers': '2264', 'Dirty': '1736'}

>>> read_from_meminfo('Active')
'2179888'

>>> read_from_meminfo()['Slab']
'733708'

>>> read_from_meminfo('foo')
  File "<stdin>", line 9, in read_from_meminfo
KeyError: 'foo'
```

apahim: I will send a PR for this part... No need to tackle this unrelated comment.

avocado/utils/memory.py
abdhaleegit: make sense :+1:

avocado/utils/memory.py
abdhaleegit: ah my bad, yes you are right. '1' represents removable and '0' non removable.
there is no case where file will not be present.

avocado/utils/memory.py
abdhaleegit: sounds good

avocado/utils/memory.py
abdhaleegit: yes, I will have that implemented

avocado/utils/memory.py
abdhaleegit: fine

avocado/utils/memory.py
abdhaleegit: I will drop those 2 functions.

avocado/utils/memory.py
abdhaleegit: will remove it

abdhaleegit:
#2161 is the latest with all comments implemented. so closing this one

PR #2130, opened 2017-08-08
runner: fix SIGINT behaviour

apahim:
According to GNU documentation, "...typing C-c sends the signal SIGINT
to terminate all the processes in the foreground process group". But
when the SIGINT in directly sent to the Avocado main process, the test
process receives nothing, making the main process to wait indefinitely
for the test process to finish.

This patch makes the runner to send a SIGINT to the test process when
the main process receives a SIGINT, replicating the behaviour we have
on C-c.

---
Note: for some reason that I fail to see, sending a `SIGINT` to the Avocado main process created by `aexpect.Expect()` makes all the children to also terminate regardless this patch. So, I've written the selftest using `subprocess.Popen`.

### DIFF
--- a/avocado/core/runner.py
+++ b/avocado/core/runner.py
@@ -447,6 +447,7 @@ class TestRunner(object):
                                            ignore_window)
                         stage_1_msg_displayed = True
                     ignore_time_started = time.time()
+                    os.kill(proc.pid, signal.SIGINT)
                 if (ctrl_c_count > 1) and (time_elapsed > ignore_window):
                     if not stage_2_msg_displayed:
                         abort_reason = "Interrupted by ctrl+c (multiple-times)"
--- a/selftests/functional/test_interrupt.py
+++ b/selftests/functional/test_interrupt.py
@@ -1,13 +1,16 @@
 import os
 import tempfile
 import time
+import signal
 import shutil
 import stat
+import subprocess
 import unittest
 
 import aexpect
 import psutil
 
+from avocado.utils import process
 from avocado.utils import wait
 from avocado.utils import script
 from avocado.utils import data_factory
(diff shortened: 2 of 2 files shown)


apahim:

selftests/functional/test_interrupt.py
```diff
@@ -184,6 +187,47 @@ def wait_until_no_goodtest():
...
+        os.kill(proc.pid, signal.SIGINT)
+
+        def is_finished():
+            return proc.poll() is not None
+        finished = wait.wait_for(is_finished, timeout=5)
+        if not finished:
+            process.kill_process_tree(proc.pid)
+
+        self.assertTrue(finished, 'Avocado is still running after '
+                        'receiving SIGINT.')
```
ldoktor: this should be `self.fail` in the `if not finished` branch, or `try/finally` block around the `wait.wait_for` inside `assertTrue`.

ldoktor:
Also I looked at this and IMO it'd be nice to forward the SIGTERM to all children as this is what we do in case of force-killing the hanged test. Without that forked processes are left behind after SIGINT...

avocado/core/runner.py
```diff
@@ -447,6 +447,7 @@ def sigtstp_handler(signum, frame):     # pylint: disable=W0613
                                            ignore_window)
                         stage_1_msg_displayed = True
                     ignore_time_started = time.time()
+                    os.kill(proc.pid, signal.SIGINT)
```
ldoktor:
```python
diff --git a/avocado/core/runner.py b/avocado/core/runner.py
index 1864a3b..face99c 100644
--- a/avocado/core/runner.py
+++ b/avocado/core/runner.py
@@ -447,7 +447,7 @@ class TestRunner(object):
                                            ignore_window)
                         stage_1_msg_displayed = True
                     ignore_time_started = time.time()
-                    os.kill(proc.pid, signal.SIGINT)
+                    process.kill_process_tree(proc.pid, signal.SIGTERM)
                 if (ctrl_c_count > 1) and (time_elapsed > ignore_window):
                     if not stage_2_msg_displayed:
                         abort_reason = "Interrupted by ctrl+c (multiple-times)"
```

apahim:
> This does not work on the reproducer from the card (at least I think so as there is no content in the reproducer, anyway the reproducer suggests a shell script). When I tried this with avocado run --external-runner /bin/sleep 100 it still hanged.

Oh, that's a different issue. Avocado will hang even with Ctrl+C in that case, but it happens because the SIGINT handler is still faulty in `process.SubProcess`.

> As we supply custom hander for SIGTERM I'd suggest using that (which works even with SIMPLE tests).

We handle the SIGINT diferently... It's handled by the `except KeyboardInterrupt`, so I don't see a need for a specific handler in the runner. Using external sunner (or SIMPLE tests), the test subprocess will not handle the SIGINT properly, but that's a bug in the `process.SubProcess` SIGINT handler. I'm fixing that in the V2.

avocado/core/runner.py
apahim: That's not needed. We received a SIGINT and the subprocess should just receive the same signal and handle it properly, just like it receives when we hit Ctrl+C. Problem is with the handler in the subprocess... sending a SIGTERM here will just hide the issue.

ldoktor: Well we can define any behavior we want and interrupting the test process by SIGTERM looks like sane behavior on ctrl+c sent to the runner. Anyway looking forward to the v2, hopefully it'll cover the hang as well as killing forked processes like this fix.

PR #2150, opened 2017-08-14
Fix SIGINT/SIGTERM [V4]

apahim:
v4:
- Selftests:

v3: #2140 

v2: #2137 
- Improve SIGINT selftest.

v1: #2130

### DIFF
--- a/selftests/functional/test_interrupt.py
+++ b/selftests/functional/test_interrupt.py
@@ -1,13 +1,15 @@
 import os
 import tempfile
 import time
+import signal
 import shutil
 import stat
+import subprocess
 import unittest
 
-import aexpect
 import psutil
 
+from avocado.utils import process
 from avocado.utils import wait
 from avocado.utils import script
 from avocado.utils import data_factory
@@ -25,13 +27,20 @@ DEFAULT_MODE = (stat.S_IRUSR | stat.S_IWUSR | stat.S_IXUSR |
                 stat.S_IROTH | stat.S_IXOTH)
 
 BAD_TEST = """#!/usr/bin/env python
+import multiprocessing
 import signal
 import time
+
+def foo():
+    while True:
+        time.sleep(0.1)
 
 if __name__ == "__main__":
     signal.signal(signal.SIGINT, signal.SIG_IGN)
     signal.signal(signal.SIGTERM, signal.SIG_IGN)
     signal.signal(signal.SIGQUIT, signal.SIG_IGN)
+    proc = multiprocessing.Process(target=foo)
+    proc.start()
     while True:
         time.sleep(0.1)
 """
@@ -54,11 +63,12 @@ class InterruptTest(unittest.TestCase):
 
     def setUp(self):
         self.tmpdir = tempfile.mkdtemp(prefix='avocado_' + __name__)
+        self.test_module = None
 
     @unittest.skipIf(int(os.environ.get("AVOCADO_CHECK_LEVEL", 0)) < 2,
                      "Skipping test that take a long time to run, are "
                      "resource intensive or time sensitve")
-    def test_badly_behaved(self):
+    def test_badly_behaved_sigint(self):
         """
         Make sure avocado can cleanly get out of a loop of badly behaved tests.
         """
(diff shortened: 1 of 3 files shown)


clebergnu:

selftests/functional/test_interrupt.py
```diff
@@ -67,64 +77,87 @@ def test_badly_behaved(self):
...
         bad_test.save()
-
+        self.test_module = bad_test.path
         os.chdir(basedir)
-        cmd_line = ('%s run --sysinfo=off --job-results-dir %s '
-                    '%s %s %s' % (AVOCADO, self.tmpdir, bad_test.path,
-                                  bad_test.path, bad_test.path))
-        proc = aexpect.Expect(command=cmd_line, linesep='')
-        proc.read_until_last_line_matches(os.path.basename(bad_test.path))
-        proc.sendline('\x03')
```
clebergnu: Right, it's fine to replace a key that will be translated as SIGINT by the signal itself.  The current difference in treatment is actually one of the points of the original card hoped to resolve.

selftests/functional/test_interrupt.py
```diff
@@ -67,64 +77,87 @@ def test_badly_behaved(self):
...
-                except psutil.NoSuchProcess:
-                    pass
-
-            return len(bad_test_processes) == 0
-
-        wait.wait_for(wait_until_no_badtest, timeout=2)
+        if not wait.wait_for(self._no_test_in_process_table, timeout=2):
+            self.fail('Avocado left processes behind.')
+
+        output = self.proc.communicate()[0]
```
clebergnu: This works, but I think it passes the wrong message to readers (it happened to me, at least).  Popen.communicate has the side effect of returning stdout (and stderr), but it's intended to send data to the process' stdin.  How about:

```diff
diff --git a/selftests/functional/test_interrupt.py b/selftests/functional/test_interrupt.py
index 4325b1a3..e9d9ce31 100644
--- a/selftests/functional/test_interrupt.py
+++ b/selftests/functional/test_interrupt.py
@@ -106,7 +106,7 @@ class InterruptTest(unittest.TestCase):
         if not wait.wait_for(self._no_test_in_process_table, timeout=2):
             self.fail('Avocado left processes behind.')
 
-        output = self.proc.communicate()[0]
+        output = self.proc.stdout.read()
         # Make sure the Interrupted requested sentence is there
         self.assertIn('Interrupt requested. Waiting 2 seconds for test to '
                       'finish (ignoring new Ctrl+C until then)', output)
@@ -150,9 +150,8 @@ class InterruptTest(unittest.TestCase):
         if not wait.wait_for(self._no_test_in_process_table, timeout=2):
             self.fail('Avocado left processes behind.')
 
-        output = self.proc.communicate()[0]
         # Make sure the Interrupted test sentence is there
-        self.assertIn('Terminated\n', output)
+        self.assertIn('Terminated\n', self.proc.stdout.read())
 
     @unittest.skipIf(int(os.environ.get("AVOCADO_CHECK_LEVEL", 0)) < 1,
                      "Skipping test that take a long time to run, are "
@@ -191,7 +190,7 @@ class InterruptTest(unittest.TestCase):
         if not wait.wait_for(self._no_test_in_process_table, timeout=2):
             self.fail('Avocado left processes behind.')
 
-        output = self.proc.communicate()[0]
+        output = self.proc.stdout.read()
         # Make sure the Interrupted requested sentence is there
         self.assertIn('Interrupt requested. Waiting 2 seconds for test to '
                       'finish (ignoring new Ctrl+C until then)', output)
@@ -235,9 +234,8 @@ class InterruptTest(unittest.TestCase):
         if not wait.wait_for(self._no_test_in_process_table, timeout=2):
             self.fail('Avocado left processes behind.')
 
-        output = self.proc.communicate()[0]
         # Make sure the Interrupted test sentence is there
-        self.assertIn('Terminated\n', output)
+        self.assertIn('Terminated\n', self.proc.stdout.read())
 
     def tearDown(self):
         shutil.rmtree(self.tmpdir)

```

selftests/functional/test_interrupt.py
```diff
@@ -134,59 +167,121 @@ def test_well_behaved(self):
...
+        self.proc = subprocess.Popen(cmd.split(),
+                                     stdout=subprocess.PIPE,
+                                     stderr=subprocess.STDOUT)
+
+        if not wait.wait_for(self._has_children, timeout=2):
+            process.kill_process_tree(self.proc.pid)
+            self.fail('Avocado did not start the test process.')
+
+        # This test will not ignore SIGINT, so it will
+        # terminate right away.
```
clebergnu: This code is fine, I just had to say that affirming the process *will* terminate right away "signals" a lot of confidence! (pun intended).  It *should*, though, and this is what we're testing.

PR #2151, opened 2017-08-14
Porting avocado.util.process to Python 3 (do not merge)

bonzini:
This pull request adds two more properties stdout_bytes and stderr_bytes to CmdResult; the decoded stdout and stderr are computed lazily. However, this still causes a few Python 2 failures, so this pull request is only provided for convenience in analyzing those failures.

### DIFF
--- a/avocado/utils/process.py
+++ b/avocado/utils/process.py
@@ -267,11 +267,14 @@ class CmdResult(object):
     :param duration: Elapsed wall clock time running the process
     :param encoding: Encoding used by the stdout and stderr properties
                  to decode stdout_bytes and stderr_bytes.
+    :param errors: The error handler used by the stdout and stderr
+                 properties when Unicode decoding fails.
     :param pid: ID of the process
     """
 
     def __init__(self, command="", stdout_bytes=b"", stderr_bytes=b"",
-                 exit_status=None, duration=0, encoding='utf-8', pid=None):
+                 exit_status=None, duration=0, encoding='utf-8',
+                 errors='strict', pid=None):
         self.command = command
         self.exit_status = exit_status
         self.stdout_bytes = stdout_bytes
@@ -280,6 +283,7 @@ class CmdResult(object):
         self._stderr = None
         self.duration = duration
         self.encoding = encoding
+        self.errors = errors
         self.interrupted = False
         self.pid = pid
 
@@ -287,14 +291,14 @@ class CmdResult(object):
     def stdout(self):
         if self._stdout is not None:
             return self._stdout
-        self._stdout = self.stdout_bytes.decode(self.encoding)
+        self._stdout = self.stdout_bytes.decode(self.encoding, self.errors)
         return self._stdout
 
     @property
     def stderr(self):
         if self._stderr is not None:
             return self._stderr
-        self._stderr = self.stderr_bytes.decode(self.encoding)
+        self._stderr = self.stderr_bytes.decode(self.encoding, self.errors)
         return self._stderr
 
     def __repr__(self):
@@ -319,7 +323,8 @@ class SubProcess(object):
 
     def __init__(self, cmd, verbose=True, allow_output_check='all',
                  shell=False, env=None, sudo=False,
-                 ignore_bg_processes=False, encoding='utf-8'):
+                 ignore_bg_processes=False, encoding='utf-8',
+                 errors='strict'):
         """
         Creates the subprocess object, stdout/err, reader threads and locks.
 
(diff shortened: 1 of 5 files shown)


avocado/utils/process.py
```diff
@@ -267,16 +267,32 @@ class CmdResult(object):
...
+            return self._stdout
+        self._stdout = self.stdout_bytes.decode('utf-8')
+        return self._stdout
+
+    @property
+    def stderr(self):
+        if self._stderr is not None:
+            return self._stderr
+        self._stderr = self.stderr_bytes.decode('utf-8')
+        return self._stderr
```
lmr: Yes, nice touch turning .stdout and .stderr to properties. I'm pondering whether .stdout_bytes and .stderr_bytes should be public attributes.

avocado/utils/process.py
```diff
@@ -268,29 +268,30 @@ class CmdResult(object):
...
-                 exit_status=None, duration=0, pid=None):
+                 exit_status=None, duration=0, encoding='utf-8', pid=None):
         self.command = command
         self.exit_status = exit_status
         self.stdout_bytes = stdout_bytes
         self.stderr_bytes = stderr_bytes
         self._stdout = None
         self._stderr = None
         self.duration = duration
+        self.encoding = encoding
```
lmr: Apart from the fact we'll need to update the docstrings in the final version, this looks good.

avocado/utils/process.py
bonzini: I'm using them in iso9660 so they definitely should. In fact, unless you know what you're doing they should be preferred to stdout and stderr for both efficiency and robustness reasons.

avocado/utils/process.py
```diff
@@ -267,16 +267,32 @@ class CmdResult(object):
...
+        self._stderr = None
         self.duration = duration
         self.interrupted = False
         self.pid = pid
 
+    @property
+    def stdout(self):
+        if self._stdout is not None:
+            return self._stdout
+        self._stdout = self.stdout_bytes.decode('utf-8')
```
ldoktor: Hello @bonzini, thank you for all the python3 related patches. This one troubles me quite a lot as we've had bugs reported by users with non-utf-8 encodings, so `decode('utf-8')` is not an answer everywhere. The thing is I don't really know the solution, we used combination of PYTHONENCODING, PYTHONIOENCODING and `sys.getdefaultencoding()` and changing `ascii` with `utf8` (as python2 reports most frequently ascii).

Ideal solution would be to include some tests for this module for encoding together with this patch, some could be inspired by https://github.com/python/cpython/blob/master/Lib/test/test_subprocess.py

PS: The `avocado.utils.process` is heavily utilized module, that's why I'm so concerned about basically any change here.

avocado/utils/process.py
```diff
@@ -510,8 +528,8 @@ def _fill_streams(self):
         # Clean subprocess pipes and populate stdout/err
         self._popen.stdout.close()
         self._popen.stderr.close()
-        self.result.stdout = self.get_stdout()
-        self.result.stderr = self.get_stderr()
+        self.result.stdout_bytes = self.get_stdout()
+        self.result.stderr_bytes = self.get_stderr()
```
ldoktor: Do we need the raw stdout? How about moving the encoding part to `get_stdout` and report decoded text in the first place?

bonzini: Yes, we do. For example iso9660 reads arbitrary binary data from isoinfo/isoread. Likewise, you can imagine tests running e.g. `pnmtopng` using `avocado.utils.process`.

avocado/utils/process.py
bonzini: Yes, this is also why I am doing the decoding lazily. Using utf-8 is definitely not the definitive thing, hence the introduction of encoding in the next patch. They can be squashed, and the default changed from utf-8 to something else as you said.

avocado/utils/process.py
ldoktor: Fair enough.

avocado/utils/process.py
ldoktor: I see, I'd not squash it as this way it's easier to follow the changes and it should not break things terribly...

ldoktor:
Btw the travis failed due to time-limit. Please rebase the code to the latest master and it should work...

bonzini:
Not really:

    ERROR: test_non_absolute_path (selftests.functional.test_basic.RunnerSimpleTest)
      File "/home/travis/build/avocado-framework/avocado/selftests/functional/test_basic.py", line 796, in test_non_absolute_path
        (expected_rc, result))
      File "/home/travis/build/avocado-framework/avocado/avocado/utils/process.py", line 305, in __repr__
        self.pid))
    UnicodeDecodeError: 'ascii' codec can't decode byte 0xca in position 158: ordinal not in range(128)
    ERROR: test_simpletest_pass (selftests.functional.test_basic.RunnerSimpleTest)
      File "/home/travis/build/avocado-framework/avocado/selftests/functional/test_basic.py", line 709, in test_simpletest_pass
        (expected_rc, result))
      File "/home/travis/build/avocado-framework/avocado/avocado/utils/process.py", line 305, in __repr__
        self.pid))
    UnicodeDecodeError: 'ascii' codec can't decode byte 0xca in position 156: ordinal not in range(128)

ldoktor:
ok, let's say it also failed because of timeout...

lmr:
@bonzini Yep, I'm going to take a look at the errors when I'm done with a PR I'm working on. Meanwhile the rebase against master should help making the travis job to not time out.

clebergnu:

clebergnu:
@bonzini we just merged code that is similar to your proposal here:

https://github.com/avocado-framework/avocado/commit/7d0e6a44491df1f44d9ddc6064fa5bbde644346c

https://github.com/avocado-framework/avocado/commit/01c848666650fc35b003b8e54a851c6e63355023

So I'm closing this PR.  Thanks for helping with the Python 3 port!

PR #2152, opened 2017-08-15
Fix SIGINT/SIGTERM [V5]

apahim:
v5:
- Add docs.
- Selftests:
  - Use `proc.stdout.read()` instead of `proc.communicate()`
  - Improve comments.

v4: #2150 
- Selftests:

v3: #2140 

v2: #2137 
- Improve SIGINT selftest.

v1: #2130

### DIFF
--- a/docs/source/ReferenceGuide.rst
+++ b/docs/source/ReferenceGuide.rst
@@ -653,4 +653,31 @@ be a valid one:
  * An end of string (or end of line) must immediately follow the
    content.
 
+.. _signal_handlers:
+
+Signal Handlers
+===============
+
+Avocado normal operation is related to run code written by
+users/test-writers. It means the test code can carry its own handlers
+for different signals or even ignore then. Still, as the code is being
+executed by Avocado, we have to make sure we will finish all the
+subprocesses we create before ending our execution.
+
+Signals sent to the Avocado main process will be handled as follows:
+
+- SIGSTOP/Ctrl+Z: On SIGSTOP, Avocado will pause the execution of the
+  subprocesses, while the main process will still be running, waiting
+  for the subprocesses to finish. A new SIGSTOP will make the
+  subprocesses to resume the execution.
+- SIGINT/Ctrl+C: This signal will be forwarded to the test process and
+  Avocado will wait until it's finished. If the test process does not
+  finish after receiving a SIGINT, user can send a second SIGINT (after
+  the 2 seconds ignore period). The second SIGINT will make Avocado
+  to send a SIGKILL to the whole subprocess tree and then complete the
+  main process execution.
+- SIGTERM: This signal will make Avocado to terminate immediately. A
+  SIGKILL will be sent to the whole subprocess tree and the main process
+  will exit without completing the execution.
+
 .. [#f1] Avocado plugins can introduce additional test types.
--- a/selftests/functional/test_interrupt.py
+++ b/selftests/functional/test_interrupt.py
@@ -1,13 +1,15 @@
 import os
 import tempfile
 import time
+import signal
 import shutil
 import stat
+import subprocess
 import unittest
 
-import aexpect
 import psutil
 
+from avocado.utils import process
 from avocado.utils import wait
 from avocado.utils import script
 from avocado.utils import data_factory
@@ -25,13 +27,20 @@ DEFAULT_MODE = (stat.S_IRUSR | stat.S_IWUSR | stat.S_IXUSR |
                 stat.S_IROTH | stat.S_IXOTH)
 
 BAD_TEST = """#!/usr/bin/env python
+import multiprocessing
 import signal
 import time
+
+def foo():
+    while True:
+        time.sleep(0.1)
 
 if __name__ == "__main__":
     signal.signal(signal.SIGINT, signal.SIG_IGN)
     signal.signal(signal.SIGTERM, signal.SIG_IGN)
     signal.signal(signal.SIGQUIT, signal.SIG_IGN)
+    proc = multiprocessing.Process(target=foo)
+    proc.start()
     while True:
         time.sleep(0.1)
 """
@@ -54,11 +63,12 @@ class InterruptTest(unittest.TestCase):
 
     def setUp(self):
         self.tmpdir = tempfile.mkdtemp(prefix='avocado_' + __name__)
+        self.test_module = None
 
     @unittest.skipIf(int(os.environ.get("AVOCADO_CHECK_LEVEL", 0)) < 2,
                      "Skipping test that take a long time to run, are "
                      "resource intensive or time sensitve")
-    def test_badly_behaved(self):
+    def test_badly_behaved_sigint(self):
         """
         Make sure avocado can cleanly get out of a loop of badly behaved tests.
         """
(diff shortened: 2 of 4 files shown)


apahim:

docs/source/ReferenceGuide.rst
```diff
@@ -653,4 +653,31 @@ be a valid one:
...
+
+Avocado normal operation is related to run code written by
+users/test-writers. It means the test code can carry its own handlers
+for different signals or even ignore then. Still, as the code is being
+executed by Avocado, we have to make sure we will finish all the
+subprocesses we create before ending our execution.
+
+Signals sent to the Avocado main process will be handled as follows:
+
+- SIGSTOP/Ctrl+Z: On SIGSTOP, Avocado will pause the execution of the
```
ldoktor: This is suppose to be SIGTSTP. SIGSTOP can't be caught.

apahim: Ops...

docs/source/ReferenceGuide.rst
```diff
@@ -653,4 +653,31 @@ be a valid one:
...
+  subprocesses, while the main process will still be running, waiting
+  for the subprocesses to finish. A new SIGSTOP will make the
```
ldoktor: timeout timer still running...

docs/source/ReferenceGuide.rst
```diff
@@ -653,4 +653,31 @@ be a valid one:
...
+  subprocesses to resume the execution.
+- SIGINT/Ctrl+C: This signal will be forwarded to the test process and
+  Avocado will wait until it's finished. If the test process does not
+  finish after receiving a SIGINT, user can send a second SIGINT (after
+  the 2 seconds ignore period). The second SIGINT will make Avocado
+  to send a SIGKILL to the whole subprocess tree and then complete the
+  main process execution.
+- SIGTERM: This signal will make Avocado to terminate immediately. A
+  SIGKILL will be sent to the whole subprocess tree and the main process
+  will exit without completing the execution.
```
ldoktor: I'd add a note that we do best-effort approach, therefor in case of fork-bomb some of the newly created processes might not be killed.

selftests/functional/test_interrupt.py
```diff
@@ -134,59 +166,120 @@ def test_well_behaved(self):
...
+        if not wait.wait_for(self._no_test_in_process_table, timeout=2):
+            self.fail('Avocado left processes behind.')
+
+        # Make sure the Interrupted test sentence is there
+        self.assertIn('Terminated\n', self.proc.stdout.read())
 
     def tearDown(self):
         shutil.rmtree(self.tmpdir)
 
+    def _has_children(self):
```
ldoktor: It's generally better to include function before their usage.

selftests/functional/test_interrupt.py
```diff
@@ -134,59 +166,120 @@ def test_well_behaved(self):
...
+        return len(psutil.Process(self.proc.pid).children()) > 0
```
ldoktor: Also you do spawn 2 test processes, I think it'd be better to wait for at least `> 1` (to avoid proceeding when sysinfo is being generated). It'd also fail in case of daemons, but it should behave way better (alternative would be to create a file inside the test, but IMO that's not necessary)

selftests/functional/test_interrupt.py
```diff
@@ -67,64 +77,86 @@ def test_badly_behaved(self):
...
-                            bad_test_processes.append(p_obj)
-                # psutil.NoSuchProcess happens when the original
-                # process already ended and left the process table
-                except psutil.NoSuchProcess:
-                    pass
-
-            return len(bad_test_processes) == 0
-
-        wait.wait_for(wait_until_no_badtest, timeout=2)
+        if not wait.wait_for(self._no_test_in_process_table, timeout=2):
```
ldoktor: nitpick: Could be `self.assertTrue` but I'm fine with this...

selftests/functional/test_interrupt.py
```diff
@@ -67,64 +77,86 @@ def test_badly_behaved(self):
...
+            self.fail('Avocado left processes behind.')
+
+        output = self.proc.stdout.read()
```
ldoktor: The `communicate` is the recommended way of obtaining the output and finish the communication with the process, can I know why did you change it to `stdout.read`?

selftests/functional/test_interrupt.py
```diff
@@ -67,64 +77,86 @@ def test_badly_behaved(self):
...
+        # Make sure the Interrupted requested sentence is there
+        self.assertIn('Interrupt requested. Waiting 2 seconds for test to '
+                      'finish (ignoring new Ctrl+C until then)', output)
         # Make sure the Killing test subprocess message did appear
-        self.assertIn('Killing test subprocess', proc.get_output())
+        self.assertIn('Killing test subprocess', output)
```
ldoktor: To be honest I imagined even greater code-reuse, but this at least shares the most delicate parts so I'm fine with this way at this point.

clebergnu:
What do you mean by "finish the communication"?

clebergnu:
Now I see, you mean that it waits for the process to finish.

selftests/functional/test_interrupt.py
apahim: It makes no difference for class methods.

selftests/functional/test_interrupt.py
apahim: Sure, I'm ok with both.

selftests/functional/test_interrupt.py
apahim: Yes, I considered what would be the best balance between readability and code re-use. Since this is a selftest (and you're fine), I think we are good here. More code re-use would make it a little harder to follow, imo.

selftests/functional/test_interrupt.py
apahim: I don't see a reason why you couldn't :)
I changed because I was requested to && at this point we already went through `self._is_finished || process.kill_process_tree(self.proc.pid)`, so the `communicate()` added value in waiting for the process to finish, in this case, is not really important.

selftests/functional/test_interrupt.py
apahim: Ok

apahim: Actually, this is checking children. So, it's the number of subprocesses, excluding the main process. It means we are good with `> 0`.

apahim:
@ldoktor @clebergnu, waiting for an agreement between the two of you regarding the `proc.stdout.read()` vs `proc.communicate()`. I ok with both.

clebergnu:
@ldoktor you say that "communicate is the recommended way of obtaining the output and finish the communication with the process", and I agree.  @apahim explained that the checking the process termination is already done.  I proposed using `proc.stdout.read()` to improve code readability (its meaning is much more straightforward IMO).

@ldoktor do you see any problem with the current approach?  If not, @apahim can go ahead and send a v6 without reverting that chunk.

ldoktor:
I'm fine with both with slight preference towards `communicate`. So just the documentation and such and we're ready to ship.

apahim:
Thanks.

PR #2154, opened 2017-08-15
Fix SIGINT/SIGTERM [v6]

apahim:
v6:
- Fixes in docs.
- Helper methods declared ahead of their calls.
- Use `self.assertTrue` instead of `if not wait.wait_for` where applies.

v5: #2152 
- Add docs.
- Selftests:
  - Use `proc.stdout.read()` instead of `proc.communicate()`
  - Improve comments.

v4: #2150 
- Selftests:

v3: #2140 

v2: #2137 
- Improve SIGINT selftest.

v1: #2130


clebergnu:

PR #2161, opened 2017-08-18
utils/memory.py : functions to memory hotplug and hotunplug [v3]

abdhaleegit:
hotpluggable, check if kernel supports memory hotplug. also functions for
swap free and swap total

Implemented review comments from #2114 

### DIFF
--- a/avocado/utils/memory.py
+++ b/avocado/utils/memory.py
@@ -25,9 +25,88 @@ import logging
 
 from . import process
 from . import genio
+from . import wait
 
 
-# Returns total memory in kb
+class MemoryError(Exception):
+
+    """
+    called when memory operations fails
+    """
+    pass
+
+
+def _check_memory_state(block):
+    """
+    Check the given memory block is online or offline
+
+    :param block: memory block id.
+    :type string: like 198
+    :return: 'True' if online or 'False' if offline
+    :rtype: bool
+    """
+    def _is_online():
+        with open('/sys/devices/system/memory/memory%s/state' % block, 'r') as state_file:
+            if 'online' in state_file.read():
+                return True
+            return False
+
+    return wait.wait_for(_is_online, timeout=120, step=1) or False
+
+
+def _check_hotplug():
+    """
+    Check kernel support for memory hotplug
+
+    :return: True if hotplug supported,  else False
+    :rtype: 'bool'
+    """
+    if glob.glob('/sys/devices/system/memory/memory*'):
+        return True
+    return False
+
+
+def _is_hot_pluggable(block):
+    """
+    Check if the given memory block is hotpluggable
+
+    :param block: memory block id.
+    :type string: like 198
+    :retrun: True if hotpluggable, else False
+    :rtype: 'bool'
+    """
+    with open('/sys/devices/system/memory/memory%s/removable' % block, 'r') as file_obj:
+        return bool(int(file_obj.read()))
+
+
+def hotplug(block):
+    """
+    Online the memory for the given block id.
+
+    :param block: memory block id.
+    :type string: like 198
+    """
+    if _is_hot_pluggable(block):
+        with open('/sys/devices/system/memory/memory%s/state' % block, 'w') as state_file:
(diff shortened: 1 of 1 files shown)


avocado/utils/memory.py
```diff
@@ -27,7 +27,95 @@
...
+    :rtype: bool
+    """
+    while True:
+        try:
+            state = open(
+                '/sys/devices/system/memory/memory%s/state' % block).read()
+            if state not in ('going-online', 'going-offline'):
+                break
+        except Exception:
+            pass
```
apahim: Why to pass? If you are trying to avoiding a known case, be explicit about the exception you're handling and let the other exceptions to break the execution so users will know what's happening.

avocado/utils/memory.py
```diff
@@ -27,7 +27,95 @@
...
+def _check_memory_state(block):
+    """
+    Check the given memory block is online or offline
+
+    :param block: memory block id.
+    :type string: like 198
+    :return: 'True' if online or 'False' if offline
+    :rtype: bool
+    """
+    while True:
```
apahim: This is too aggressive. Not sure if the `state` can stay for a long time in 'going-online' or 'going-offline', but the fact is that you should wait nicely, with a timeout and steps that will avoid killing the cpu in charge of the wait process.
We have a `wait` lib in our utils, it might be useful here. You could have something like:

```python
from avocado.utils import wait
...

def _check_memory_state(block):
    def _is_online():
        # Use `with open()` here. If block is online, return True. Else, return False.
        # No need to check 'going-online'/'going-offline' here.
        # No need to handle exception, IMO. Let users receive the exceptions
        # and handle them, if they want to.

    return wait.wait_for(_is_online, timeout=10, step=1) or False
```

avocado/utils/memory.py
```diff
@@ -27,7 +27,95 @@
...
+def _is_hot_pluggable(block):
+    """
+    Check if the given memory block is hotpluggable
+
+    :param block: memory block id.
+    :type string: like 198
+    :retrun: True if hotpluggable, else False
+    :rtype: 'bool'
+    """
+    if '1' in open('/sys/devices/system/memory/memory%s/removable' % block).read():
```
apahim: - Use `with open()`.
- You can test the content directly:
```python
>>> bool(0)
False
>>> bool(1)
True
```
So...
```python
with open(my_file, 'r') as my_file_obj:
    return bool(int(my_file_obj.read()))
```

avocado/utils/memory.py
abdhaleegit: yes, I will have that raised explicitly

avocado/utils/memory.py
```diff
@@ -25,9 +25,88 @@
...
+    Check the given memory block is online or offline
+
+    :param block: memory block id.
+    :type string: like 198
+    :return: 'True' if online or 'False' if offline
+    :rtype: bool
+    """
+    def _is_online():
+        with open('/sys/devices/system/memory/memory%s/state' % block, 'r') as state_file:
+            if 'online' in state_file.read():
```
apahim: I think `going-online` will return `True` here, right? Maybe `if state_file.read() == 'online\n'`?

avocado/utils/memory.py
```diff
@@ -25,9 +25,88 @@
...
+
+
+def hotplug(block):
+    """
+    Online the memory for the given block id.
+
+    :param block: memory block id.
+    :type string: like 198
+    """
+    if _is_hot_pluggable(block):
```
apahim: I miss the `else` here. No action if the memory is not hot-pluggable?

apahim:
@abdhaleegit please rebase it so it will pass on Travis.

avocado/utils/memory.py
abdhaleegit: May be we can leave it to user to handle the _is_hot_pluggable(block) and else part.

so keeping just hotplug and hotunplug operation here

abdhaleegit:
rebased and implemented a new PR raised #2165

PR #2165, opened 2017-08-22
utils/memory: Memory hotplug and hotunplug functions [v3]

abdhaleegit:
hotpluggable, check if kernel supports memory hotplug.

Implemented review comments from #2114, #2116

### DIFF
--- a/avocado/utils/memory.py
+++ b/avocado/utils/memory.py
@@ -25,9 +25,86 @@ import logging
 
 from . import process
 from . import genio
+from . import wait
 
 
-# Returns total memory in kb
+class MemoryError(Exception):
+
+    """
+    called when memory operations fails
+    """
+    pass
+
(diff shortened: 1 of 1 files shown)


avocado/utils/memory.py
```diff
@@ -25,9 +25,86 @@
...
+
+    :return: True if hotplug supported,  else False
+    :rtype: 'bool'
+    """
+    if glob.glob('/sys/devices/system/memory/memory*'):
+        return True
+    return False
+
+
+def _is_hot_pluggable(block):
```
apahim: This helper function is not being used anywhere now. If it's intended to be used by users, it should not have the leading underscore in the name.

apahim:
@abdhaleegit, just a final minor adjustment, if you don't mind.

PR #2180, opened 2017-08-29
test: flag that test is being skipped

apahim:
Now that the teardown() is always called, we have to explicitly flag
when it should be skipped. That's the case when someone decorates the
setUp() with a skip decorator or when someone calls self.skip() in
setUp().

Reference: https://trello.com/c/B6DJe0JS

### DIFF
--- a/selftests/functional/test_statuses.py
+++ b/selftests/functional/test_statuses.py
@@ -25,8 +25,6 @@ ALL_MESSAGES = ['setup pre',
 #                     [msg_in, ...])
 EXPECTED_RESULTS = {'skip-setup-d304': ('SKIP',
                                         ['setup pre',
-                                         'teardown pre',
-                                         'teardown post',
                                          "[WARNING: self.skip() will be "
                                          "deprecated. Use 'self.cancel()' "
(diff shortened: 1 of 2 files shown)


apahim:
[[BUG] Skip decorators in setUp() will make tearDown() to be executed](https://trello.com/c/B6DJe0JS/1109-bug-skip-decorators-in-setup-will-make-teardown-to-be-executed)

selftests/functional/test_statuses.py
```diff
@@ -25,8 +25,6 @@
 #                     [msg_in, ...])
 EXPECTED_RESULTS = {'skip-setup-d304': ('SKIP',
                                         ['setup pre',
-                                         'teardown pre',
-                                         'teardown post',
```
ldoktor: Yeah, I should have noticed this as not-expected behavior, but it sounded right when I was merging it...

ldoktor: I mean it shouldn't really matter as `tearDown` should be written the way it does not require `setUp` to be executed, but this version better sticks to what we define as SKIP.

PR #2187, opened 2017-09-07
Added two more functions to multipath.py library file

Naresh-ibm:
Added fail_path() and reinstate_path() functions which fail given
multipath disk and reinstate the multipath disk respectively.

### DIFF
--- a/avocado/utils/multipath.py
+++ b/avocado/utils/multipath.py
@@ -161,6 +161,36 @@ def get_path_status(disk_path):
                     return(paths['dm_st'], paths['dev_st'], paths['chk_st'])
 
 
+def fail_path(path):
+    """
+    failing the individual paths
+    :param disk_path: disk path. Example: sda, sdb.
+    :return: True or False
+    """
+    cmd = 'multipathd -k"fail path %s"' % path
+    if process.system(cmd) == 0:
+        time.sleep(2)
+        path_stat = get_path_status(path)
+        if path_stat[0] == 'failed' or path_stat[2] == 'faulty':
+            return True
+    return False
+
+
+def reinstate_path(path):
+    """
+    reinstating the individual paths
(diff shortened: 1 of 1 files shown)


avocado/utils/multipath.py
```diff
@@ -161,6 +161,36 @@ def get_path_status(disk_path):
...
+    """
+    failing the individual paths
+    :param disk_path: disk path. Example: sda, sdb.
+    :return: True or False
+    """
+    cmd = 'multipathd -k"fail path %s"' % path
+    if process.system(cmd) == 0:
+        time.sleep(2)
+        path_stat = get_path_status(path)
+        if path_stat[0] == 'failed' or path_stat[2] == 'faulty':
```
narasimhan-v: `and` or `or` here ? ie, should you check both are only one here ? Since you are checking both while reinstating the path

avocado/utils/multipath.py
```diff
@@ -161,6 +161,36 @@ def get_path_status(disk_path):
...
 
+def fail_path(path):
+    """
+    failing the individual paths
+    :param disk_path: disk path. Example: sda, sdb.
+    :return: True or False
+    """
+    cmd = 'multipathd -k"fail path %s"' % path
+    if process.system(cmd) == 0:
+        time.sleep(2)
```
narasimhan-v: Please avoid giving random sleep values. A uniform value of sleep everywhere would be nice.

apahim: Actually the best approach here would be a helper function that returns True/False and a call to `avocado.utils.wait.wait_for()` with a timeout, passing that function. See comment here with an example:
https://github.com/avocado-framework-tests/avocado-misc-tests/pull/605#discussion_r133906038

apahim:
Closed in favor of #2203

PR #2203, opened 2017-09-15
Added two more functions to multipath.py library file

Naresh-ibm:
Added fail_path() and reinstate_path() functions which fail given
multipath disk and reinstate the multipath disk respectively.

### DIFF
--- a/avocado/utils/multipath.py
+++ b/avocado/utils/multipath.py
@@ -161,6 +161,36 @@ def get_path_status(disk_path):
                     return(paths['dm_st'], paths['dev_st'], paths['chk_st'])
 
 
+def fail_path(path):
+    """
+    failing the individual paths
+    :param disk_path: disk path. Example: sda, sdb.
+    :return: True or False
+    """
+    cmd = 'multipathd -k"fail path %s"' % path
+    if process.system(cmd) == 0:
+        time.sleep(10)
+        path_stat = get_path_status(path)
+        if path_stat[0] == 'failed' and path_stat[2] == 'faulty':
+            return True
+    return False
+
+
+def reinstate_path(path):
+    """
+    reinstating the individual paths
+    :param disk_path: disk path. Example: sda, sdb.
+    :return: True or False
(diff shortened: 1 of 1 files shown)


avocado/utils/multipath.py
```diff
@@ -161,6 +161,36 @@ def get_path_status(disk_path):
...
 
+def fail_path(path):
+    """
+    failing the individual paths
+    :param disk_path: disk path. Example: sda, sdb.
+    :return: True or False
+    """
+    cmd = 'multipathd -k"fail path %s"' % path
+    if process.system(cmd) == 0:
+        time.sleep(10)
```
apahim: Please use a helper function that returns True/False and a call to `avocado.utils.wait.wait_for()` with a timeout. See comment here with an example:
avocado-framework-tests/avocado-misc-tests#605 (comment)

Naresh-ibm: hi @apahim wait_for() accepting only function and not the strings.
so here in this case for a single line I have to write one more function which is not a big thing but I feel may not be right way.
So I thought can we add some line of code to wait_for() where it can accept the a string (which may be a command) and return true or false?

apahim: No nee to touch `wait_for()`. This might do the trick (it needs to be tested):

```python
def fail_path(path):
    """
    failing the individual paths
    :param disk_path: disk path. Example: sda, sdb.
    :return: True or False
    """
    def is_failed():
        path_stat = get_path_status(path)
        if path_stat[0] == 'failed' and path_stat[2] == 'faulty':
            return True
        return False

    cmd = 'multipathd -k"fail path %s"' % path
    if process.system(cmd) == 0:
        return wait.wait_for(is_failed, timeout=10) or False

```

Naresh-ibm: @apahim Thank you so much detail suggestion and help. tested this. working fine.
I have sent #2305 PR with the latest changes.
hence closing this.

PR #2217, opened 2017-09-20
Revert Test ID change

apahim:
We changed the wrong character. We should have changed the colon (`:`), but we ended up changing the semi-colon (`;`).

### DIFF
--- a/avocado/core/test.py
+++ b/avocado/core/test.py
@@ -83,7 +83,7 @@ class TestID(object):
             self.str_variant = ""
         else:
             self.variant = variant["variant_id"]
-            self.str_variant = "+%s" % self.variant
+            self.str_variant = ";%s" % self.variant
 
     def __str__(self):
         return "%s-%s%s" % (self.str_uid, self.name, self.str_variant)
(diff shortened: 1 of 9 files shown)


ldoktor:
Oh, my fault. I know we moved from `:` to `;` and back. I should have been more careful on review. The question is, @clebergnu, how far are you in the release? I see you already tagged it, can it be undone? Or can we only release 54.1 otherwise?

apahim:
@ldoktor, v54 packages are already published. Now we have 2 options:
- Revert; release v54.1; reopen the bug.
- Keep it as is; reopen the bug.

clebergnu:
Let's revert and release 54.1.

clebergnu:
@apahim @ldoktor I don't think we should rewrite history, I mean, rewrite the release notes of 54.0.  I'll write a 54.1 release notes doc, and in there it'll be documented that that change was reverted.  Do you guys agree?

clebergnu:
And alternative version has been pushed and tagged as 54.1.

PR #2234, opened 2017-10-03
test: log the raw output on stdout/stderr

apahim:
Python logging system is hard-coded to add a new line character at the
end of each call to emit(). Since the Avocado recording to test
stdout/stderr uses the logging system, this patch creates/uses a custom
FileHandler to be used in the cases where we need the raw output to
be recorded to the files.

Reference: https://trello.com/c/dH7FLT75
Signed-off-by: Amador Pahim <apahim@redhat.com>

PR #2238, opened 2017-10-04
log the raw output on stdout/stderr

apahim:
Avocado process module has a routine to drain the std* buffer with a
per-line basis loop. This loop makes the new-line character to be
dropped.

On the other hand, the Python logging system is hard-coded to add a new-line
character at the end of the msg string.

The problem takes place when `process` writes a line which originally
had new-line character at the end but the logging system still adds the
new-line character to that line.

In this patch:
- avocado.utils.process to include the new-line character in the lines
  that originally had one.
- avocado.core.test to use for raw outputs a custom logging FileHandler
  that does not add a new-line character at the end of each line.

Verified with (checking `.../test-results/1-.../[stdout|stderr|output]`):

    $ avocado run --external-runner '/bin/echo' 'foo'

    $ avocado run --external-runner '/bin/echo -n' 'foo'

    $ avocado run --external-runner '/bin/ls' '.'

    $ avocado run --external-runner '/bin/ls' 'nonexistingfile VERSION'

----
v2:
- `output` file to also use the `RawFileLogger`.
- `avocado.utils.process` to re-include the dropped new-line character to the lines where it applies.

v1: #2234

### DIFF
--- a/avocado/core/test.py
+++ b/avocado/core/test.py
@@ -52,6 +52,33 @@ from .output import LOG_JOB
 COMMON_TMPDIR_NAME = 'AVOCADO_TESTS_COMMON_TMPDIR'
 
 
+class RawFileHandler(logging.FileHandler):
+
+    """
+    File Handler that doesn't include arbitrary characters to the
+    logged stream but still respects the formatter.
+    """
+
+    def __init__(self, *args, **kwargs):
+        logging.FileHandler.__init__(self, *args, **kwargs)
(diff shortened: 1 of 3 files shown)


apahim:
[BUG: Output check recording adds a newline for SIMPLE tests](https://trello.com/c/dH7FLT75/1113-bug-output-check-recording-adds-a-newline-for-simple-tests)

PR #2245, opened 2017-10-06
Introducing the golang plugin

apahim:
golang ships a builtin test runner, which is called by
`go test <opts> <package or path>`. This patch introduces the golang
optional plugin, which can be used to discover and run golang tests
using Avocado.


apahim:
[Introduce Golang pluign](https://trello.com/c/velCpIvk/1143-introduce-golang-pluign)

docs/source/optional_plugins/golang.rst
```diff
@@ -0,0 +1,75 @@
+=============
+Golang Plugin
+=============
+
+This optional plugin enables Avocado to work with tests originally
```
clebergnu: Suggestion:

"This optional plugin enables Avocado to list and run tests written using ..."

The idea being that "originally", to me, sounds like tests have to change to work with Avocado.

docs/source/optional_plugins/golang.rst
```diff
@@ -0,0 +1,75 @@
...
+written using the `Go testing package <https://golang.org/pkg/testing/>`_.
+
+To install the Golang plugin from pip, use::
+
+    $ sudo pip install avocado-framework-plugin-golang
+
+After installed, you can list/run Golang tests providing the package name::
+
+    ~$ avocado list golang.org/x/text/unicode/norm
```
clebergnu: While running with an invalid reference:

```
$ avocado list /path/doesnot/exist
Test discovery plugin <avocado_golang.GolangLoader object at 0x7f4a6be46e90> failed: 'NoneType' object has no attribute 'endswith'
Unable to resolve reference(s) '/path/doesnot/exist' with plugins(s) 'file', 'golang', 'yaml_testsuite', 'robot', 'external', try running 'avocado list -V /path/doesnot/exist' to see the details.
```

The first line is most probably not a good idea.

optional_plugins/golang/avocado_golang/__init__.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def __init__(self, args, extra_params):
+        super(GolangLoader, self).__init__(args, extra_params)
+
+    def discover(self, url, which_tests=loader.DEFAULT):
+        avocado_suite = []
+        test_files = []
+        _subtest = None
+        tests_filter = None
+        go_src = os.path.join(os.environ.get('GOPATH'), 'src')
```
clebergnu: Maybe this should be something like:

```python
go_src = os.path.join(os.environ.get('GOPATH', os.path.expanduser('~/go'), 'src'))
```

From what I read, `~/go` is the default location.

apahim:
@clebergnu, yes, my fault. `GOROOT` is not being checked in this version. It will be present in V2. Thanks.

optional_plugins/golang/avocado_golang/__init__.py
apahim: Ack

docs/source/optional_plugins/golang.rst
apahim: Ops :)

docs/source/optional_plugins/golang.rst
apahim: Ok, good.

PR #2270, opened 2017-10-16
test: check-output to record diffs [V2]

apahim:
v2:
- Add the diffs to the log (in addition to the files).
- Remove the assert message.
- Use python logging (instead of `with open()`) to achieve the items above.
- Rename the diff files from `std*_expected.diff` to `std*.diff`.
- Docs.

v1: #2258

### DIFF
--- a/selftests/functional/test_output_check.py
+++ b/selftests/functional/test_output_check.py
@@ -1,3 +1,4 @@
+import json
 import os
 import tempfile
 import shutil
@@ -15,6 +16,7 @@ basedir = os.path.abspath(basedir)
 AVOCADO = os.environ.get("UNITTEST_AVOCADO_CMD", "./scripts/avocado")
 OUTPUT_SCRIPT_CONTENTS = """#!/bin/sh
 echo "Hello, avocado!"
+echo "Hello, stderr!" >&2
 """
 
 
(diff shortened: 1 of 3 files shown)


apahim:
[Check output functionality should also print a diff (expected/found)](https://trello.com/c/CXDP1khV/1106-check-output-functionality-should-also-print-a-diff-expected-found)

selftests/functional/test_output_check.py
```diff
@@ -98,6 +100,57 @@ def test_output_tamper_stdout(self):
...
+        self.test_output_record_all()
+        tampered_msg_stdout = "I PITY THE FOOL THAT STANDS ON STDOUT!"
+        tampered_msg_stderr = "I PITY THE FOOL THAT STANDS ON STDERR!"
+
+        stdout_file = os.path.join("%s.data/stdout.expected" %
+                                   self.output_script.path)
+        with open(stdout_file, 'w') as stdout_file_obj:
+            stdout_file_obj.write(tampered_msg_stdout)
+
+        stderr_file = os.path.join("%s.data/stderr.expected" %
```
clebergnu: No need for `os.path.join` here.

apahim: Oh, right! Updating.

apahim:
@clebergnu updated, including docs.

PR #2277, opened 2017-10-26
vmimage: make the list of providers expandable

apahim:
Creating a public interface to append custom providers to the list will
help users with private/internal image repositories not to edit the
source code in order to create their own providers.


apahim:
Sure, v2 is coming.

PR #2297, opened 2017-11-20
Travis fixes

apahim:
- Revert "travis: Use parallel check in travis"

    This reverts commit 4b29699.

    Parallel check was enabled in Travis, but it's overloading the system
    and making the time sensitive tests to FAIL.

- avocado.utils.process: fix default prefix

    Log prefix needs at least the %-style string formatter. Let's include
    that as the default so we don't break under 'no prefix'.


apahim:
Guys, fixes are in the same commit because travis will still fail with only one of them per commit.

PR #2303, opened 2017-11-24
Subtests filter fixes

apahim:
- Make subtests filter to work in directory discoveries.
- Make subtests filter to work for SIMPLE tests.

### DIFF
--- a/avocado/core/loader.py
+++ b/avocado/core/loader.py
@@ -875,9 +875,8 @@ class FileLoader(TestLoader):
                         for name in python_unittests]
             else:
                 if os.access(test_path, os.X_OK):
-                    # Module does not have an avocado test class inside but
-                    # it's executable, let's execute it.
-                    return self._make_test(test.SimpleTest, test_path)
+                    return self._make_test(test.SimpleTest, test_path,
+                                           subtests_filter=subtests_filter)
                 else:
                     # Module does not have an avocado test class inside, and
                     # it's not executable. Not a Test.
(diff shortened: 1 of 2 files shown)


apahim:
[[BUG] Test Reference filter not working for directories](https://trello.com/c/7VQ3syLt/1167-bug-test-reference-filter-not-working-for-directories)

avocado/core/loader.py
```diff
@@ -615,7 +615,8 @@ def skip_non_test(exception):
                             break
                     else:
                         pth = os.path.join(dirpath, file_name)
-                        tests.extend(self._make_tests(pth, which_tests))
+                        tests.extend(self._make_tests(pth, which_tests,
+                                                      subtests_filter))
```
clebergnu: Yep, LGTM.

avocado/core/loader.py
```diff
@@ -893,18 +892,22 @@ def _make_existing_file_tests(self, test_path, make_broken,
...
                 return make_broken(NotATest, test_path, self.__not_test_str)
 
     @staticmethod
-    def _make_test(klass, uid, description=None):
+    def _make_test(klass, uid, description=None, subtests_filter=None):
         """
         Create test template
         :param klass: test class
         :param uid: test uid (by default used as id and name)
         :param description: Description appended to "uid" (for listing purpose)
```
clebergnu: For consistency, let's add this new parameter to the docstring.

selftests/functional/test_loader.py
```diff
@@ -350,6 +350,19 @@ def test_python_unittest(self):
         self.assertFalse(exps, "Some expected result not matched to actual"
                          "results:\n%s\n\nexps = %s" % (result, exps))
 
+    def test_list_subtests_filter(self):
+        """
+        Check whether the subtests filter works for both INSTRUMENTED
+        and SIMPLE in a directory list.
+        """
+        cmd = "%s list examples/tests/:fail" % AVOCADO
+        result = process.run(cmd, ignore_status=True)
```
clebergnu: I don't see the reason for `ignore_status=True` here.  If avocado spits out a non-zero status, we should take notice.

avocado/core/loader.py
apahim: yes, sure

selftests/functional/test_loader.py
clebergnu: Yeah, but think about it: in theory the filter could generate some kind of unrelated error and still generate the expected output.  Better to be safe, as it costs nothing.

apahim: Sure thing.

PR #2304, opened 2017-11-24
correction of typo error in multipath.py

Naresh-ibm:
There is a typo error in command execution in multipath.py in
get_path_status() method. so corrected it and it is working fine now.


Naresh-ibm:
actual command is as follow
[root@ltciofvtr-hab1 utils]# multipathd -k"help" | grep -i json
 list|show maps|multipaths json
 list|show map|multipath $map json
[root@ltciofvtr-hab1 utils]# 

it is "json".

PR #2318, opened 2017-11-30
examples: fix doublefree test [v2]

apahim:
v2:
- Keep test pattern specific to `darwin`.

v1: #2316

### DIFF
--- a/examples/tests/doublefree.py
+++ b/examples/tests/doublefree.py
@@ -2,7 +2,6 @@
 
 import os
 import shutil
-import signal
 import sys
 
 from avocado import Test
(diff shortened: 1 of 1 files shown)


examples/tests/doublefree.py
```diff
@@ -41,15 +40,14 @@ def test(self):
         Execute 'doublefree'.
         """
         cmd = os.path.join(self.srcdir, 'doublefree')
-        cmd_result = process.run(cmd, ignore_status=True)
+        cmd_result = process.run(cmd, ignore_status=True,
+                                 env={'MALLOC_CHECK_': '1'})
         self.log.info(cmd_result)
-        expected_exit_status = -signal.SIGABRT
         output = cmd_result.stdout + cmd_result.stderr
-        self.assertEqual(cmd_result.exit_status, expected_exit_status)
```
lmr: It would be worth a small note in the commit message mentioning that `MALLOC_CHECK_ = 1` makes the process to not be aborted when using the GNU libc, rather simply printing the `invalid pointer` string to stdout. Either that or a comment on top of the `process.run` call.

Otherwise, it looks good, thanks!

apahim: @lmr the absense of `MALLOC_CHECK_ = 1` is not making the process to be aborted anymore (i.e. on Fedora 27), so that message would loose relevance really quickly.

lmr: Fair enough.

PR #2346, opened 2017-12-12
varianter_pict: enable higher verbosity

apahim:
the variater_pict plugin has no verbosity level other than the default,
which is not enough to expose the parameters content.

This patch enables higher verbosity so users can see the parameters
content when using `-c` or `--variants 2`(or above) in the `avocado
variants` command.


apahim:
[Implement verbose mode for PICT varianter](https://trello.com/c/Aio6IvOs/1171-implement-verbose-mode-for-pict-varianter)

PR #2347, opened 2017-12-12
Fix golang loader

apahim:
We should not discover the golang tests when we are not able to run them. This PR makes the golang plugin to skip the discovery process when the `go` bin is not present on the system.


apahim:
[BUG: Tests are discovered as broken GO when they are suppose to be discovered as other tests](https://trello.com/c/nc5bs8Ke/1156-bug-tests-are-discovered-as-broken-go-when-they-are-suppose-to-be-discovered-as-other-tests)

PR #2354, opened 2017-12-14
utils.crypto: Fix two errors in hash_file

luckyh:
Fix the following two errors in hash_file:

* A wrong indentation caused a dead loop
* Forgot returning `None` when the given hash algorithm is unknown

Signed-off-by: Xu Han <xuhan@redhat.com>

### DIFF
--- a/avocado/utils/crypto.py
+++ b/avocado/utils/crypto.py
@@ -66,11 +66,11 @@ def hash_file(filename, size=None, algorithm="md5"):
         while size > 0:
             if chunksize > size:
                 chunksize = size
-                data = file_to_hash.read(chunksize)
-                if len(data) == 0:
-                    logging.debug("Nothing left to read but size=%d", size)
-                    break
-                hash_obj.update(data)
-                size -= len(data)
+            data = file_to_hash.read(chunksize)
+            if len(data) == 0:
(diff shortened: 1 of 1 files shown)


luckyh:
@clebergnu @apahim please help review this ASAP, thanks.

avocado/utils/crypto.py
```diff
@@ -66,11 +66,11 @@ def hash_file(filename, size=None, algorithm="md5"):
         while size > 0:
```
apahim: The try/except clause above is broken. Mind fixing (in a separate commit)? Just add a `return None` right after the `logging.error(Unknown...`

avocado/utils/crypto.py
```diff
@@ -66,11 +66,11 @@ def hash_file(filename, size=None, algorithm="md5"):
...
             if chunksize > size:
                 chunksize = size
-                data = file_to_hash.read(chunksize)
-                if len(data) == 0:
-                    logging.debug("Nothing left to read but size=%d", size)
-                    break
-                hash_obj.update(data)
-                size -= len(data)
```
apahim: ok, checking

avocado/utils/crypto.py
luckyh: Ah, sure, will fix it.

PR #2393, opened 2018-01-04
spell: comply with new checks

apahim:
make spell is now covering more cases. This patch fixes the actual spell
errors and adds the required exceptions.


apahim:
[BUG: make spell broken](https://trello.com/c/Ibvuwl5V/1195-bug-make-spell-broken)

spell.ignore
```diff
@@ -453,3 +453,99 @@ dm
...
+lucas
+meneghel
+rodrigues
+lukas
+doktor
+ruda
+moura
+amador
+pahim
+rosa
```
ldoktor: nice section :-) (did you consider adding `# authors` section perhaps on top of this file?)

ldoktor: (I mean that is not a request, it's just a finishing touch which is not really needed)

spell.ignore
```diff
@@ -453,3 +453,99 @@ dm
...
+travis
+testclass
+extensionmanager
+jobpre
+jobpost
+joberror
+admanski
+basename
+bligh
+andy
```
ldoktor: those are also authors ;-)

spell.ignore
apahim: No, I didn't consider, but now that you mentioned, sounds like a good touch. I will send a V2 including that. Thanks.

ldoktor: As we discussed let's keep it this way

PR #2420, opened 2018-01-25
process.py: Add subprocess stdin.write() method

harish-24:
Add support to provide input with subprocess.Popen's stdin.write method. Optionally flush the input buffer.

### DIFF
--- a/avocado/utils/process.py
+++ b/avocado/utils/process.py
@@ -531,6 +531,7 @@ class SubProcess(object):
                 else:
                     stderr = subprocess.PIPE
                 self._popen = subprocess.Popen(cmd,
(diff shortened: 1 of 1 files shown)


avocado/utils/process.py
```diff
@@ -701,6 +702,16 @@ def wait(self):
             self._fill_results(rc)
         return rc
 
+    def stdin_write(self, value, flush=False):
+        """
+        Call the subprocess stdin.write() method for any input to subprocess
+        Optionally flush the input buffer
+        """
+        self._init_subprocess()
+        self._popen.stdin.write(value)
```
apahim: ok. eve though this is very prone to IOError, let the caller to handle that.

PR #2421, opened 2018-01-25
html: add warn in success rate

apahim:
Count WARN in to calculate the success rate in the HTML report.


apahim:
[The "pass rate" in html plugin does not include "warn" tests.](https://trello.com/c/6GsdbDLF/1169-the-pass-rate-in-html-plugin-does-not-include-warn-tests)

PR #2430, opened 2018-02-03
SIMPLE tests: improve status API [V2]

apahim:
v2:

v1: #2425

### DIFF
--- a/avocado/core/test.py
+++ b/avocado/core/test.py
@@ -1101,9 +1101,6 @@ class SimpleTest(Test):
 
     DATA_SOURCES = ["variant", "file"]
 
-    re_avocado_log = re.compile(r'^\d\d:\d\d:\d\d DEBUG\| \[stdout\]'
-                                r' \d\d:\d\d:\d\d WARN \|')
-
     def __init__(self, name, params=None, base_logdir=None, job=None):
         super(SimpleTest, self).__init__(name=name, params=params,
                                          base_logdir=base_logdir, job=job)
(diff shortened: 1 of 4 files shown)


apahim:
[SIMPLE tests good status based on regex](https://trello.com/c/jT4jtbOV/1237-simple-tests-good-status-based-on-regex)

avocado/core/test.py
```diff
@@ -1146,12 +1143,51 @@ def _execute_cmd(self):
...
+                                        'skip_regex',
+                                        key_type='str',
+                                        default=None)
+
+        skip_location = settings.get_value('simpletests.status',
+                                           'skip_location',
+                                           default='all')
+
+        # Keeping compatibility with 'avocado_warn' libexec
+        for regex in [warn_regex, r'^\d\d:\d\d:\d\d WARN \|']:
```
ldoktor: You only want to compile the regexps ones. Perhaps we could lazily do that even for all tests (global variable), what do you think?

apahim: Sure.

ldoktor: Btw it was just brainstorming, not a requirement 😉

avocado/etc/avocado/avocado.conf
```diff
@@ -84,3 +84,18 @@ skip_broken_plugin_notification = []
 # options or test types).
 # The keyword "@DEFAULT" will be replaced with all available unused loaders.
 loaders = ['file', '@DEFAULT']
+
+[simpletests.status]
+# Regular Expression that will make the test
```
ldoktor: Maybe it'd be worth explicitly saying `python regular expression`, maybe even giving there a link to regexp documentation, because this feature is for simple tests, therefor bash-oriented comunity.

docs/source/WritingTests.rst
```diff
@@ -1753,6 +1753,41 @@ by ``avocado exec-path`` (if any).  Also, the example test
...
          RPM packages, look for the ``bash`` sub-package.
 
+SIMPLE Tests Status
+===================
+
+With SIMPLE tests, Avocado checks the exit code of the test to determine
+whether the test PASSed or FAILed.
+
+If your test exits with exit code 0 but you still want to set a different test
+status in some conditions, Avocado can search a given regular expression in
```
ldoktor: the same here, I'd stress out it's python regexp and here I'd definitely put a link to regexp documentation.

selftests/functional/test_basic.py
```diff
@@ -858,6 +858,51 @@ def tearDown(self):
...
+                                                  "#!/bin/sh\necho WARN",
+                                                  'avocado_simpletest_'
+                                                  'functional')
+        self.warn_script.save()
+
+        self.skip_script = script.TemporaryScript('avocado_warn.sh',
+                                                  "#!/bin/sh\necho SKIP",
+                                                  'avocado_simpletest_'
+                                                  'functional')
+        self.skip_script.save()
```
ldoktor: You don't need to prepare both scripts for both tests.

selftests/functional/test_basic.py
```diff
@@ -858,6 +858,51 @@ def tearDown(self):
...
+                                                  "skip_regex = ^SKIP$\n")
+        self.config_file.save()
+
+        self.warn_script = script.TemporaryScript('avocado_warn.sh',
+                                                  "#!/bin/sh\necho WARN",
+                                                  'avocado_simpletest_'
+                                                  'functional')
+        self.warn_script.save()
+
+        self.skip_script = script.TemporaryScript('avocado_warn.sh',
```
ldoktor: nitpick: This should be called `avocado_skip.sh` and not `warn`.

selftests/functional/test_basic.py
```diff
@@ -858,6 +858,51 @@ def tearDown(self):
...
+
+    def test_simpletest_warn(self):
+        cmd_line = ('%s --config %s run --job-results-dir %s --sysinfo=off'
+                    ' %s --json -' % (AVOCADO, self.config_file.path,
+                                      self.tmpdir, self.warn_script.path))
+        result = process.system_output(cmd_line, ignore_status=True)
+        json_results = json.loads(result)
+        self.assertEquals(json_results['tests'][0]['status'], 'WARN')
+
+    def test_simpletest_skip(self):
```
ldoktor: Hmm, have you considered merging those two into a single test? For unittests it's useful to have better granularity but here it means 1s vs. 0.6s...

avocado/core/test.py
ldoktor: Actually please don't do this. I thought about this deeper and with JobAPI we might want to have different behavior for different SimpleTests in a single job and caching the regexp might be counterproductive.

PR #2432, opened 2018-02-05
html: rework success rate

apahim:
Success rate to use the following formula:

    ( PASS + WARN ) / ( TOTAL - SKIP - CANCEL ) * 100

### DIFF
--- a/optional_plugins/html/avocado_result_html/__init__.py
+++ b/optional_plugins/html/avocado_result_html/__init__.py
@@ -82,7 +82,8 @@ class ReportModel(object):
         return self.result.warned
 
     def rate(self):
-        total = float(self.result.tests_total)
+        total = float(self.result.tests_total - self.result.skipped -
+                      self.result.cancelled)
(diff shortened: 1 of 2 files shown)


apahim:
[The "pass rate" in html plugin does not include "warn" tests.](https://trello.com/c/6GsdbDLF/1169-the-pass-rate-in-html-plugin-does-not-include-warn-tests)

ldoktor:
Would you please rebase this on the current master so it passes the manual check?

apahim:
Updated

optional_plugins/html/avocado_result_html/__init__.py
```diff
@@ -82,7 +82,8 @@ def warned(self):
         return self.result.warned
 
     def rate(self):
-        total = float(self.result.tests_total)
+        total = float(self.result.tests_total - self.result.skipped -
+                      self.result.cancelled)
```
lmr: OK, makes sense.

PR #2437, opened 2018-02-09
Support variants dump/load

apahim:
Depending on the number of parameters, generating the Variants can be
compute intensive. This patch adds the ability to dump/load the variants
to/from a JSON file so the Variants generation can be offloaded from the
machine under test.

Reference: https://trello.com/c/llZJKBdi

### DIFF
--- a/avocado/plugins/run.py
+++ b/avocado/plugins/run.py
@@ -17,12 +17,14 @@ Base Test Runner Plugins.
 """
 
 import argparse
+import json
 import sys
 
 from avocado.core import exit_codes
 from avocado.core import job
 from avocado.core import loader
 from avocado.core import output
+from avocado.core import varianter
 from avocado.core.output import LOG_UI
 from avocado.core.plugin_interfaces import CLICmd
 from avocado.core.dispatcher import ResultDispatcher
@@ -89,6 +91,10 @@ class Run(CLICmd):
         parser.add_argument('--ignore-missing-references', choices=('on', 'off'),
                             help="Force the job execution, even if some of "
                             "the test references are not resolved to tests.")
+        parser.add_argument("--dump-variants", default=None,
+                            help="Dump the Variants to a JSON file")
+        parser.add_argument("--load-variants", default=None,
+                            help="Load the Variants from a JSON file")
 
         sysinfo_default = settings.get_value('sysinfo.collect',
                                              'enabled',
@@ -189,6 +195,16 @@ class Run(CLICmd):
 
         :param args: Command line args received from the run subparser.
         """
+        load_variants = getattr(args, 'load_variants', None)
+        if load_variants is not None:
+            try:
+                with open(load_variants, 'r') as var_file:
+                    setattr(args, "avocado_variants",
+                            varianter.Varianter(state=json.load(var_file)))
+            except IOError:
+                LOG_UI.error('Cannot read %s', load_variants)
+                sys.exit(exit_codes.AVOCADO_FAIL)
+
         if 'output_check_record' in args:
             process.OUTPUT_CHECK_RECORD_MODE = getattr(args,
                                                        'output_check_record',
@@ -224,4 +240,13 @@ class Run(CLICmd):
             result_dispatcher.map_method('render',
                                          job_instance.result,
                                          job_instance)
+        # Export the serialized avocado_variants
+        if args.dump_variants is not None:
+            try:
+                with open(args.dump_variants, 'w') as variants_file:
+                    json.dump(args.avocado_variants.dump(), variants_file)
+            except IOError:
+                LOG_UI.error("Cannot write %s", args.dump_variants)
+                sys.exit(exit_codes.AVOCADO_FAIL)
+
         return job_run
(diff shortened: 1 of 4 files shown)


avocado/plugins/run.py
```diff
@@ -89,6 +91,10 @@ def configure(self, parser):
         parser.add_argument('--ignore-missing-references', choices=('on', 'off'),
                             help="Force the job execution, even if some of "
                             "the test references are not resolved to tests.")
+        parser.add_argument("--dump-variants", default=None,
+                            help="Dump the Variants to a JSON file")
+        parser.add_argument("--load-variants", default=None,
+                            help="Load the Variants from a JSON file")
```
ldoktor: This is tricky, because it has the same issue as `--replay-ignore mux`. One might expect the loaded variants would be combined with the newly defined ones (which is not happening). Replay warns the user about this and I'm wondering whether we should do the same here.

avocado/plugins/run.py
```diff
@@ -89,6 +91,10 @@ def configure(self, parser):
         parser.add_argument('--ignore-missing-references', choices=('on', 'off'),
                             help="Force the job execution, even if some of "
                             "the test references are not resolved to tests.")
+        parser.add_argument("--dump-variants", default=None,
+                            help="Dump the Variants to a JSON file")
```
ldoktor: I don't think we need `--dump-variants` here, it might give false impression that it'd also store the tests associated with the execution and other things. Do you think people will run job just to produce the variants? I think having it in `variants` is enough.

avocado/plugins/variants.py
```diff
@@ -65,6 +66,8 @@ def configure(self, parser):
         parser.add_argument('-c', '--contents', action='store_true',
                             default=False, help="[obsoleted by --variants] "
                             "Shows the node content (variables)")
+        parser.add_argument('--dump-variants', default=None,
+                            help="Dump the Variants to a JSON file")
```
ldoktor: It'd be nice to have the `--load-variants` here as well to be able to list the dumped variants other way than just with `--dry-run`.

avocado/plugins/run.py
ldoktor: Actually with replay it's even more confusing:

```
avocado run --replay d3aa78d --load-variants mux-0.json 
Using src job Mux data only, use `--replay-ignore variants` to override them.
```
but it uses the variants from `--load-variants`.

I see 3 options:

1. The current (with warning in -h, maybe even on execution, not sure how important this is)
2. The current but adding detection of variants. We failed doing that before, but we can re-try. In case variants are detected together with `--load-variants` it should warn user about ignored variants.
3. Implementing this as a varianter plugin, therefor it'd simply add the loaded variants to already specified ones (therefor `-m foo.yaml --load-variants bar.json` will yield all variants from `foo.yaml` as well as from `bar.json`.

I'd probably favor (3) as it should be fairly simple and will allow best of all worlds.

docs/source/TestParameters.rst
```diff
@@ -137,6 +137,46 @@ is ``None``, which still produces an empty `AvocadoParams`_. Also, the
...
+Dumping/Loading Variants
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Depending on the number of parameters, generating the Variants can be very
+compute intensive. And since the Variants are generated as part of the Job
+execution, that compute intensive task will be executed by the systems under
+test, causing an unwanted cpu load on those systems.
+
+To avoid such situation, you can export (dump) the variants to a JSON file and
+use that file (load) on the system where the Job will be executed. Example:
```
ldoktor: Do you think it'd be useful to mention one can use `jobdata/variants.json` by `--load-variants`? I think it's useful...

avocado/plugins/run.py
```diff
@@ -224,4 +240,13 @@ def run(self, args):
             result_dispatcher.map_method('render',
                                          job_instance.result,
                                          job_instance)
+        # Export the serialized avocado_variants
+        if args.dump_variants is not None:
+            try:
+                with open(args.dump_variants, 'w') as variants_file:
+                    json.dump(args.avocado_variants.dump(), variants_file)
```
ldoktor: If you chose to keep the `--dump-variants` for `run` consider copying the `jobdata/variants.json` instead of serializing it again. It's quite costly operation compare to pure copy.

selftests/functional/test_variants.py
```diff
@@ -0,0 +1,75 @@
...
+            self.assertEqual(file_obj.read(), content)
+
+    def test_run_dump(self):
+        content = ('[{"paths": ["/run/*"], '
+                   '"variant": [["/", []]], '
+                   '"variant_id": null}]')
+
+        cmd_line = ('%s run passtest.py --dump-variants %s '
+                    '--job-results-dir %s' %
+                    (AVOCADO, self.variants_file, self.tmpdir))
```
ldoktor: you know me, I'm not a friend of duplicate tests, how about having `dump_and_run` test, check the dumped file and then run it? (it'd eliminate the one `avocado variants` execution but in case of failure it'd require manual check whether `run` works well in case of `dump` failure). Anyway this is really minor question...

avocado/plugins/run.py
apahim: Yes, I was unsure about this. Let me remove it and and we can re-consider in case some user asks for it.

apahim:

PR #2475, opened 2018-02-23
Support variants dump/load [v4]

apahim:
v4:
- Implement  `__len__` on `avocado.core.varianter.Varianter`
- On `avocado.core.varianter.Varianter.parse()`, handle `None` from `self._variant_plugins.map_method("__len__")`
- Move `json_variants` from optional_plugins to `avocado.plugins`, unifying tests and documentation.
-  `json_variants.JsonVariants.__len__()` to return `None` before initialization.

v3: #2467 
- Better naming.
- Tests inside plugin module.
- Separate docs for plugin.

v2: #2459 

v1: #2437

### DIFF
--- a/avocado/plugins/json_variants.py
+++ b/avocado/plugins/json_variants.py
@@ -1,0 +1,136 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: Red Hat Inc. 2018
+# Authors: Amador Pahim <apahim@redhat.com>
+
+import json
+import sys
+
+from six import iteritems
+
+from avocado.core import exit_codes
+from avocado.core import varianter
+from avocado.core.output import LOG_UI
+from avocado.core.plugin_interfaces import CLI
+from avocado.core.plugin_interfaces import Varianter
+
+
+class JsonVariantsCLI(CLI):
+
+    """
+    Serialized based Varianter options
+    """
+
+    name = 'json variants'
+    description = "JSON serialized based Varianter options for the 'run' " \
+                  "subcommand"
+
+    def configure(self, parser):
+
+        for name in ("run", "variants"):  # intentionally omitting "multiplex"
+            subparser = parser.subcommands.choices.get(name, None)
+            if subparser is None:
+                continue
+            sparser = subparser.add_argument_group('JSON serialized based '
+                                                   'varianter options')
+            sparser.add_argument('--json-variants-load', default=None,
+                                 help=('Load the Variants from a JSON '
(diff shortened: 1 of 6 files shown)


apahim:

avocado/plugins/json_variants.py
```diff
@@ -0,0 +1,136 @@
...
+        """
+        if not self.variants:
+            return ""
+
+        out = []
+        verbose = variants > 1
+        if summary:
+            # TODO: tree representation
+            out.append("No tree representation for JSON serialized variants")
+        else:
```
ldoktor: This condition is incorrect, it should be `if summary` followed by `if variants`, because both can be at the same time.

ldoktor: (see `optional_plugins/varianter_yaml_to_mux/avocado_varianter_yaml_to_mux/mux.py` for details.)

avocado/plugins/json_variants.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def __iter__(self):
+        if self.variants is None:
+            return
+
+        for variant in self.variants.itertests():
+            yield variant
+
+    def __len__(self):
+        if self.variants is None:
```
ldoktor: As mentioned during the meeting I don't think this is a correct approach. Using `__len__` before `initialize` can't be supported as we don't know the value. How about something like this:

```diff
diff --git a/avocado/plugins/json_variants.py b/avocado/plugins/json_variants.py
index 077451d3..b87c2b98 100644
--- a/avocado/plugins/json_variants.py
+++ b/avocado/plugins/json_variants.py
@@ -24,6 +24,9 @@ from avocado.core.plugin_interfaces import CLI
 from avocado.core.plugin_interfaces import Varianter
 
 
+_NO_VARIANTS = -1
+
+
 class JsonVariantsCLI(CLI):
 
     """
@@ -64,6 +67,7 @@ class JsonVariants(Varianter):
         load_variants = getattr(args, "json_variants_load", None)
 
         if load_variants is None:
+            self.variants = _NO_VARIANTS
             return
         try:
             with open(load_variants, 'r') as var_file:
@@ -77,15 +81,19 @@ class JsonVariants(Varianter):
                 sys.exit(exit_codes.AVOCADO_FAIL)
 
     def __iter__(self):
-        if self.variants is None:
+        if self.variants == _NO_VARIANTS:
             return
+        elif self.variants is None:
+            raise RuntimeError("__iter__ executed without 'initialize'")
 
         for variant in self.variants.itertests():
             yield variant
 
     def __len__(self):
-        if self.variants is None:
-            return None
+        if self.variants == _NO_VARIANTS:
+            return 0
+        elif self.variants is None:
+            raise RuntimeError("__len__ executed without 'initialize'")
 
         return len(self.variants)
```

(obviously with better exceptions)

avocado/plugins/json_variants.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+class JsonVariants(Varianter):
+
+    """
+    Processes the serialized file into variants
+    """
+
+    name = 'json variants'
+    description = "JSON serialized based Varianter"
```
ldoktor: Please define `variants = None` here instead of in `initialize`.

apahim:
@ldoktor, oh, I think finally got your point. I'm ok in differentiating the non-initialized Varianter from the initialized one without variants. Sensing a new version. Tks.

PR #2480, opened 2018-02-23
Support variants dump/load [v5]

apahim:
v5:
- Raise exceptions when Varianter methods `__len__` or `__iter__` are used before initialization.
- Fix summary/variants checks in `to_str()`.
- Drop unneeded commit `varianter: expect plugins to return len() as None at parse time`

v4: #2475 
- Implement  `__len__` on `avocado.core.varianter.Varianter`
- On `avocado.core.varianter.Varianter.parse()`, handle `None` from `self._variant_plugins.map_method("__len__")`
- Move `json_variants` from optional_plugins to `avocado.plugins`, unifying tests and documentation.
-  `json_variants.JsonVariants.__len__()` to return `None` before initialization.

v3: #2467 
- Better naming.
- Tests inside plugin module.
- Separate docs for plugin.

v2: #2459 

v1: #2437

### DIFF
--- a/avocado/core/varianter.py
+++ b/avocado/core/varianter.py
@@ -346,3 +346,6 @@ class Varianter(object):
             yield {"variant": self._default_params.get_leaves(),
                    "variant_id": None,
                    "paths": ["/run/*"]}
+
+    def __len__(self):
+        return self._no_variants
--- a/avocado/plugins/json_variants.py
+++ b/avocado/plugins/json_variants.py
@@ -1,0 +1,147 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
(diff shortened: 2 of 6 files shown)


avocado/core/varianter.py
```diff
@@ -346,3 +346,6 @@ def itertests(self):
             yield {"variant": self._default_params.get_leaves(),
                    "variant_id": None,
                    "paths": ["/run/*"]}
+
+    def __len__(self):
+        return self._no_variants
```
ldoktor: Hmm, shouldn't this also raise exception in case `_no_variants` is `None`? Well for now I'm fine with this as it should be obvious that `None` is not really good value.

avocado/plugins/json_variants.py
```diff
@@ -0,0 +1,147 @@
...
+                sys.exit(exit_codes.AVOCADO_JOB_FAIL)
+            else:
+                sys.exit(exit_codes.AVOCADO_FAIL)
+
+    def __iter__(self):
+        if self.variants == _NO_VARIANTS:
+            return
+        elif self.variants is None:
+            raise RuntimeError("Iterating Varianter before initialization is"
+                               "not supported")
```
ldoktor: Hmm, is it not supported, or not allowed? If it's not (yet) supported, than it should be `NotImplementedError`. If it's not allowed the message should be different, but that should not block the merge...

clebergnu:

PR #2481, opened 2018-02-23
Introducing the GLib Test Plugin

apahim:
Now users can run in Avocado the tests created with the GLib Test
Framework with the same level of granularity as they have when executing
the tests directly.


apahim:
@clebergnu, @ldoktor, due to the gtester but report and subsequent information that gtester is deprecated, I figured out that we don't really need the gtester to list/run the GLib tests. Every test accepts the `-l` and `-p` parameters, so we can drop that dependency.

apahim:
[Create a "GLib" optional plugin](https://trello.com/c/RzM2EvBg/1252-create-a-glib-optional-plugin)

clebergnu:
Your description makes a lot of sense... the glib library itself provides those features, when linked to the (test) binaries, those are readily available.

Anyway, to make Travis happy, you need to rebase this to the latest master, as Travis now requires unittests to pass on Python 3 (and this branch has 3dbc31c7c59a4d68bf349d74b3bb5e8d31757afb as its base).

apahim:
@clebergnu, rebased.

PR #2486, opened 2018-02-26
glib: adapt to changes in SimpleTest

apahim:
Now there's a new parameter `executable` to use as the command line
information. The previous parameter, `name` can now be hijacked by
the yaml_loader, making this plugin to break.

This patch adapts to that change of internal API.


apahim:
Closing to include all the commits in one PR only.

PR #2487, opened 2018-02-26
Adapt optional_plugins to SimpleTest changes

apahim:
Now there's a new parameter `executable` to use as the command line
information. The previous parameter, `name` can now be hijacked by
the yaml_loader, so this PR addresses that change in all the affected
optional_plugins.

### DIFF
--- a/optional_plugins/glib/avocado_glib/__init__.py
+++ b/optional_plugins/glib/avocado_glib/__init__.py
@@ -32,21 +32,23 @@ class GLibTest(test.SimpleTest):
     Run a GLib test command as a SIMPLE test.
     """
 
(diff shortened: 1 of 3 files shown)


clebergnu:
@apahim I got to crash the `GolangTest` class when manually supplying a `name` (`TestID` instance) but without an `executable` parameter.  I'll double check in the morning if my "test code" is to blame, or if there's indeed something wrong here.

apahim:
@clebergnu, could you please share the mentioned test code?

clebergnu:
[[BUG] Optinal Plugins should adapt changes in executable path](https://trello.com/c/Bx5M9XDC/1261-bug-optinal-plugins-should-adapt-changes-in-executable-path)

PR #2494, opened 2018-03-07
autotest: Move the function from autotest.client.shared.utils

hsmj1412:

### DIFF
--- a/avocado/utils/output.py
+++ b/avocado/utils/output.py
@@ -22,6 +22,7 @@ Utility functions for user friendly display of information.
 """
 
 import sys
+import re
 
 
 def display_data_size(size):
@@ -42,6 +43,28 @@ def display_data_size(size):
         i += 1
 
     return '%.2f %s' % (size, prefixes[i])
+
+
+def convert_data_size(size, default_sufix='B'):
+    '''
+    Convert data size from human readable units to an int of arbitrary size.
+    :param size: Human readable data size representation (string).
+    :param default_sufix: Default sufix used to represent data.
+    :return: Int with data size in the appropriate order of magnitude.
+    '''
+    orders = {'B': 1,
+              'K': 1024,
+              'M': 1024 * 1024,
+              'G': 1024 * 1024 * 1024,
+              'T': 1024 * 1024 * 1024 * 1024,
+              }
+
+    order = re.findall("([BbKkMmGgTt])", size[-1])
+    if not order:
+        size += default_sufix
+        order = [default_sufix]
+
+    return int(float(size[0:-1]) * orders[order[0].upper()])
 
 
 class ProgressBar(object):


avocado/utils/output.py
```diff
@@ -44,6 +45,28 @@ def display_data_size(size):
     return '%.2f %s' % (size, prefixes[i])
 
 
+def convert_data_size(size, default_sufix='B'):
+    '''
+    Convert data size from human readable units to an int of arbitrary size.
```
apahim: arbitrary size?

avocado/utils/output.py
```diff
@@ -44,6 +45,28 @@ def display_data_size(size):
...
+    :param size: Human readable data size representation (string).
+    :param default_sufix: Default sufix used to represent data.
+    :return: Int with data size in the appropriate order of magnitude.
```
apahim: By `appropriate order of magnitude` you mean `bytes`? Why not just say `in bytes`?

avocado/utils/output.py
```diff
@@ -44,6 +45,28 @@ def display_data_size(size):
...
+    '''
+    orders = {'B': 1,
+              'K': 1024,
+              'M': 1024 * 1024,
+              'G': 1024 * 1024 * 1024,
+              'T': 1024 * 1024 * 1024 * 1024,
```
apahim: No need to calculate this every time. You can calculate them beforehand and add a comment with the formula.
```python
orders = {'B': 1,
          'K': 1024,
          'M': 1048576, # 1024 * 1024
...
```

avocado/utils/output.py
```diff
@@ -44,6 +45,28 @@ def display_data_size(size):
...
+              }
+
+    order = re.findall("([BbKkMmGgTt])", size[-1])
+    if not order:
+        size += default_sufix
+        order = [default_sufix]
+
+    return int(float(size[0:-1]) * orders[order[0].upper()])
+
+
```
apahim: Seems like it's time to create a DataSize object, as discussed here:
https://github.com/avocado-framework/avocado/pull/2142#discussion_r133494278

hsmj1412:
@apahim Sorry, I just copy this function from autotest because this function is used in avocado-vt and tp-libvirt. So  I didn't check the annotation and  I do not want to update any in code as I do not want to involve other problems unless you think it is necessary

apahim:
@hsmj1412, yes, it is necessary. Regardless where the code comes from, it will be properly reviewed here. You can skip my last comment, about the DataSize object. It was not a request and I will handle that myself, but please address the other comments.

apahim:
Btw, this should be in `avocado.utils.data_structures`, not in `avocado.utils.output`.

apahim:
Closed in favour of:
- https://github.com/avocado-framework/avocado/pull/2495
- https://github.com/avocado-framework/avocado-vt/pull/1385

PR #2511, opened 2018-03-14
test results: optimize params information

apahim:
Currently we propagate to test results the AvacadoParams object used by
the test. Holding that object in the test results makes the Avocado
process to consume a lot of memory.

For instance, in an Avocado job with 5000 tests (passtest.py), with
Avocado-VT in place, the Job object grows up from 59MB before running
the tests to 274MB after running the tests, a 215MB growth::

    (Pdb) pre_tests
    Partition of a set of 440754 objects. Total size = 59066528 bytes.

    (Pdb) post_tests
    Partition of a set of 1570821 objects. Total size = 274340184 bytes.

    (Pdb) growth
    Partition of a set of 1130073 objects. Total size = 215274464 bytes.

Since the 'params' item is important for both the HTML report and the
ResultsDB report, instead of dropping the 'params' item from the
results, this patch parses the AvocadoParams, populating the test
results with a list of tuples, each one containing the 'path', 'key' and
'value' per parameter basis. This improves the situation a lot. The same
Avocado Job will now only grow up from 59MB to 89MB::

    (Pdb) pre_tests
    Partition of a set of 440747 objects. Total size = 59066424 bytes.

    (Pdb) post_tests
    Partition of a set of 840812 objects. Total size = 89619904 bytes.

    (Pdb) growth
    Partition of a set of 400071 objects. Total size = 30554288 bytes.

The HTML and ResultsDB plugins were adapted accordingly.

### DIFF
--- a/avocado/core/test.py
+++ b/avocado/core/test.py
@@ -60,7 +60,7 @@ TEST_STATE_ATTRIBUTES = ('name', 'logdir', 'logfile',
                          'status', 'running', 'paused',
                          'time_start', 'time_elapsed', 'time_end',
                          'fail_reason', 'fail_class', 'traceback',
-                         'params', 'timeout', 'whiteboard')
+                         'timeout', 'whiteboard')
 
 
 class RawFileHandler(logging.FileHandler):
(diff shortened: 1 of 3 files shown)


avocado/core/test.py
```diff
@@ -630,6 +630,12 @@ def get_state(self):
         state['class_name'] = self.__class__.__name__
         state['job_logdir'] = self.job.logdir
         state['job_unique_id'] = self.job.unique_id
+        params = getattr(self, 'params', None)
```
clebergnu: Actually, getting a `None` here can be currently be considered an erroneous situation, because `params` is not an optional attribute or value of the `avocado.core.Test` class.   I recommend dropping this and the `if params is not None` check altogether.

clebergnu:
Closing as v2 was merged.

apahim:
[Optimize Avocado](https://trello.com/c/GioaaT1y/1275-optimize-avocado)

Issue #2524, opened 2018-03-19
Memory utility has static values after import 


harish-24:
With the recent introduction of Datasize patch in `avocado/utils/memory.py`, once it has been imported in a test, the values throughout the tests are persistent. This is probably due to object initialized when import and usage in test does not reflect new values and persists to values when initialized.

With the change, thp swap test https://github.com/avocado-framework-tests/avocado-misc-tests/blob/master/memory/transparent_hugepages_swapping.py has following issue.

Ex: Assuming we have a `memory.meminfo.SwapFree.m=2047` .
After some operations actual SwapFree is 1127

But with the latest changes in the utility, the value persists to old one
`memory.meminfo.SwapFree.m=2047`

My understanding is that this should be fixed in utility, as initializing `MemInfo` object from test does not justify proper usage of the utility. If not, please let know if tests has to adapt to these changes.

apahim:
Yes, it should be fixed indeed. Working on that.

harish-24:
@apahim Thanks :+1:

harish-24:
Closing this as #2525 fixes this.

PR #2525, opened 2018-03-19
meminfo: attributes should return current values

apahim:
The last change in meminfo object to use the new DataSize object
introduced a wrong behaviour. Currently meminfo is not returning updated
values. This patch fixes it.


apahim:
Fixes https://github.com/avocado-framework/avocado/issues/2524
@harish-24, your feedback is welcomed here.

PR #2527, opened 2018-03-19
Extend avocado/avocado/utils/kernel.py to include kernel install.

tambewilliam:
Extend avocado/avocado/utils/kernel.py to include kernel install.

Signed-off-by: William Tambe <william.tambe@amd.com>

### DIFF
--- a/avocado/utils/kernel.py
+++ b/avocado/utils/kernel.py
@@ -19,7 +19,7 @@ import logging
 import tempfile
 from distutils.version import LooseVersion  # pylint: disable=E0611
 
-from . import asset, archive, build
+from . import asset, archive, build, distro, process
 
 log = logging.getLogger('avocado.test')
 
@@ -46,6 +46,7 @@ class KernelBuild(object):
         """
         self.version = version
         self.config_path = config_path
+        self.distro = distro.detect()
         if work_dir is None:
             work_dir = tempfile.mkdtemp(prefix='avocado_' + __name__)
         self.work_dir = work_dir
@@ -62,13 +63,21 @@ class KernelBuild(object):
                                               self.config_path,
                                               self.work_dir)
 
-    def download(self):
+    def download(self, url=None):
         """
         Download kernel source.
+
+        :param url: override the url from where to fetch the kernel
+                    source tarball
+        :type url: str or None
         """
-        self.kernel_file = self.SOURCE.format(version=self.version)
-        base_url = self.URL.format(major=self.version.split('.', 1)[0])
-        full_url = base_url + self.kernel_file
+        kernel_file = self.SOURCE.format(version=self.version)
+        if url is not None:
+            base_url = self.URL.format(major=self.version.split('.', 1)[0])
+        else:
+            base_url = url
+        full_url = base_url + kernel_file
+
         self.asset_path = asset.Asset(full_url, asset_hash=None,
                                       algorithm=None, locations=None,
                                       cache_dirs=self.data_dirs).fetch()
@@ -85,21 +94,40 @@ class KernelBuild(object):
         Configure/prepare kernel source to build.
         """
         self.linux_dir = os.path.join(self.work_dir, 'linux-%s' % self.version)
-        build.make(self.linux_dir, extra_args='O=%s mrproper' % self.build_dir)
+        build.make(self.linux_dir, extra_args='-C %s mrproper' % self.linux_dir)
         if self.config_path is not None:
             dotconfig = os.path.join(self.linux_dir, '.config')
             shutil.copy(self.config_path, dotconfig)
 
-    def build(self):
+    def build(self, binary_package=False):
         """
         Build kernel from source.
+
+        :param binary_package: when True, the appropriate
+                                  platform package is built
+                                  for install() to use
+        :type binary_pacakge: bool
         """
         log.info("Starting build the kernel")
+        build_output_format = ""
+        if binary_package is True:
+            if self.distro.name == "Ubuntu":
+                build_output_format = "deb-pkg"
         if self.config_path is None:
-            build.make(self.linux_dir, extra_args='O=%s defconfig' % self.build_dir)
+            build.make(self.linux_dir, extra_args='-C %s defconfig' % self.linux_dir)
         else:
-            build.make(self.linux_dir, extra_args='O=%s olddefconfig' % self.build_dir)
-        build.make(self.linux_dir, extra_args='O=%s' % self.build_dir)
+            build.make(self.linux_dir, extra_args='-C %s olddefconfig' % self.linux_dir)
+        build.make(self.linux_dir, extra_args='-C %s %s' % (self.linux_dir, build_output_format))
+
+    def install(self):
+        """
+        Install built kernel.
+        """
+        log.info("Starting kernel install")
+        if self.distro.name == "Ubuntu":
+            process.run('dpkg -i %s/*.deb' % self.work_dir, shell=True, sudo=True)
+        else:
+            log.info("Skipping kernel install")
 
     def __del__(self):
         shutil.rmtree(self.work_dir)
--- a/examples/tests/linuxbuild.py
+++ b/examples/tests/linuxbuild.py
@@ -17,8 +17,10 @@ class LinuxBuildTest(Test):
     """
 
     def setUp(self):
+        kernel_src_url = self.params.get('linux_src_url', default='https://www.kernel.org/pub/linux/kernel/v3.x')
         kernel_version = self.params.get('linux_version', default='3.19.8')
         linux_config = self.params.get('linux_config', default=None)
+        self.do_kernel_install = self.params.get('do_kernel_install', default=None)
         if linux_config is not None:
             linux_config = self.get_data(linux_config)
         if linux_config is None:
@@ -26,14 +28,16 @@ class LinuxBuildTest(Test):
 
         self.linux_build = kernel.KernelBuild(kernel_version,
                                               linux_config,
-                                              self.workdir,
+                                              self.srcdir,
                                               self.cache_dirs)
-        self.linux_build.download()
+        self.linux_build.download(kernel_src_url)
         self.linux_build.uncompress()
         self.linux_build.configure()
 
     def test(self):
-        self.linux_build.build()
+        self.linux_build.build(True if self.do_kernel_install is not None else False)
+        if self.do_kernel_install is not None:
+            self.linux_build.install()
 
 
 if __name__ == "__main__":
--- a/examples/tests/linuxbuild.py.data/linuxbuild.yaml
+++ b/examples/tests/linuxbuild.py.data/linuxbuild.yaml
@@ -1,0 +1,4 @@
+linux_src_url: https://www.kernel.org/pub/linux/kernel/v4.x/
+linux_version: 4.15.1
+linux_config: /boot/config-4.13.0-31-generic
+#do_kernel_install: y


avocado/utils/kernel.py
```diff
@@ -18,8 +18,9 @@
 import logging
 import tempfile
 from distutils.version import LooseVersion  # pylint: disable=E0611
+from avocado.utils import distro
```
clebergnu: This is actually flagged by the "boundary check":

https://travis-ci.org/avocado-framework/avocado/jobs/355619977#L1518

Please use:

```python
from . import distro
```

avocado/utils/kernel.py
```diff
@@ -34,7 +35,7 @@ class KernelBuild(object):
     SOURCE = 'linux-{version}.tar.gz'
 
     def __init__(self, version, config_path=None, work_dir=None,
-                 data_dirs=None):
+                 data_dirs=None, src_url, do_install=None):
```
clebergnu: IMO there's no need to have this as a class parameter.  Users are free to either call (or not) the `install()` method if they choose to have the kernel installed.

avocado/utils/kernel.py
```diff
@@ -44,15 +45,18 @@ def __init__(self, version, config_path=None, work_dir=None,
...
         self.config_path = config_path
+        self.distro = distro.detect()
         if work_dir is None:
             work_dir = tempfile.mkdtemp(prefix='avocado_' + __name__)
         self.work_dir = work_dir
         if data_dirs is not None:
             self.data_dirs = data_dirs
         else:
             self.data_dirs = [self.work_dir]
+        self.do_install=do_install
```
clebergnu: As per my previous comment, I believe there's no need to set this as a class parameter, and then as an attribute.  But, here goes a style tip: when doing other assignments, please add a space separating the LHS, equal sign, and RHS.

avocado/utils/kernel.py
```diff
@@ -85,7 +89,7 @@ def configure(self):
         Configure/prepare kernel source to build.
         """
         self.linux_dir = os.path.join(self.work_dir, 'linux-%s' % self.version)
-        build.make(self.linux_dir, extra_args='O=%s mrproper' % self.build_dir)
+        build.make(self.linux_dir, extra_args='-C %s mrproper' % self.linux_dir)
```
clebergnu: It looks like this is actually fixing a bug.  Did you have success building a kernel with the current version of this library?

avocado/utils/kernel.py
```diff
@@ -95,11 +99,26 @@ def build(self):
...
+            if self.distro.name == "Ubuntu":
+                build_output_format = "deb-pkg"
         if self.config_path is None:
-            build.make(self.linux_dir, extra_args='O=%s defconfig' % self.build_dir)
+            build.make(self.linux_dir, extra_args='-C %s defconfig' % self.linux_dir)
         else:
-            build.make(self.linux_dir, extra_args='O=%s olddefconfig' % self.build_dir)
-        build.make(self.linux_dir, extra_args='O=%s' % self.build_dir)
+            build.make(self.linux_dir, extra_args='-C %s olddefconfig' % self.linux_dir)
+        build.make(self.linux_dir, extra_args='-C %s %s' % (self.linux_dir, build_output_format))
```
clebergnu: IIUC it correctly, you can just build the usual kernel (the default target) on `build()`, and then, on install, you can build the distro specific target (`build_output_format`), which will reuse the previous work done on `build()`.

avocado/utils/kernel.py
```diff
@@ -95,11 +99,26 @@ def build(self):
...
+
+    def install(self):
+        """
+        Install built kernel.
+        """
+        if self.do_install is not None:
+            log.info("Starting kernel install")
+            if self.distro.name == "Ubuntu":
+                process.run('sh -c "exec sudo dpkg -i %s/*.deb"' % self.work_dir)
```
clebergnu: You should be able to use `process.run()` with the `sudo` and `shell` arguments here to simplify this command.

avocado/utils/kernel.py
```diff
@@ -95,11 +99,26 @@ def build(self):
...
+        else:
+            log.info("Skipping kernel install")
```
clebergnu: It'd be nice to also include generic kernel installation code.  There's no need to guarantee that it would work on every single system out there, but it feels wrong to add a method that will only work on a given system/package type.

examples/tests/linuxbuild.py
```diff
@@ -17,24 +17,31 @@ class LinuxBuildTest(Test):
     """
 
     def setUp(self):
+        kernel_src_url = self.params.get('linux_src_url', default=None)
         kernel_version = self.params.get('linux_version', default='3.19.8')
         linux_config = self.params.get('linux_config', default=None)
+        self.do_kernel_install = self.params.get('do_kernel_install', default=None)
+        if kernel_src_url is None:
```
clebergnu: The `avocado.utils.kernel` library is supposed to be able to use the default kernel.org repo to download sources based on the given version number.  This effectively disables that capability.  The kernel source URL should be given to `kernel.KernelBuild` if present in the parameters, or the default should be used.

examples/tests/linuxbuild.py.data/linuxbuild.yaml
```diff
@@ -0,0 +1,5 @@
+
```
clebergnu: There's no need for this first white line.

avocado/utils/kernel.py
tambewilliam: Thank you; making the correction.

avocado/utils/kernel.py
tambewilliam: Yes, I tested that I could successfully build the kernel with the current version of this library.

avocado/utils/kernel.py
tambewilliam: Yes; that is correct.

avocado/utils/kernel.py
tambewilliam: We (here at AMD) planned to extend the functionality to other distribution, as we included them in our test-infrastructure; we just happened to have started working with Ubuntu.

examples/tests/linuxbuild.py.data/linuxbuild.yaml
tambewilliam: Making the correction.

examples/tests/linuxbuild.py
tambewilliam: The issue we ran into is that all kernel versions are not available from a single url. ei:
For v3.x kernels, the following url must be used:
https://www.kernel.org/pub/linux/kernel/v3.x/
For v4.x kernels, the following url must be used:
https://www.kernel.org/pub/linux/kernel/v4.x/

Due to having to use different url for different major kernel versions, we thought providing the url as parameter could be one way around.

avocado/utils/kernel.py
tambewilliam: Should we turn this as method parameter instead ?

avocado/utils/kernel.py
tambewilliam: Should we turn this as method parameter ?

avocado/utils/kernel.py
clebergnu: I'd move `src_url` to `download()`, and make it optional, something like:

```diff
diff --git a/avocado/utils/kernel.py b/avocado/utils/kernel.py
index 654530ee..30a11b00 100644
--- a/avocado/utils/kernel.py
+++ b/avocado/utils/kernel.py
@@ -62,13 +62,20 @@ class KernelBuild(object):
                                               self.config_path,
                                               self.work_dir)
 
-    def download(self):
+    def download(self, url=None):
         """
         Download kernel source.
+
+        :param url: override the url from where to fetch the kernel
+                    source tarball
+        :type url: str or None
         """
-        self.kernel_file = self.SOURCE.format(version=self.version)
-        base_url = self.URL.format(major=self.version.split('.', 1)[0])
-        full_url = base_url + self.kernel_file
+        kernel_file = self.SOURCE.format(version=self.version)
+        if url is not None:
+            base_url = self.URL.format(major=self.version.split('.', 1)[0])
+        else:
+            base_url = url
+        full_url = base_url + kernel_file
         self.asset_path = asset.Asset(full_url, asset_hash=None,
                                       algorithm=None, locations=None,
                                       cache_dirs=self.data_dirs).fetch()
```

That way, it won't be required, but still can be overridden.

avocado/utils/kernel.py
clebergnu: I don't understand how this can be added as a method (which one?) parameter.  IMO this fits into `install()`, but it wouldn't need any extra parameter.

avocado/utils/kernel.py
clebergnu: OK... please forgive me, but would you please elaborate on the reason for this change? I'm confused because `build.make()` will `chdir()` into `self.linux_dir` anyway, so `-C` + `self.linux_dir` as an argument to `make` seems to unnecessary.  The only change I can see here is that before this change, the build output would go to `build_dir` and now it seems that it'll go to `linux_dir`, but I don't get why this is necessary or a positive change.

avocado/utils/kernel.py
clebergnu: OK, that's fine.

examples/tests/linuxbuild.py
clebergnu: See my previous comment about the `url` parameter to `download()`.  Also, the default `kernel.KernelBuild` will get the right URL based on the kernel version.  This was fixed recently:

http://avocado-framework.readthedocs.io/en/latest/release_notes/59_0.html#bug-fixes

avocado/utils/kernel.py
tambewilliam: sorry for lack of details in my previous question.
I meant to ask whether the do_install flag can be implemented as an argument of the method build(). ei:

```diff
diff --git a/avocado/utils/kernel.py b/avocado/utils/kernel.py
index 37b41d2f..4e93de7d 100644
--- a/avocado/utils/kernel.py
+++ b/avocado/utils/kernel.py
@@ -100,13 +100,17 @@ class KernelBuild(object):
             dotconfig = os.path.join(self.linux_dir, '.config')
             shutil.copy(self.config_path, dotconfig)

-    def build(self):
+    def build(self, build_for_install=none):
         """
         Build kernel from source.
+
+        :param build_for_install: when not None, the appropriate
+                                  platform package is built
+                                  for install() to use
         """
         log.info("Starting build the kernel")
         build_output_format = ""
-        if self.do_install is not None:
+        if build_for_install is not None:
             if self.distro.name == "Ubuntu":
                 build_output_format = "deb-pkg"
         if self.config_path is None:
```

avocado/utils/kernel.py
tambewilliam: Did you mean "without" the `sudo` and `shell` arguments ?

If yes:
- Most systems require admin right to install a package; in our test-infrastructure, we use passwordless-sudo for automation and avoid password prompt (ei: `%sudo   ALL=(ALL:ALL) NOPASSWD: ALL`).
- The shell was invoked for two reasons:
   So that the paths to sudo and dpkg would be automatically resolved for execution.
   So that wildcart *.deb would be resolved, as the result of the kernel-build generate more than one packages to install: linux-headers, linux-libc-dev, linux-image.

examples/tests/linuxbuild.py
tambewilliam: Thank you,
I am just noticing the fix.

Would it be ok to keep the url functionality of this pull-request for cases where we need to pull the kernel to build from a different location ?

avocado/utils/kernel.py
clebergnu: There's nothing *wrong* with your proposal for a `build_for_install` parameter to `build()`, but when designing APIs, it makes sense to brainstorm a bit more than usual, so, let's do it.

Initially, based on my previous comment (https://github.com/avocado-framework/avocado/pull/2527#discussion_r175624169), I thought it made more sense to have a generic (unchanged) `build()` method, and the `install()` method would handle the last build steps too (assuming most of the build would already be done, and the packaging would take little time and could be absorbed by it).

But, now I think your idea is better.  The target to be built, be it a stock "vmlinuz" image or a package, is really *build* related, and not installation related.  So, having said that, I'd only tweak the type of the parameter (bool) and optionally the name (suggestion: `binary_package`).

avocado/utils/kernel.py
clebergnu: Sure, my point is that `process.run()` has parameters for both `shell` and `sudo`, so you can use something like:

```
process.run('dpkg -i %s/*.deb' % self.workdir, shell=True, sudo=True)
```

examples/tests/linuxbuild.py
clebergnu: You mean adding a `url` parameter to `kernel.KernelBuild.donwload()` API? That's a nice improvement.  Just account for the two situations (passing a URL and using the default one).

avocado/utils/kernel.py
tambewilliam: Never knew! thank you; making the change.

avocado/utils/kernel.py
tambewilliam: Please let me know if this change would work:

```diff
@@ -99,17 +99,18 @@ class KernelBuild(object):
             dotconfig = os.path.join(self.linux_dir, '.config')
             shutil.copy(self.config_path, dotconfig)

-    def build(self, build_for_install=None):
+    def build(self, binary_package=None):
         """
         Build kernel from source.

-        :param build_for_install: when not None, the appropriate
+        :param binary_package: when not None, the appropriate
                                   platform package is built
                                   for install() to use
+        :type binary_pacakge: bool or None
         """
         log.info("Starting build the kernel")
         build_output_format = ""
-        if build_for_install is not None:
+        if binary_package is not None:
             if self.distro.name == "Ubuntu":
                 build_output_format = "deb-pkg"
         if self.config_path is None:
```

avocado/utils/kernel.py
clebergnu: This:

http://avocado-framework.readthedocs.io/en/59.0/api/utils/avocado.utils.html

Should be handy to have in a bookmark.

avocado/utils/kernel.py
clebergnu: What I meant by using a bool is that `binary_package=False` can be the default. The parameter would simply be:

```
:type binary_pacakge: bool
```

No need for None here.

avocado/utils/kernel.py
tambewilliam: Thank you; bookmarked !

avocado/utils/kernel.py
tambewilliam: did I get it right?

```diff
diff --git a/avocado/utils/kernel.py b/avocado/utils/kernel.py
index ee566f52..9e7db63c 100644
--- a/avocado/utils/kernel.py
+++ b/avocado/utils/kernel.py
@@ -99,18 +99,18 @@ class KernelBuild(object):
             dotconfig = os.path.join(self.linux_dir, '.config')
             shutil.copy(self.config_path, dotconfig)

-    def build(self, binary_package=None):
+    def build(self, binary_package=False):
         """
         Build kernel from source.

-        :param binary_package: when not None, the appropriate
+        :param binary_package: when True, the appropriate
                                   platform package is built
                                   for install() to use
-        :type binary_pacakge: bool or None
+        :type binary_pacakge: bool
         """
         log.info("Starting build the kernel")
         build_output_format = ""
-        if binary_package is not None:
+        if binary_package is True:
             if self.distro.name == "Ubuntu":
                 build_output_format = "deb-pkg"
         if self.config_path is None:
```

clebergnu:
@tambewilliam given the amount of things we discussed, please send a v2, and I'll review it thoroughly. BTW:

http://avocado-framework.readthedocs.io/en/59.0/ContributionGuide.html#contributing-to-avocado

Describes how to deal with newer versions of a given PR.

tambewilliam:
@clebergnu sending v2.
Also looking into the link to deal with newer versions of a given PR.

tambewilliam:
The latest Travis CI check succeed for Python2.7, but fail for Python3.4 .

from the logs, I am not able to see that the failure is due to a file I changed; is there something I am overlooking in the logs ?

clebergnu:
@tambewilliam yes, there has been issues (not always reproducible) with Python 3.4 jobs. We're tracking the issue here:

https://trello.com/c/Axy3t4s0/1259-functional-tests-on-python-3-leave-directories-behind

For now, please either attempt to reschedule the Python 3.4 job of the build.  If that doesn't work, we'll test it on our side.

clebergnu:
@tambewilliam btw, if you're about to send a v2 (which should be a different pull request), please close this one.

tambewilliam:
@clebergnu Thank you; new pull-request at:
https://github.com/avocado-framework/avocado/pull/2542

PR #2529, opened 2018-03-20
python 3 :fix avocado-vt bootstrap issue due to process.run return by…

chunfuwen:
…tes instead of

string in python3

### DIFF
--- a/avocado/utils/memory.py
+++ b/avocado/utils/memory.py
@@ -115,7 +115,8 @@ def read_from_meminfo(key):
     :param key: Key name, such as ``MemTotal``.
     """
     cmd_result = process.run('grep %s /proc/meminfo' % key, verbose=False)
-    meminfo = cmd_result.stdout
+    # Get text mode result for compatibility.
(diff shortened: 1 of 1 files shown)


avocado/utils/memory.py
```diff
@@ -115,7 +115,7 @@ def read_from_meminfo(key):
     :param key: Key name, such as ``MemTotal``.
     """
     cmd_result = process.run('grep %s /proc/meminfo' % key, verbose=False)
-    meminfo = cmd_result.stdout
+    meminfo = cmd_result.stdout_text
```
hsmj1412: It is a function 
   def stdout_text(self):
        if type(self.stdout) in string_types:
            return self.stdout
        if hasattr(self.stdout, 'decode'):
            return self.stdout.decode(self.encoding)
        raise TypeError("Unable to decode stdout into a string-like type")

chunfuwen: updated

apahim:
@chunfuwen could you please rebase?

PR #2563, opened 2018-04-04
tap: replace the text_output broken call

apahim:
text_output was removed from test status. Now we have to open the log
file directly.


apahim:
[BUG: non-existing `text_output` field is still used by TAP and runner](https://trello.com/c/21EO3Uda/1286-bug-non-existing-textoutput-field-is-still-used-by-tap-and-runner)

PR #2579, opened 2018-04-13
Wait 10s for the memory state to change

abdhaleegit:
Waiting 120s is not appropriate as the system takes less time for
memory state to change from 'going-online' to 'online' so changed
it to 10s and with freq of 0.2s


abdhaleegit:
This is with reference to #2216

PR #2612, opened 2018-05-02
python3: fix 'str' does not support the buffer interface

chunfuwen:
when writing log into log file,string is used instead of bytes

### DIFF
--- a/avocado/core/sysinfo.py
+++ b/avocado/core/sysinfo.py
@@ -327,8 +327,8 @@ class LogWatcher(Collectible):
             self.inode = current_inode
             self.size = current_size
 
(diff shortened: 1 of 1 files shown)


chunfuwen:
paste error information here:
```
DATA (filename=output.expected) => NOT FOUND (data sources: variant, test, file)
DATA (filename=stdout.expected) => NOT FOUND (data sources: variant, test, file)
DATA (filename=stderr.expected) => NOT FOUND (data sources: variant, test, file)
Log file /var/log/messages collection failed: 'str' does not support the buffer interface
```

chunfuwen:
@ldoktor 
@clebergnu 
please help take a look.

PR #2683, opened 2018-06-21
multipath: Replaced the sleep with wait_for

narasimhan-v:
We were sleeping instead of waiting for the service to become
stable, after a service restart. So, replaced the sleep with
wait_for.


narasimhan-v:
Similar fix has gone into the test as well: https://github.com/avocado-framework-tests/avocado-misc-tests/pull/928

narasimhan-v:
@ldoktor Thanks

PR #2705, opened 2018-07-11
utils/archive: return extracted name

balamuruhans:
tests or framework might need the name of the extracted archive to
perform necessary steps, it is better and harmless to return the name
in extract().

### DIFF
--- a/avocado/utils/archive.py
+++ b/avocado/utils/archive.py
@@ -168,10 +168,14 @@ class ArchiveFile(object):
         Extract all files from the archive.
 
         :param path: destination path.
+        :return: extracted folder name
         """
         self._engine.extractall(path)
         if self.is_zip:
             self._update_zip_extra_attrs(path)
+            return self._engine.namelist()[0].strip(os.sep)
+
+        return self._engine.getnames()[0]
 
     def _update_zip_extra_attrs(self, dst_dir):
         if platform.system() != "Linux":
@@ -249,7 +253,7 @@ def uncompress(filename, path):
     :param path: destination path to extract to.
     """
     with ArchiveFile.open(filename) as x:
-        x.extract(path)
+        return x.extract(path)
 
 
 # Some aliases


balamuruhans:
@ldoktor Thank you for reviewing the PR.

I use it in Avocado-vt for retrieving the folder name of extracted archive, where any stress tools like (stress-ng, stress, io-zone or any existing / future tools) are downloaded and extracted in the host. Extracted folder should be copied to VMs or remote host to run stress and working directory if stress to be run on local host. 

As you have said, users can specify the directory where they want to archive to be extracted by giving `path` which we generally use the avocado's tmp directory where we might have other files like guest xml, device xml etc., as framework takes care of creating and cleaning the tmp directories.

Also the extracted directory name might not be always same as the tar/zip compressed file name. so IMHO getting the extracted name from the `tarfile` object would be the right thing to do.

ldoktor:
Well my problem is you're not returning the extracted name, you're returning first part of extracted path. So if your tar consists of:

```
README.rst
netperf/netperf.exe
```
it might return either `README.rst` or `netperf` depending on which one is the first in filelist. For this reason I don't like this much.

balamuruhans:
@ldoktor you are right! I have fixed the issue, please help to review.

ldoktor:
Well that fixes the tar/zip difference, but how does this fixes the:

```
README.rst
netperf/netperf.exe
```

issue? I mean depending on the order the files were archived you can get either `README.rst` or `netperf` returned from the `archive.decompress` which could lead to hard to debug issues. I'd suggest using `Archive.list` to get full list of items or simply `os.listdir` on the directory that was provided. 

Anyway let's get another opinion @clebergnu, what is your take on this one?

balamuruhans:
@ldoktor generally a folder (eg: `netperf/` which contain `README.rst`, `netperf.exe`, etc.,) would be archived as netperf.tar.gz or netperf.zip. The extract API would return the folder name (netperf) which is archived.
The folder name might be different than the archive name, so the current fix would return correct folder name which is archived (irrespective of any archived name)

ldoktor:
It's merged, but let's discuss this a bit further. It might simplify some workflows, but it might as well bring unpredictable behavior.
[archives.zip](https://github.com/avocado-framework/avocado/files/2222274/archives.zip)

First the `empty.tar.bz2` crashes (fix is simple and if we decided not to revert this I can fix that). Then there are 2 packages containing the same files, but the return value of `extract` is completely different and that is what I complained about earlier.

You can say it's fairly safe, but imagine you're downloading `netperf`. You use AssetFetcher and in 99% of the time it's downloaded from `github`, therefor `netperf-master` folder. But in 1% of the cases it could be downloaded from source-forge (not sure whether Netperf is there, but other projects are) and source-forge used to inject README file in main directory (not sure whether it still does that, but other providers can do that). Now looking at the test results, in 99% it'll work well but in 1% of the time it'll say `Not a directory`. And that is what I complained about.

Better solution for the `netperf` sources would be `os.path.join(path, os.walk(path).next()[1][0])`, which at least makes sure it returns a directory, or (what I'd probably use) `glob.glob(os.path.join(path, 'netperf-*'))`. Adding this line to every test-file that expects a directory in archive is IMO cleaner, than always returning first thing.

ldoktor:
Actually it's even more dangerous than I though. Have a look at following 
[netperf.zip](https://github.com/avocado-framework/avocado/files/2222368/netperf.zip)

netperf archive. I had to compress it into "zip" because github does not allows to attach `tar.bz2` but when you extract it and then use `archive.extract('netperf.tar.bz2', '.')` it returns `netperf/src` and not `netperf`.

Do you still believe this is a simplification, or potential box full of gun power?

vivianQizhu:
@ldoktor Even it might not fit for all scenarios, this patch does extend the usage of `archive` module. Previously I would use like build-in tarfile instead of this module to get the directory by myself. From user's perspective, I would accept it returns correct path for most case, and I guess user can handle the exceptions by themselves if it is not easy to handle all possible cases inside `archive` module. But for sure it should not crash.
So, first I can accept wrong path, and second I would request fix for the crash case, instead of revert it.

clebergnu:
I have the same opinion, it used to return nothing, so the first member of
the archive is a freebie. Users can always do a full listing of the
contents later.

ldoktor:
OK, but once it starts misbehave remember my words...

clebergnu:
@ldoktor I have to say I have a short memory, but GitHub is here to prove your point! :wink:

PR #2760, opened 2018-08-13
Adds egg-info directories to gitignore

cacarrara:
When `make develop` is run egg-info directories are created. We
shouldn't add them to version control, so this commit adds the proper
entry to gitignore file.


cacarrara:
I integrate my shell with git so the shell sign pointing that there's modifications not commited kinda annoy me :smile:

PR #2778, opened 2018-08-17
Fix profiler1

PraveenPenguin:
Fixed avocado:core:sysinfo
1-fixed pep8 issue
2- Add __init__ constructor that can be hold instance variable
3-
a--rename pipe variable to meaningful name
b- Above variable moved as in __init as it should
   instance variable
4-avocado crashed if user provide non exsting profiler
this patch address that , if profiler not avilable than
framework  handle gracefully handle


avocado/core/sysinfo.py
```diff
@@ -202,21 +206,28 @@ def run(self, logdir):
...
         stdout = open(logf_path, "w")
-        self.pipe = subprocess.Popen(shlex.split(self.cmd), stdin=stdin, stdout=stdout,
-                                     stderr=subprocess.STDOUT, shell=False, env=env)
+        try:
+            self.daemon_process = subprocess.Popen(shlex.split(self.cmd),
+                                                   stdin=stdin, stdout=stdout,
+                                                   stderr=subprocess.STDOUT,
+                                                   shell=False, env=env)
+        except OSError:
+            log.debug("Not logging  %s (command not found) " % self.cmd)
```
cacarrara: Since the exception being caught here is not explicitly the `FileNotFoundError`, what do you think to update the log message replacing `(command not found)` by `(command could not be run)`?

PraveenPenguin: @cacarrara yes make sense will adopt , thanks for review

PraveenPenguin:
@clebergnu can you please help me why travis is failing , i am kind of stuck here :(

avocado/core/sysinfo.py
```diff
@@ -205,24 +205,28 @@ def run(self, logdir):
...
-        retcode = self.daemon_process.poll()
-        if retcode is None:
-            process.kill_process_tree(self.daemon_process.pid)
-            retcode = self.daemon_process.wait()
-        else:
-            log.error("Daemon process '%s' (pid %d) "
-                      "terminated abnormally (code %d)",
-                      self.cmd, self.daemon_process.pid, retcode)
-        return retcode
+        if self.daemon_process:
```
apahim: `if self.daemon_process is None: `

avocado/core/sysinfo.py
```diff
@@ -186,6 +186,8 @@ class Daemon(Command):
     :param logf: Basename of the file where output is logged (optional).
     :param compress_logf: Whether to compress the output of the command.
     """
+    def __init__(self, cmd):
```
apahim: Now you're limiting the arguments and potentially breaking the API. The parent class can also take `logf` and `compress_log`, so you have to accept them here as well. By the way, both classes docstrings have a typo (`compress_logf`).

PraveenPenguin:
@apahim thanks for review  push the changes please have a look

apahim:
@PraveenPenguin you forgot to fix the docstrings. There, we read `:param compress_logf:`, when it should actually be `:param compress_log:`. That happens on both `Daemon` and `Command` classes.

avocado/core/sysinfo.py
```diff
@@ -202,21 +207,29 @@ def run(self, logdir):
...
-        retcode = self.pipe.poll()
-        if retcode is None:
-            process.kill_process_tree(self.pipe.pid)
-            retcode = self.pipe.wait()
-        else:
-            log.error("Daemon process '%s' (pid %d) terminated abnormally (code %d)",
-                      self.cmd, self.pipe.pid, retcode)
-        return retcode
+
+        if self.daemon_process is None:
```
cacarrara: Here the check must be `if self.daemon_process is not None`. You need a daemon_process to stop. It's weird this didn't broke any test :thinking: .

avocado/core/sysinfo.py
```diff
@@ -187,6 +187,11 @@ class Daemon(Command):
     :param compress_logf: Whether to compress the output of the command.
     """
 
+    def __init__(self, cmd, logf=None, compress_log=False):
```
cacarrara: @PraveenPenguin Since the parent class constructor parameters aren't relevant in this class, I would suggest to keep API this way:
```python
def __init__(self, *args, **kwargs):
    super().__init__(*args, **kwargs)
```
So if the parent class adds new constructor arguments there's no need to update anything here and the API is still valid. 

What do you think @apahim?

apahim: No strong opinion. Particularly I'd prefer the `*args`/`**kwargs`, even though that would make this class the only one in this module doing that.

avocado/core/sysinfo.py
apahim: Yes, that's what I actually meant... :)

PraveenPenguin: @cacarrara ah sure i miss that :)

avocado/core/sysinfo.py
PraveenPenguin: @cacarrara ah good point i will adopt this

cacarrara:
@PraveenPenguin thanks for the contributions. We're facing this false-positive tests failure om CI currently, but we're also already working to fix that. I can restart the build for this PR.

Meanwhile do you mind to consolidate all your changes in a new PR? It's explained here in the docs the contribution workflow: https://avocado-framework.readthedocs.io/en/63.0/ContributionGuide.html#git-workflow

PraveenPenguin:
@cacarrara @clebergnu @apahim any thing needed in this PR

clebergnu:
If you go through the commits you'll find some opportunities for squashing
commits together, cleaning up the commit message and providing an updated
and consolidated description of the changes.

PraveenPenguin:
@clebergnu pushed the changes now travis is failing , can you please have a look , if any thing needed let me know

clebergnu:
@PraveenPenguin attempting to restart the Travis job.

PR #2823, opened 2018-08-29
Enables pylint W0705

cacarrara:
This check makes sure the excepts in try/except are not duplicated
regarding the exception type.


avocado/utils/partition.py
```diff
@@ -258,7 +258,7 @@ def _get_pids_on_mountpoint(self, mnt):
             msg = 'Could not find lsof executable'
             LOG.error(msg)
             raise PartitionError(self, msg, details)
-        except OSError as details:
+        except OSError as details:  # pylint: disable=W0705
```
apahim: That's quite a complex logic for an error handler... all that to show different messages on different python versions for the same problem. And `IOError` is an uninvited guest. Not your patch's fault, but we can take advantage of this pylint warning to make it better.

For python 2, an `OSError` will be raised on subprocess "No such file or directory". On python 3, a `FileNotFoundError` will be raised instead. But both of them can be caught with an `except OSError`, since `FileNotFoundError` inherits from it.

So, this entire function should be something like this:
```python
def _get_pids_on_mountpoint(self, mnt):
    try:
        ...
    except OSError:
        ...
        raise PartitionError
    except process.CmdError:
        ...
        raise PartitionError
```

optional_plugins/html/avocado_result_html/__init__.py
```diff
@@ -98,10 +98,8 @@ def _get_sysinfo(self, sysinfo_file):
         try:
             with open(sysinfo_path, 'r') as sysinfo_file:
                 sysinfo_contents = sysinfo_file.read()
-        except OSError as details:
+        except (OSError, IOError) as details:
```
apahim: ok

avocado/utils/partition.py
cacarrara: Sure. I spent some time trying to simplify it, but I couldn't refactor the current behavior that identify when `lsof` is not installed on system and when any other error happen. I thought make what you suggest but this way we're not just refactoring it but changing it.

I'm assuming the `OSError` raised by Python `subprocess.run` is not always because some command was not found. The documentation[1] actually doesn't make it very clear, though.

It would be nice hear some @clebergnu thoughts as well.

[1] - https://docs.python.org/3/library/subprocess.html#exceptions

apahim: I see only 2 clear error situations: `lsof` not present and `lsof` failed.

Python 2:
```python
>>> process.run('foo')
OSError: [Errno 2] No such file or directory (foo)

>>> process.run('lsof foo')
avocado.utils.process.CmdError
```

Python 3:
```python
>>> process.run('foo')
FileNotFoundError: [Errno 2] No such file or directory: 'foo' (foo): 'foo'

>>> process.run('lsof foo')
avocado.utils.process.CmdError: ('lsof foo', <avocado.utils.process.CmdResult object at 0x7f6184751e10>)
```

The `No such file or directory` can be caught by `except OSError` on both python2 and python3. The errors caused by the `lsof` execution are caught by `except CmdError`. If we know about something else that can raise an `OSError`, then we should be parsing the exception message. Or just forwarding the `OSError` message to the `PartitionError` and let user deal with that.

PR #2826, opened 2018-09-03
Enables pylint w0223

cacarrara:
This change makes de proper implementations to enable pylint
W0223 checks. It was splitted in two commits. More details can be found
on them descriptions.


cacarrara:
Ok, doing...

cacarrara:
Closing in favor of #2832

PR #2829, opened 2018-09-04
Enables pylint w0631

cacarrara:
This change enables pylint check w0631. Since the only occurrence is on
external spark library, it was disabled specifically.


cacarrara:
Closing in favor of #2830

PR #2831, opened 2018-09-06
Enables pylint W0705 v2

cacarrara:
This check makes sure the excepts in try/except are not duplicated
regarding the exception type.

---
Change from v1 (#2823)
* Makes `utils.partition.Partition._get_pids_on_mountpoint()` 
handle `FileNotFoundError` as `OSError` only.


avocado/utils/partition.py
```diff
@@ -254,10 +254,6 @@ def _get_pids_on_mountpoint(self, mnt):
             cmd = "lsof " + mnt
```
apahim: You forgot to remove the try/except for `FileNotFoundError` above.

PR #2836, opened 2018-09-07
enhance avocado:utils:genio as a new utility

PraveenPenguin:
added utility which takes as argument a file and pattern
file or not

>>> genio.grep_pattern("/home/praveen/s","pattern")
  File "avocado/utils/genio.py", line 222, in grep_pattern
    % (filename, command))
avocado.utils.genio.GenIOError: invalid file /home/praveen/s to cat to command grep "pattern" > /dev/null
>>> genio.grep_pattern("/home/praveen/file1","advik")
True
>>> genio.grep_pattern("/home/praveen/file1","ss")
False
>>>

### DIFF
--- a/avocado/utils/genio.py
+++ b/avocado/utils/genio.py
@@ -201,3 +201,23 @@ def write_file_or_fail(filename, data):
     except OSError as details:
         raise GenIOError("The write to %s failed: %s" % (
                          filename, details))
+
+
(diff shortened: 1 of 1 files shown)


PraveenPenguin:
@clebergnu thanks for review

PraveenPenguin:
closing in favour of #2837

PR #2838, opened 2018-09-10
Enables pylint w0702

cacarrara:
**Important:**
Beyond the enablement of W0702 overall this change also updates 
the `scripts.avocado.handle_exception` function so it doesn't need to
use a bare-except anymore. The code except responsible for get the crash
directory was kept in a separeted function. This specific code snippet
can't raise any exception. Internally it uses the
`avocado.utils.path.usable_rw_dir` that already deals with possible
exceptions.

Another possible point of exception could be the import of
avocado.core.data_dir module before avocado be installed itself. However
this import was moved to the initial block of import and dealed proper
as ImportError.

This way seems all the code insisde try block keep being processed by
try-except blocks without bare-except like before.


cacarrara:
Seems that some problem doesn't let Travis to post status in this PR. But the build started: https://travis-ci.org/avocado-framework/avocado/builds/426837307

PR #2841, opened 2018-09-14
changes in avocado:utils:cpu utility

PraveenPenguin:
1-Fixed style issue in avocado:utils:cpu utility
fixed pep8 issue in multiple places
2-enhance avocado:utils:cpu as a new utility 
added utility which return a list include all cpus the process
used

after this patch

>>> cpu.get_pid_cpu("22207")
['1']
>>> cpu.get_pid_cpu("")
[]
>>>

### DIFF
--- a/avocado/utils/cpu.py
+++ b/avocado/utils/cpu.py
@@ -212,14 +212,17 @@ def get_cpuidle_state():
     :rtype: Dict of dicts
     """
     cpus_list = cpu_online_list()
-    states = range(len(glob.glob("/sys/devices/system/cpu/cpu0/cpuidle/state*")))
+    states = range(
+        len(glob.glob("/sys/devices/system/cpu/cpu0/cpuidle/state*")))
     cpu_idlestate = {}
     for cpu in cpus_list:
         cpu_idlestate[cpu] = {}
         for state_no in states:
-            state_file = "/sys/devices/system/cpu/cpu%s/cpuidle/state%s/disable" % (cpu, state_no)
+            state_file = "/sys/devices/system/cpu/cpu%s/cpuidle/state%s/disable" % (
+                cpu, state_no)
             try:
-                cpu_idlestate[cpu][state_no] = int(open(state_file, 'rb').read())
+                cpu_idlestate[cpu][state_no] = int(
+                    open(state_file, 'rb').read())
             except IOError as err:
                 logging.warning("Failed to read idle state on cpu %s "
                                 "for state %s:\n%s", cpu, state_no, err)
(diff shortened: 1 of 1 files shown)


avocado/utils/cpu.py
```diff
@@ -212,14 +214,17 @@ def get_cpuidle_state():
     :rtype: Dict of dicts
     """
     cpus_list = cpu_online_list()
-    states = range(len(glob.glob("/sys/devices/system/cpu/cpu0/cpuidle/state*")))
+    states = range(
```
cacarrara: I think all these indentation changes should not be in this PR. Keep only the new function.

avocado/utils/cpu.py
```diff
@@ -340,3 +348,24 @@ def get_cpufreq_governor():
...
+
+def get_pid_cpu(pid):
+    """
+    Get the process used cpus.
+    :param pid: process id
+    :type pid: string
+    :return: A list include all cpus the process used
+    :rtype: list
+    """
+    cmd = "ps -o cpuid -L -p %s" % pid
```
cacarrara: We're trying to minimize use of external commands to grab informations which are in standard files[1]. Seems this is also an use case where we can get the information com `proc/[pid]/stat` file instead of to depend on external command. Looking to the kernel doc[2], there's information about `which CPU the task is scheduled on`. Is it what you need?

[1] - https://github.com/avocado-framework/avocado/pull/2851
[2] - https://www.kernel.org/doc/Documentation/filesystems/proc.txt (`Table 1-4: Contents of the stat files (as of 2.6.30-rc7)`)

avocado/utils/cpu.py
PraveenPenguin: @cacarrara ok sure  but this PR has two commits  one for pep8 changes that for indentation changes and other is for new function , is that not ok ? will raise in two PR

avocado/utils/cpu.py
PraveenPenguin: ah sure let me check

avocado/utils/cpu.py
cacarrara: It's weird, because as far as I could see, our lint check is not failing for these lines. So, seems they're correct taking our current validation rules.

PraveenPenguin: @cacarrara yes i know i just run pep8 and give me warning  about these  so thought we can make pep8 complaint , if you think it not needed will re spin the patch  having only new added functionality please suggest

cacarrara: I would prefer keep the rules of our current `make check` (that run inspektor lint). We can discuss improvements on our lint and apply them wide in project after.

PraveenPenguin: @cacarrara sure make sense

cacarrara:
@PraveenPenguin are you going to send the updates here?

cacarrara:
Since this is a valuable contribution but hadn't updates for some time, I'm closing it but continuing with the changes and request adjustments in a new PR (#2989) giving the proper credit for the initial contributor. Thanks Praveen.

PR #2866, opened 2018-09-25
Python 3: fix a byte-str issue of utils.service

sitoliu:
Fix issue: re.search(r"Loaded: loaded", output) failed
| TypeError: cannot use a string pattern on a bytes-like object

ID: 1632665
Signed-off-by: Sitong Liu <siliu@redhat.com>


sitoliu:
@clebergnu Thanks, updated.

cacarrara:
@clebergnu could you see this again?

sitoliu:
Thanks @clebergnu @cacarrara , updated commit info to make it more clear.

PR #2876, opened 2018-09-27
linux_modules: eliminate duplicates in submodules

lolyu:
The result in `get_submodules` contains duplicates, which result
from no check of visited in DFS.
Solved by using `ordered_list_unique` from `data_structures` to
del duplicates before returning from each recursion.


lolyu:
@clebergnu thanks.

PR #2879, opened 2018-09-27
vmimage: more tests for parser and base and minor improvements

cacarrara:
This change focus mostly on increasing tests for `VmImageHtmlParser` and
`ImageProviderBase`. During the tests implementation some improvements
were detected: the process of feed a html parser run both for get an
image version and image url was duplicated so it was unified in a single
method.

Besides that the process to get a best version for an image was put on a
specific method so each provider can override it if required. Today the
default rule for a best version is based on calling max() function on
found versions. However we can't assume it's the default for every
provider. As far as we know, for a possible OpenSuse provider in future
this rule will not be valid.

Reference: https://trello.com/c/vtM1SPpJ

### DIFF
--- a/selftests/unit/test_utils_vmimage.py
+++ b/selftests/unit/test_utils_vmimage.py
@@ -1,9 +1,15 @@
 import unittest
+try:
+    from unittest import mock
+except ImportError:
+    import mock
+
+from six.moves.urllib.error import HTTPError
 
 from avocado.utils import vmimage
 
 
-class VMImage(unittest.TestCase):
+class VMImageTest(unittest.TestCase):
 
     def test_list_providers(self):
         self.assertIsNotNone(vmimage.list_providers())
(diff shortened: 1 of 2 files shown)


selftests/unit/test_utils_vmimage.py
```diff
@@ -1,9 +1,15 @@
...
+except ImportError:
+    import mock
+
+from six.moves.urllib.error import HTTPError
 
 from avocado.utils import vmimage
 
 
-class VMImage(unittest.TestCase):
+class VMImageTest(unittest.TestCase):
```
clebergnu: This is also a matter of style: I've been favoring not putting the `Test` suffix, because:
 * It already lives in the module namespace `selftests.unit.<module>`
 * It inherits from `unittest.TestCase`
So I think it's too verbose.  Feel free to give your opinion... in the end we it'd be nice to have a consistent style across all tests.

selftests/unit/test_utils_vmimage.py
```diff
@@ -13,5 +19,101 @@ def test_concrete_providers_have_name(self):
...
+        html = ('<html><head><title>Test</title></head>'
+                '<body>%s</body></html>' % ''.join(links))
+        self.parser.feed(html)
+        self.assertTrue(self.parser.items)
+        self.assertEqual(len(self.parser.items), len(versions))
+        self.assertTrue(all(version in self.parser.items for version in versions))
+
+
+class ImageProviderBaseTest(unittest.TestCase):
+    def _get_html_with_versions(self, versions):
```
clebergnu: This can be a staticmethod.

selftests/unit/test_utils_vmimage.py
```diff
@@ -13,5 +19,101 @@ def test_concrete_providers_have_name(self):
...
+        html = '<html><head><title>Test</title></head><body>%s</body></html>'
+        return html % ''.join(['<a href="%s/" />' % v for v in versions])
+
+    @mock.patch('avocado.utils.vmimage.urlopen')
+    def test_get_version(self, urlopen_mock):
+        html_fixture = self._get_html_with_versions([10, 11, 12])
+        urlread_mocked = mock.Mock(return_value=html_fixture)
+        urlopen_mock.return_value = mock.Mock(read=urlread_mocked)
+        base_image = vmimage.ImageProviderBase(version='[0-9]+', build=None, arch=None)
+
```
clebergnu: Style wise, *I* would not add a blank line here.

selftests/unit/test_utils_vmimage.py
cacarrara: Sure, I can change it. What's the advantage to have this method as static? I know it doesn't use the instance (self) for anything, but beyond that is there any real advantage having static method (when it could be) over a common method?

clebergnu:
Yep, I can look for some pointers on the advantages, but out of the top of
my head I can think of reduced memory usage because of one less argument,
the static method can and will be shared across instances of the same
class, but, above all IMO is readability and keeping the goal of the method
more "honest".

selftests/unit/test_utils_vmimage.py
cacarrara: Cool. It's a good point. I also think putting "Test" suffix keep things too verbose.

PR #2899, opened 2018-10-10
utils.service: enforce uniform string type

lolyu:
for result parser functions, enforce all strings to be of bytes.

### DIFF
--- a/avocado/utils/service.py
+++ b/avocado/utils/service.py
@@ -153,7 +153,7 @@ def sys_v_init_result_parser(command):
             status_per_target_dict = {}
             # Dict to store the status for service based on xinetd.
             xinet_services_dict = {}
-            lines = cmd_result.stdout.strip().splitlines()
+            lines = cmd_result.stdout_text.strip().splitlines()
             for line in lines:
                 sublines = line.strip().split()
                 if len(sublines) == 8:
@@ -231,7 +231,7 @@ def systemd_result_parser(command):
                 raise process.CmdError(cmd_result.command, cmd_result)
             # Dict to store service name to status.
             _service2status_dict = {}
-            lines = cmd_result.stdout.strip().splitlines()
+            lines = cmd_result.stdout_text.strip().splitlines()
             for line in lines:
                 sublines = line.strip().split()
                 if ((not len(sublines) == 2) or
--- a/selftests/unit/test_utils_service.py
+++ b/selftests/unit/test_utils_service.py
@@ -190,7 +190,7 @@ class TestSystemdServiceManager(TestServiceManager):
 
     def test_list(self):
         list_result_mock = mock.Mock(exit_status=0,
-                                     stdout="sshd.service enabled\n"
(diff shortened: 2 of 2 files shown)


lolyu:
There seems some failures of unit tests in CI due to that unit tests use `str` in python3. So there are two options actually to fix this issue.
1. use `stdout_text` instead `stdout`.
2. enforce all strings to be of `bytes`
there was a previous patch choosing second approach to enforce `bytes`, but there is a further need to update the unit tests scripts. I think this needs to be discussed.

cacarrara:
@lolyu are you checking the CI failures?

lolyu:
@cacarrara updated, thanks.

avocado/utils/service.py
```diff
@@ -161,22 +161,22 @@ def method(cmd_result):
                     service_name = sublines[0]
                     # Store the status of each target in status_per_target_dict
                     for target in range(7):
-                        status = sublines[target + 1].split(":")[-1]
+                        status = sublines[target + 1].split(b":")[-1]
```
cacarrara: @lolyu I think we can use only strings instead of bytes here. You can change line 156 above to get `stdout_text`[1] instead of `stdout` from `cmd_result`. So we can keep using only string objects itself.

[1] - https://github.com/avocado-framework/avocado/blob/master/avocado/utils/process.py#L365

avocado/utils/service.py
```diff
@@ -161,22 +161,22 @@ def method(cmd_result):
...
+                    service_name = sublines[0].strip(b":")
                     status = sublines[-1]
                     xinet_services_dict[service_name] = status
 
                 else:
                     # Header or some lines useless.
                     continue
             # Add xinetd based service in the main dict.
-            services_statuses_dict["xinetd"] = xinet_services_dict
+            services_statuses_dict[b"xinetd"] = xinet_services_dict
```
cacarrara: I think we can keep string to dict keys here.

avocado/utils/service.py
```diff
@@ -235,10 +235,10 @@ def method(cmd_result):
             for line in lines:
                 sublines = line.strip().split()
                 if ((not len(sublines) == 2) or
-                        (not sublines[0].endswith("service"))):
+                        (not sublines[0].endswith(b"service"))):
```
cacarrara: Same comment above about `stdout_text`.

selftests/unit/test_utils_service.py
```diff
@@ -190,9 +190,9 @@ def test_start(self):
 
     def test_list(self):
         list_result_mock = mock.Mock(exit_status=0,
-                                     stdout="sshd.service enabled\n"
-                                     "vsftpd.service disabled\n"
-                                     "systemd-sysctl.service static\n")
+                                     stdout=b"sshd.service enabled\n"
```
cacarrara: If you change the use to `stdout_text` as mentioned before I think all this tests can be keeped untouched.

lolyu:
@cacarrara updated, thanks.

PR #2901, opened 2018-10-13
utils.partition: rewrite with filelock

lolyu:
1. rewrite MtabLock, as a proxy class for filelock.FileLock.
2. update related unit test.

### DIFF
--- a/avocado/utils/partition.py
+++ b/avocado/utils/partition.py
@@ -24,11 +24,11 @@ Utility for handling partitions.
 
 import logging
 import os
-import time
-
-import fcntl
+import hashlib
+import tempfile
 
 from . import process
+from . import filelock
 
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


avocado/utils/partition.py
```diff
@@ -50,30 +50,29 @@ def __str__(self):
...
-                if details.errno == 11:
-                    time.sleep(0.1)
-                else:
-                    raise
+        try:
+            super(MtabLock, self).__enter__()
+        except (filelock.LockFailed, filelock.AlreadyLocked) as e:
+            reason = "Unable to obtain '/etc/mtab' lock in %ds" % self.timeout
+            raise PartitionError(self, reason, e)
         else:
```
cacarrara: Probably you can drop this `else` without problems.

avocado/utils/partition.py
```diff
@@ -50,30 +50,29 @@ def __str__(self):
...
-                break
-            except IOError as details:
-                if details.errno == 11:
-                    time.sleep(0.1)
-                else:
-                    raise
+        try:
+            super(MtabLock, self).__enter__()
+        except (filelock.LockFailed, filelock.AlreadyLocked) as e:
+            reason = "Unable to obtain '/etc/mtab' lock in %ds" % self.timeout
```
cacarrara: Probaby you should also get the device from the new created attribute `self.device` here for the message. For consistency.

Besides that, I think you forgot to put the `timeout` init parameter in a instance attribute (self). As none test got this, it could be good see if it's not necessary to write a proper test.

avocado/utils/partition.py
```diff
@@ -50,30 +50,29 @@ def __str__(self):
                                       super(PartitionError, self).__str__())
 
 
-class MtabLock(object):
+class MtabLock(filelock.FileLock):
```
cacarrara: Well, at least for me seems that we should be using composition over inheritance here[1]. On my understanding `MtabLock` uses a `FileLock` and do not *is a* `FileLock`. So it's a composition relation. A smell about that is the changing API of the initialization method. On `MtabLock` the file name argument was dropped.

I do not think it is going to be too different from what you've proposed. I've tried a quick modification from your PR like:

```python
class MtabLock(object):
    def __init__(self, timeout=60):
        self.timeout = timeout
        self.mtab = None
        self.device = 'etc/mtab'
        self.lock_filename = os.path.join(tempfile.gettempdir(), 'some_name')
        self.file_lock = filelock.FileLock(self.lock_filename, timeout=self.timeout)

    def __enter__(self):
        try:
            self.file_lock.__enter__()
        except (filelock.LockFailed, filelock.AlreadyLocked) as e:
            reason = "Unable to obtain '/etc/mtab' lock in %ds" % self.timeout
            raise PartitionError(self, reason, e)
        self.mtab = open(self.device)
        return self

    def __exit__(self, exc_type, exc_value, exc_traceback):
        if self.mtab:
            self.mtab.close()
        self.file_lock.__exit__()
```

[1] - https://en.wikipedia.org/wiki/Composition_over_inheritance

lolyu: IMHO, as the demos wiki provides, composition over inheritance is better in scenario like using inheritance as composition(as the demo wiki provides: to use multiple inheritance to implement class `Player`). As for `MtabLock`, if there is a need to modify the locking scheme someday, composition is better since it forms loose coupling.

lolyu:
updated, thanks. @cacarrara

avocado/utils/partition.py
```diff
@@ -51,29 +51,32 @@ def __str__(self):
...
+            raise PartitionError(self, reason, e)
+        self.mtab = open(self.device)
         return self
 
     def __exit__(self, exc_type, exc_value, exc_traceback):
         if self.mtab:
             self.mtab.close()
+        self.lock.__exit__(exc_type, exc_value, exc_traceback)
+
+    def __getattr__(self, attrname):
```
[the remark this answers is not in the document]
lolyu: I thought it might be better to expose some attributes: `FileLock.filename`. If it is not a neccesity, I will delete it.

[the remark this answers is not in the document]
cacarrara: I don't think is a good approach. The file lock is still accessible by lock attribute so, probably we shouldn't change the __getattr__ method here.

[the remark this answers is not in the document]
lolyu: updated, please take a view.

PR #2992, opened 2019-01-08
yaml_to_mux: bump pyyaml version [v2]

cacarrara:
This changes addresses the CVE-2017-18342[1] that defines: "In PyYAML
before 4.1, the yaml.load() API could execute arbitrary code. In other
words, yaml.safe_load is not used."

The change affects only installation using Python 3 because when
building EL7 RPMs, we want to use version 3.10, which is packaged on
EPEL.

[1] - https://nvd.nist.gov/vuln/detail/CVE-2017-18342

---
Changes from v1 (#2990)
* Apply the PyYAML version bump only for Python 3 installation


cacarrara:
So, for Python 2 it's going to be necessary set `==3.10`, because 4.2b2 is >= 3.10. It's weird...

PR #2998, opened 2019-01-15
Be a bit more informative about plugin loading failures

pevogam:

### DIFF
--- a/avocado/core/output.py
+++ b/avocado/core/output.py
@@ -20,6 +20,7 @@ import logging
 import os
 import re
 import sys
+import traceback
 
 from six import string_types, iterkeys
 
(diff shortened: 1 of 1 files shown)


avocado/core/output.py
```diff
@@ -703,12 +704,13 @@ def log_plugin_failures(failures):
...
     silenced = settings.get_value('plugins',
                                   'skip_broken_plugin_notification',
                                   list, [])
     for failure in failures:
         if failure[0].module_name in silenced:
             continue
         LOG_UI.error(msg_fmt, failure[0].module_name,
-                     failure[1].__repr__())
+                     failure[1].__repr__(),
+                     ''.join(traceback.format_tb(failure[1].__traceback__)))
```
cacarrara: Hello, @pevogam. Thanks for your contribution.

I would like to suggest you to put a line break before the traceback log. It can make easier to read the output. What do you think?

Thanks.

pevogam: Hi @cacarrara,
You mean something like

```diff
        LOG_UI.error(msg_fmt, failure[0].module_name,
-                     failure[1].__repr__())
+                     failure[1].__repr__(),
+                     '\n' + ''.join(traceback.format_tb(failure[1].__traceback__)))
```
?

cacarrara: Or you can make a new call for `LOG_UI.error()`.

pevogam: I have pushed the most preferred method just now - an extra new line to separate the stack trace from the original error message. I have inspected the output and this is the best formatting I could see (e.g. compared to two lines if an extra `LOG_UI.error()` call is made before the above call.

PR #3001, opened 2019-01-22
improve avocado.utils.memory.read_from_meminfo

lolyu:
1. read directly from /proc/meminfo.
2. improve the regex to get more accurate result.


lolyu:
current issue:
```python
# python
>>> from avocado.utils.memory import read_from_meminfo as rfm
>>> rfm("DirectMap4k")
4
# grep DirectMap4k /proc/meminfo 
DirectMap4k:      306836 kB
```

PR #3019, opened 2019-02-01
utils.archive.extract_lzma: open in binary

lolyu:
fix that opens compressed file in binary instead of text mode.


lolyu:
#3020

PR #3115, opened 2019-05-05
Fix of tests execution from cit variants

richtja:
The variants were added to the runner in the wrong format. This commit solves the problem.


richtja:
@clebergnu thanks for the review. You are right I was using varianter_pict implementation and yours is a better solution so I will rewrite varianter_pict implementation as well.

PR #3119, opened 2019-05-13
utils/memory: add method to get all supported hugepage sizes

balamuruhans:
add `get_supported_huge_pages_size()` to get multiple hugepage
sizes supported by the host.

Suggested-by: Amador Pahim <amador@pahim.org>

### DIFF
--- a/avocado/utils/memory.py
+++ b/avocado/utils/memory.py
@@ -250,6 +250,18 @@ def get_page_size():
     return int(output)
 
 
+def get_supported_huge_pages_size():
+    """
+    Get all supported huge page sizes for this system.
+
+    :return: list of Huge pages size (kB).
+    """
+    output = os.listdir('/sys/kernel/mm/hugepages/')
+    # Given the items in this directory are in the format hugepages-<size>kB,
+    # the <size> will always start from index 10.
+    return [int(each[10:].rstrip('kB')) for each in output]
+
+
 def get_huge_page_size():
     """
     Get size of the huge pages for this system.


balamuruhans:
@clebergnu Thank you for reviewing, I have made the changes as suggested.

PraveenPenguin:
@clebergnu thanks for review

balamuruhans:
@clebergnu I made changes as suggested by you, request for your review and approval. Thanks!

PraveenPenguin:
@clebergnu can you please provide your ACK  if looks good to you as merging  is blocked ACK is needed form you

avocado/utils/memory.py
```diff
@@ -250,6 +250,16 @@ def get_page_size():
...
 
 
+def get_supported_huge_pages_size():
+    """
+    Get all supported huge page sizes for this system.
+
+    :return: list of Huge pages size (kB).
+    """
+    output = os.listdir('/sys/kernel/mm/hugepages/')
+    return [int(each.split('-')[1].strip('kB')) for each in output]
```
apahim: `strip()` removes characters from both left and right. Because you're interested only in the right end, you can make it a bit faster by using `rstrip()`.

apahim: About the split... you break the string into a list, then you get the list string item by index. You could make this faster by using a string slice, starting from the index right after the `-`, since the index of the `-` is fixed:
```python
>>> for file in os.listdir('/sys/kernel/mm/hugepages/'):
...     file.index('-')
...
9
```
So, just as a suggestion, already incorporating the `rstrip()`:
```python
output = os.listdir('/sys/kernel/mm/hugepages/')
# Given the items in this directory are in the format hugepages-<size>kB,
# the <size> will always start from index 10. 
return [int(each[10:].rstrip('kB')) for each in output]
```
We are talking about ~5% of performance improvement, not that much though.

avocado/utils/memory.py
```diff
@@ -250,6 +250,18 @@ def get_page_size():
...
 
+def get_supported_huge_pages_size():
+    """
+    Get all supported huge page sizes for this system.
+
+    :return: list of Huge pages size (kB).
+    """
+    output = os.listdir('/sys/kernel/mm/hugepages/')
+    # Given the items in this directory are in the format hugepages-<size>kB,
+    # the <size> will always start from index 10.
```
apahim: There's a trailing whitespace here making the CI unhappy.

avocado/utils/memory.py
balamuruhans: Thank you for the suggestion to improvement. I have adopted it.

apahim:
LGTM. @clebergnu, it's on you now :)

PR #3173, opened 2019-07-04
Added Power9 in get_cpu_vendor_name subroutine

PraveenPenguin:
Added Power9 in get_cpu_vendor_name subroutine as
in power9 system it returns None


PraveenPenguin:
without this Patch 
```Type "help", "copyright", "credits" or "license" for more information.
>>> print cpu.get_cpu_vendor_name()
None```

after this patch 

```Type "help", "copyright", "credits" or "license" for more information.
>>> print cpu.get_cpu_vendor_name()
power9
>>> ```

PR #3174, opened 2019-07-04
Added support in asset.fetch as locations can be single url

PraveenPenguin:
Added support in asset.fetch as locations can be single url
and that case it can be string url

### DIFF
--- a/avocado/utils/asset.py
+++ b/avocado/utils/asset.py
@@ -65,7 +65,7 @@ class Asset:
         :param name: the asset filename. url is also supported
         :param asset_hash: asset hash
         :param algorithm: hash algorithm
-        :param locations: list of locations where the asset can be fetched from
+        :param locations: location(s) where the asset can be fetched from
         :param cache_dirs: list of cache directories
         :param expire: time in seconds for the asset to expire
         """
@@ -75,7 +75,11 @@ class Asset:
             self.algorithm = DEFAULT_HASH_ALGORITHM
         else:
             self.algorithm = algorithm
-        self.locations = locations
+
+        if isinstance(locations, str):
+            self.locations = [locations]
+        else:
+            self.locations = locations
         self.cache_dirs = cache_dirs
         self.expire = expire
 


avocado/utils/asset.py
```diff
@@ -169,8 +170,11 @@ def fetch(self):
         # Now we have a writable cache_dir. Let's get the asset.
         # Adding the user defined locations to the urls list:
         if self.locations is not None:
-            for item in self.locations:
-                urls.append(item)
+            if isinstance(self.locations, list):
+                for item in self.locations:
+                    urls.append(item)
+            else:
+                urls.append(self.locations)
```
apahim: What if it's a tuple? :)
The beauty of the current implementation is that it works for any kind of sequence or iterator (lists, tuples, dictionaries, sets, custom iterators, ...).
The only exception is your case: a string. So, you should handle your case without breaking anything else:
```python
    def __init__(...):
        ...
        if isinstance(locations, str):
            self.locations = [locations]
        else:
            self.locations = locations
```
The rest stays the same.

PraveenPenguin: @apahim ah  yes right make sense , will do that

PraveenPenguin:
@apahim pushed the changes please have a look

avocado/utils/asset.py
```diff
@@ -65,7 +65,8 @@ def __init__(self, name, asset_hash, algorithm, locations, cache_dirs,
         :param name: the asset filename. url is also supported
         :param asset_hash: asset hash
         :param algorithm: hash algorithm
-        :param locations: list of locations where the asset can be fetched from
+        :param locations: list of locations/single location Url as String
+                          where the asset can be fetched from
```
apahim: Shall we rephrase this? We don't mention types for the other arguments... we can also omit that here now. Something like:
```
:param locations: location(s) where the asset can be fetched from
```
Considering that we now support basically anything, that's accurate and consistent.

PraveenPenguin: yes sure that right

PraveenPenguin:
@apahim adopted the suggested review comment , please have look

PR #3183, opened 2019-07-08
Fix: os.write expects string encode

harish-24:
Patch fixes genio utility with os.write expecting string encoding with python3.

ERROR|   File "avocado/utils/genio.py", line 143, in write_file_or_fail
ERROR|     os.write(fd, data)
ERROR| TypeError: a bytes-like object is required, not 'str'


harish-24:
@clebergnu Yes, that makes more sense. Pushed the suggested changes, Thanks.

PR #3188, opened 2019-07-09
Yaml_to_mux SafeLoader

richtja:
Usage of SafeLoader for ymal file instead of normal Loader, because Loader may construct an arbitrary Python object and that is not safe.

Reference: https://trello.com/c/1tTwO1Tu

### DIFF
--- a/optional_plugins/varianter_yaml_to_mux/avocado_varianter_yaml_to_mux/__init__.py
+++ b/optional_plugins/varianter_yaml_to_mux/avocado_varianter_yaml_to_mux/__init__.py
@@ -21,10 +21,11 @@ import re
 import sys
 
 import yaml
+
 try:
-    from yaml import CLoader as Loader
+    from yaml import CSafeLoader as SafeLoader
 except ImportError:
-    from yaml import Loader
+    from yaml import SafeLoader
 
 from avocado.core import exit_codes
 from avocado.core.output import LOG_UI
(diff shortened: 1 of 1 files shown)


richtja:
[yaml2mux: Use safeloader instead of the dangerous loader](https://trello.com/c/1tTwO1Tu/1492-yaml2mux-use-safeloader-instead-of-the-dangerous-loader)

optional_plugins/varianter_yaml_to_mux/avocado_varianter_yaml_to_mux/__init__.py
```diff
@@ -234,10 +239,10 @@ def mapping_to_tree_loader(loader, node, looks_like_node=False):
         _value = []
         for key_node, value_node in node.value:
             # Allow only strings as dict keys
-            if key_node.tag.startswith('!'):    # reflect tags everywhere
+            if key_node.tag.startswith('!'):  # reflect tags everywhere
                 key = loader.construct_object(key_node)
             else:
-                key = loader.construct_python_str(key_node)
+                key = loader.construct_scalar(key_node)
```
clebergnu: Took me a while to understand what if a scalar should be constructed here... according to 

https://yaml.org/spec/1.1/#scalar/information%20model

This seems correct.

richtja:
@clebergnu Thanks for the review, I restrict the commit just for relevant changes. Sorry about that, I totally miss that my editor reformats the file.

PR #3206, opened 2019-07-18
Merge of stdout/stderr.expected files

richtja:
Until now when avocado records multiple variants of one test, it creates stdout/stderr.expected for each variant. Now when the records are same for each variant user can use command line option --output-check-record-merge and avocado creates just one stdout/stderr.expected


clebergnu:
I had a different idea on how to solve this issue, possibly by using a contrib script to keep the extra logic out of  the test class (already pretty much complicated).  I've noticed this command line option help message:

"This should be enable only when the records are same for all variants."

Do you expect the user to have a say here?

richtja:
[Tool to de-duplicate/merge stdout/stderr.expected files](https://trello.com/c/tERMjBjy/1110-tool-to-de-duplicate-merge-stdout-stderrexpected-files)

richtja:
Hi @clebergnu, what do you think about this PR when #3263 was merged. This PR does almost the same thing. It is useful just for one case when user knows that expected files are equal and their size is bigger than memory. I think that we don't need this feature because this use-case is very rare. But I want to know your opinion.

PR #3226, opened 2019-07-29
Enable pylint w0212

richtja:
This pylint checks if a protected member is access outside the class or a descendant of the class where it’s defined.


selftests/inspekt-lint.sh
```diff
@@ -1,2 +1,3 @@
 #!/bin/sh -e
 inspekt lint --exclude=.git --enable R0401,W0101,W0102,W0104,W0105,W0106,W0107,W0108,W0109,W0111,W0120,W0122,W0123,W0124,W0125,W0143,W0150,W0199,W0211,W0222,W0223,W0232,W0233,W0301,W0311,W0312,W0401,W0404,W0406,W0410,W0601,W0602,W0603,W0604,W0611,W0612,W0613,W0614,W0621,W0622,W0623,W0631,W0640,W0702,W0705,W0711,W1201,W1202,W1300,W1301,W1302,W1303,W1304,W1305,W1306,W1307,W1401,W1402,W1501,W1503,W1645
+inspekt lint --exclude=.git,selftests --enable W0212
```
clebergnu: Why is this a new lint execution?

clebergnu: Oh, and BTW, if there's a valid reason for a new lint execution, there's a newline missing.

clebergnu:
The `job` case is kind of tricky, because of current bad design... I wouldn't like to make the results dispatcher a "public" attribute of the job, as it shouldn't be used outside of the job class (but currently is).  Let me think a bit more about the alternatives...

selftests/inspekt-lint.sh
richtja: > Why is this a new lint execution?

The new lint execution is because I think this check isn't valid for tests. So there is need for excluding test files for this check. If you have any ideas about how to solve this problem without running new lint execution let me know. 

> Oh, and BTW, if there's a valid reason for a new lint execution, there's a newline missing.

Thanks, I miss that.

richtja:

selftests/inspekt-lint.sh
clebergnu: Good point, there's a lot of extra work to stop using protected/private attributes in the selftests (if that's actually possible).

PR #3228, opened 2019-07-30
Enable pylint w0221

richtja:
This pylint checks if method arguments differ from the implemented interface or the overridden method.


richtja:

avocado/core/test.py
```diff
@@ -1028,31 +1028,31 @@ def _report(self):
             self.log.info("%s %s", self.status,
                           self.name)
 
-    def fail(self, message=None):
```
clebergnu: As simple as this change may seem, it may break tests out there, that is:

```
class MyTest(avocado.test):
   def test(self):
      self.fail(message="failing because...")
```

Won't work as expected after this.  We try very hard not to break the Test APIs:

https://avocado-framework.readthedocs.io/en/70.0/api/test/avocado.html

And a release right before LTS is the best place for this.  I recommend you:
 * Add "pylint: disable=W0221" marks here
 * Add a message to the docstring saying that this parameter will change name in the next LTS release (kind of a deprecation warning for the current parameter name)
 * Create a card on Trello and add the "LTS release" label to it

richtja: Yes, you are right, I will change that and prepare it for change before LTS release.

richtja:
Closing this because of #3235

PR #3229, opened 2019-07-30
Enable pylint w0231

richtja:
This pylint checks if an ancestor class method has an `__init__` method which is not called by a derived class.


richtja:

PR #3243, opened 2019-08-06
Merge of stdout/stderr.expected files second implemantation

richtja:
It is implemantation of `.expected` files merge, which after the job exceution cascadely merge equal `.expected` files  from variant level to file level. When the `.expected` files at the same level are equal, they are deleted and it is created new file in the upper level.

Reference: https://trello.com/c/tERMjBjy

### DIFF
--- a/avocado/plugins/run.py
+++ b/avocado/plugins/run.py
@@ -31,6 +31,7 @@ from avocado.core.dispatcher import JobPrePostDispatcher
 from avocado.core.settings import settings
 from avocado.utils.data_structures import time_to_seconds
 from avocado.utils import process
+from avocado.utils import expected_files_merge
 
 
 class Run(CLICmd):
@@ -224,6 +225,8 @@ class Run(CLICmd):
                 job_run = job_instance.run()
             finally:
                 # Run JobPost plugins
+                if args.output_check_record:
+                    expected_files_merge.merge_expected_files(job_instance.references)
                 pre_post_dispatcher.map_method('post', job_instance)
 
             result_dispatcher = ResultDispatcher()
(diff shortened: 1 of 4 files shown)


richtja:
[Tool to de-duplicate/merge stdout/stderr.expected files](https://trello.com/c/tERMjBjy/1110-tool-to-de-duplicate-merge-stdout-stderrexpected-files)

avocado/plugins/run.py
```diff
@@ -224,6 +225,8 @@ def run(self, args):
                 job_run = job_instance.run()
             finally:
                 # Run JobPost plugins
+                if args.output_check_record:
```
clebergnu: This is fine... but, to be even more in line with Avocado plugin mechanism, it'd make sense to make this a plugin itself.  The plugin can check if the option was given and then merge the files.  With that, this line would not be necessary.

avocado/utils/genio.py
```diff
@@ -191,3 +191,21 @@ def is_pattern_in_file(filename,  pattern):
         if re.search(pattern, content_file.read(), re.MULTILINE):
             return True
     return False
+
+
+def is_files_equal(filename, other):
```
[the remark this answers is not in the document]
richtja: Great idea thanks, that will be much faster.

avocado/plugins/run.py
richtja: Ok, I will create a plugin for this.

richtja:
@clebergnu Thanks for the review. I will reproduce and solve the problem.

richtja:
Closing because of #3259

PR #3293, opened 2019-09-06
Remove old libvirt dependency from requirements-selftests.txt

willianrampazzo:
As libvirt is handled in the spec file, it is not necessary to
install it using pip anymore.


willianrampazzo:
Looks like libvirt is still needed when installing purely using pip.

PR #3298, opened 2019-09-09
The path.usable_ro_dir shouldn't create a dir

willianrampazzo:
Change the behavior of usable_ro_dir so it does not create a dir.
If a new dir is needed, usable_rw_dir should be used.

Reference: https://trello.com/c/xsKNnIYu/


willianrampazzo:
[The `path.usable_ro_dir` shouldn't create the dir](https://trello.com/c/xsKNnIYu/1104-the-pathusablerodir-shouldnt-create-the-dir)

PR #3333, opened 2019-10-03
[RFC/POC] Add initial support to 'assets fetch' command with a simple parser [v1]

willianrampazzo:
This is the v1 of the `assets fetch` command initially discussed at [1]. This implementation uses a parser on instrumented test source to find `fetch_asset` calls composed of simple strings as parameters, and fetch those assets without running the test.

```
$ avocado assets fetch -h
usage: avocado assets fetch [-h] [TEST_REFERENCE [TEST_REFERENCE ...]]

positional arguments:
  TEST_REFERENCE  List of test references

optional arguments:
  -h, --help      show this help message and exit
```
[1] https://www.redhat.com/archives/avocado-devel/2019-September/msg00009.html

ps. v1 missing functional/unit tests. They will be added on v2.

### DIFF
--- a/avocado/plugins/assets.py
+++ b/avocado/plugins/assets.py
@@ -1,0 +1,179 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: Red Hat Inc. 2019
+# Authors: Willian Rampazzo <willianr@redhat.com>
+
+import ast
+import errno
+import json
+import os
+import tempfile
+
+from avocado.utils import process
+from avocado.core.output import LOG_UI
+from avocado.core.plugin_interfaces import CLICmd
+
+
+TEST_TEMPLATE = r"""
+from avocado import Test
+
+class FetchAssets(Test):
+    def test_fetch_assets(self):
+{fetch_assets}
+"""
+
+
+class FetchAssetHandler(ast.NodeVisitor):
+
+    def __init__(self, file_name):
+        self.file_name = file_name
+        # Right now, the class handles just 'fetch_assets' call.
+        # In the future, it can be extended to search for other call patterns.
+        self.pattern = 'fetch_asset'
+        self.calls = []
+        self.result = None
+
+        # create Abstract Syntax Tree from test source file
+        with open(self.file_name) as source_file:
+            self.tree = ast.parse(source_file.read(), self.file_name)
+
+        # build calls list
+        self.visit(self.tree)
+    # __init__()
+
+    def _build_call(self, node):
+        """
+        Build the call string from ast nodes.
+        :param node: Call root node
+        :type node: ast.Call
+        :return: Call prepared to be executed into a test class
+        :rtype: str
+        """
+        # we handle code indentation
+        call = "        self." + node.func.attr + "("
+        for a in node.args:
+            if isinstance(a, ast.Name):
+                return None
+            elif isinstance(a, ast.Str):
+                call = call + '"' + ast.literal_eval(a) + '", '
+        for k in node.keywords:
+            call = call + k.arg + "="
+            if isinstance(k.value, ast.Name):
+                return None
+            elif isinstance(k.value, ast.Str):
+                call = call + '"' + ast.literal_eval(k.value) + '", '
+        # remove last comma and add parenthesis
+        call = call[:-2] + ")\n"
+
+        return call
+    # _build_call()
+
+    def visit_Call(self, node):
+        """
+        Visit calls and build call string if call matches the pattern
+        :param node: (Sub)Tree root to start the parse
+        :type node: ast.*
+        """
+        if isinstance(node.func, ast.Attribute):
+            if self.pattern in node.func.attr:
+                call = self._build_call(node)
+                if call is not None:
+                    self.calls.append(call)
+    # visit_call()
+
+    def fetcher(self):
+        """
+        Build an intrumented avocado test with all fetch_asset calls and
+        execute it.
+        """
+        # build raw code from list of calls, if available
+        if len(self.calls) > 0:
+            rcalls = ""
+            for call in self.calls:
+                rcalls = rcalls + call
+        else:
+            LOG_UI.debug('No supported assets found to fetch from %s'
+                         % self.file_name)
+            return
+
+        # temporary test source file
+        fetch_test = tempfile.NamedTemporaryFile(suffix=".py")
+        fetch_test.write(TEST_TEMPLATE.format(fetch_assets=rcalls).encode())
+        fetch_test.flush()
+
+        # run teporary test
+        LOG_UI.debug('Fetching assets from %s' % self.file_name)
+        self.result = process.run("avocado run --json - %s" % fetch_test.name,
+                                  ignore_status=True)
+        fetch_test.close()
+
+        # check if test ran without problems
+        fetch_test_stdout = json.loads(self.result.stdout)
+        if (self.result.exit_status != 0
+                or fetch_test_stdout['pass'] != fetch_test_stdout['total']):
+            LOG_UI.error('%s Check %s for more details.' % (
+                fetch_test_stdout['tests'][0]['fail_reason'],
+                fetch_test_stdout['tests'][0]['logfile']))
+    # fetcher()
+# FetchAssetHandler()
+
+
+class Assets(CLICmd):
+    """
+    Implements the avocado 'assets' subcommand
+    """
+    name = 'assets'
+    description = 'Manage assets'
+
+    def configure(self, parser):
+        """
+        Add the subparser for the assets action.
+
+        :param parser: Main test runner parser.
+        """
+        parser = super(Assets, self).configure(parser)
+
+        subcommands = parser.add_subparsers(dest='assets_subcommand')
+        subcommands.required = True
+
+        fetch_subcommand_parser = subcommands.add_parser('fetch',
+                                help='Fetch assets from test source'\
+                                ' or config file if it\'s not already in the'\
+                                ' cache')
+        fetch_subcommand_parser.add_argument('references', type=str,
+                                             default=[], nargs='*',
+                                             metavar='TEST_REFERENCE',
+                                             help='List of test references')
+    # configure()
+
+    def run(self, config):
+        subcommand = config.get('assets_subcommand')
+
+        if subcommand == 'fetch':
+
+            # check if we have something to work on
+            if len(config['references']) < 1:
+                LOG_UI.error('No test references provided nor any other' \
+                               ' arguments.')
+
+            # fetch assets from instrumented tests
+            fetch_asset_handler = {}
+            for test_file in config['references']:
+                if os.path.isfile(test_file) and test_file.endswith('.py'):
+                    fetch_asset_handler[test_file] = FetchAssetHandler(
+                                                                    test_file)
+                    fetch_asset_handler[test_file].fetcher()
+                else:
+                    LOG_UI.warn('No such file or file not supported: %s',
+                                test_file)
+    # run()
+
--- a/setup.py
+++ b/setup.py
@@ -97,6 +97,7 @@ if __name__ == '__main__':
                   'nrun = avocado.plugins.nrun:NRun',
                   'vmimage = avocado.plugins.vmimage:VMimage',
                   'nlist = avocado.plugins.nlist:List',
+                  'assets = avocado.plugins.assets:Assets',
                   ],
               'avocado.plugins.job.prepost': [
                   'jobscripts = avocado.plugins.jobscripts:JobScripts',


willianrampazzo:
@clebergnu, @apahim, @ldoktor feel free to play with this initial implementation!

clebergnu:
@willianrampazzo I'm on it.

avocado/plugins/assets.py
```diff
@@ -0,0 +1,179 @@
...
+
+        subcommands = parser.add_subparsers(dest='assets_subcommand')
+        subcommands.required = True
+
+        fetch_subcommand_parser = subcommands.add_parser('fetch',
+                                help='Fetch assets from test source'\
+                                ' or config file if it\'s not already in the'\
+                                ' cache')
+        fetch_subcommand_parser.add_argument('references', type=str,
+                                             default=[], nargs='*',
```
ldoktor: There is no need to specify `type=str`, nor `default=[]` when `nargs='*'` is used. Also I'd recommend to use `nargs='+'` which always requires at least one argument (therefor check on line 164 would become unnecessary).

avocado/plugins/assets.py
```diff
@@ -0,0 +1,179 @@
...
+                                             metavar='TEST_REFERENCE',
+                                             help='List of test references')
+    # configure()
```
ldoktor: ... marking the ends of classes? Never used it myself and at first looks likes TODO or forgotten debug line, but pylint is happy and someone might find it useful...

avocado/plugins/assets.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def run(self, config):
+        subcommand = config.get('assets_subcommand')
+
+        if subcommand == 'fetch':
```
ldoktor: I know the argparse should not let the `assets_subcommand` to be different than supported ones, but just to be safe I'd recommend adding `else` with error message about unsupported subcommand. That way we'd spot typos easier...

avocado/plugins/assets.py
```diff
@@ -0,0 +1,179 @@
...
+
+            # check if we have something to work on
+            if len(config['references']) < 1:
```
ldoktor: This shouldn't be necessary when `nargs='+'` is used...

avocado/plugins/assets.py
```diff
@@ -0,0 +1,179 @@
...
+                LOG_UI.error('No test references provided nor any other' \
+                               ' arguments.')
+
+            # fetch assets from instrumented tests
+            fetch_asset_handler = {}
+            for test_file in config['references']:
+                if os.path.isfile(test_file) and test_file.endswith('.py'):
+                    fetch_asset_handler[test_file] = FetchAssetHandler(
+                                                                    test_file)
+                    fetch_asset_handler[test_file].fetcher()
```
ldoktor: The method `fetcher` doesn't make much sense to my non-native ear. How about `fetch` or `__call__`? (or could you elaborate on what the `fetcher` should stand for?

avocado/plugins/assets.py
```diff
@@ -0,0 +1,179 @@
...
+
+        if subcommand == 'fetch':
+
+            # check if we have something to work on
+            if len(config['references']) < 1:
+                LOG_UI.error('No test references provided nor any other' \
+                               ' arguments.')
+
+            # fetch assets from instrumented tests
+            fetch_asset_handler = {}
```
ldoktor: Hmm, why are you storing the FetchAssetHandlers? I don't see any use of it...

avocado/plugins/assets.py
```diff
@@ -0,0 +1,179 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: Red Hat Inc. 2019
+# Authors: Willian Rampazzo <willianr@redhat.com>
+
+import ast
+import errno
```
ldoktor: This is why the CI is failing (unused import)

ldoktor: ... btw please use `pylint` and fix as much issues as possible (some doesn't make sense, but in most cases we do want 10 out of 10 score...)

ldoktor: ... also, cool PR number and about the commit message, it's not necessary to include the `This is the v1 ...` stuff. Simply describe what this is trying to solve and eventually the limitations....

avocado/plugins/assets.py
```diff
@@ -0,0 +1,179 @@
...
+        Build the call string from ast nodes.
+        :param node: Call root node
+        :type node: ast.Call
+        :return: Call prepared to be executed into a test class
+        :rtype: str
+        """
+        # we handle code indentation
+        call = "        self." + node.func.attr + "("
+        for a in node.args:
+            if isinstance(a, ast.Name):
```
ldoktor: Still processing this loop, but I'd suggest:

1. using longer variable names (arg and kwarg?)
2. add comment before the `None` explaining that specifying arguments from variables is not yet supported. It's obvious now, but in 1, 2 or 10 years we'll have to go through the code and think about it again ;-)

PS: According to github statistics Czechs are the most chatty developers, we just like comments :-D

avocado/plugins/assets.py
willianrampazzo: > This is why the CI is failing (unused import)

This was forgotten as part of a previous implementation idea. I already removed it from the next version.

> ... btw please use `pylint` and fix as much issues as possible (some doesn't make sense, but in most cases we do want 10 out of 10 score...)

Sure, will do.

> ... also, cool PR number and about the commit message, it's not necessary to include the `This is the v1 ...` stuff. Simply describe what this is trying to solve and eventually the limitations....

I just wanted to make clear that this implementation is not intended to be a final version. I will improve it in the next version.

avocado/plugins/assets.py
```diff
@@ -0,0 +1,179 @@
...
+                return None
+            elif isinstance(a, ast.Str):
```
ldoktor: Well on one side you return None when variable is used, while on the other hand you silently skip everything that is not string (including lists).

I'd suggest skipping all values that can't be parsed (`ast.Name`) while including all values that can be extracted (including lists), eventually checking we have at least the bare-minimum to execute the fetch_asset call (one argument or name kwarg; although name might not be enough without locations... still we can live with partial knowledge)...

Alternatively I'd suggest chickening-out any time we can't parse something, therefore `else: return None` (safest but we'd not download calls with `hashes` or other "unimportant" things specified via variable...)

avocado/plugins/assets.py
```diff
@@ -0,0 +1,179 @@
...
+                call = call + '"' + ast.literal_eval(a) + '", '
+        for k in node.keywords:
+            call = call + k.arg + "="
+            if isinstance(k.value, ast.Name):
+                return None
+            elif isinstance(k.value, ast.Str):
+                call = call + '"' + ast.literal_eval(k.value) + '", '
```
ldoktor: Hmm the situation from line 66 is even worse here, because partial matches leads to `call = 'self.fetch_asset(name="hello-2.9.tar.gz", location)'`, which is obviously wrong (locations was list)

avocado/plugins/assets.py
willianrampazzo: Old habits... I can remove it to keep some coding standards in the project.

avocado/plugins/assets.py
```diff
@@ -0,0 +1,179 @@
...
+                    self.calls.append(call)
+    # visit_call()
+
+    def fetcher(self):
+        """
+        Build an intrumented avocado test with all fetch_asset calls and
+        execute it.
+        """
+        # build raw code from list of calls, if available
+        if len(self.calls) > 0:
```
ldoktor: There is no need to use `len`, simple `if self.calls` should do.

avocado/plugins/assets.py
willianrampazzo: There is already an error message for unsupported subcommands by default:

```
$ avocado assets fetche
usage: avocado assets [-h] {fetch} ...

positional arguments:
  {fetch}
    fetch     Fetch assets from test source or config file if it's not already
              in the cache

optional arguments:
  -h, --help  show this help message and exit

avocado assets: error: argument assets_subcommand: invalid choice: 'fetche' (choose from 'fetch')
```

That is the reason I have not added another check, as I would expect other commands from avocado follow the same to keep subcommand error messages standard.

avocado/plugins/assets.py
willianrampazzo: The first implementation I had in mind was considering to handle the results from the download out of the object, but in the end, I just implemented it into the same class, so, this indeed is not necessary.

avocado/plugins/assets.py
```diff
@@ -0,0 +1,179 @@
...
+            return
+
+        # temporary test source file
+        fetch_test = tempfile.NamedTemporaryFile(suffix=".py")
+        fetch_test.write(TEST_TEMPLATE.format(fetch_assets=rcalls).encode())
+        fetch_test.flush()
+
+        # run teporary test
+        LOG_UI.debug('Fetching assets from %s' % self.file_name)
+        self.result = process.run("avocado run --json - %s" % fetch_test.name,
```
ldoktor: And now about this, I don't really like the idea of creating a test file just to fetch the assets. Not sure what am I missing here, but wouldn't it be safer and easier to simply use:

```
cache_dirs = data_dir.get_cache_dirs()
err = []
for fetch_item in fetch_assets:
    try:
        expire = fetch_item.kwargs.pop('expire', None)
        # 5th positional argument is 6th argument in asset.Assets
        if len(fetch_item.args) == 5:
            expire = fetch_item.args[4]
            fetch_item.args = fetch_item.args[:4]
        elif len(fetch_item.args) > 5:
            err.append(...)
            continue
        if expire is not None:
        asset.Asset(*fetch_item.args, **fetch_item.kwargs, cache_dirs=cache_dirs, expire=expire)
    except ...:
        err.append(...)
```

... only after preparing this I noticed the `metadata` new argument in `asset.Asset`, it might need to also be treated. Although it's not supported by `self.fetch_asset` so either we should add it there as well, or not to care about it at all.

ldoktor: ^^ btw this was written in a hurry and you might need to improve it (checking for duplicate args...), but it should serve as an inspiration.

ldoktor: ... and I forgot to mention that the `fetch_assets` is a list of arguments collected in `_build_call` (stored in `self.calls`)

ldoktor: also note that by calling it directly you'd actually reflect potential `avocado --config ... assets fetch ...` changes. By calling another `avocado` process you might end-up fetching the assets to a different location(s).

avocado/plugins/assets.py
willianrampazzo: > ... btw please use `pylint` and fix as much issues as possible (some doesn't make sense, but in most cases we do want 10 out of 10 score...)

I just noticed pylint was disabled in my editor and I thought it was enabled. I enabled it and now I see all the pylint warning. Thanks for that!

avocado/plugins/assets.py
willianrampazzo: > 1. using longer variable names (arg and kwarg?)
> 2. add comment before the `None` explaining that specifying arguments from variables is not yet supported. It's obvious now, but in 1, 2 or 10 years we'll have to go through the code and think about it again ;-)

Makes sense, will do.
 
> PS: According to github statistics Czechs are the chattiest developers, we just like comments :-D

And I really appreciate that! :D

avocado/plugins/assets.py
willianrampazzo: I like to do explicitly checks where possible to make the code more readable, but I noticed pylint complains about it. I will change that.

avocado/plugins/assets.py
willianrampazzo: The main objective here was to make use of all logging structure avocado command gives.

> also note that by calling it directly you'd actually reflect potential `avocado --config ... assets fetch ...` changes. By calling another `avocado` process you might end-up fetching the assets to a different location(s).

When considering this use case, the current implementation does not fit it, so I will think about other ways to implement that, taking into consideration your suggestions.

avocado/plugins/assets.py
willianrampazzo: This method is into `FetchAssetHandler` class and is responsible for fetching the assets. My first idea was to call it `downloader,` but everywhere else naming follows the `fetch` convention, so `fetcher` for me makes more sense, keeps the naming convention, and has an explicit call to it. Using `__call__` IMHO, sometimes makes reading the code difficult to understand what that object is really doing without looking at the source `__call__` method.

clebergnu:
Hi Willian,

I'm not going to the lower level details of the code at this point,
given that it's an RFC/POC. I'll focus on the overall design and
strategy and how I see you

1) The strategy of using a Python AST parser doesn't make the solution
100% fail proof, but we should attempt to cover the basic scenarios.
For instance, look at this hypothetical BitTorrent project, and
suppose this is part of of a user's test file:

```python
class TorrentAssembler:
    """
    Hypothetical class that assembles a file from a torrent definition

    It may use local assets, that is, torrent blocks that may have been
    downloaded and cached
    """
    def fetch_asset(self, asset):
        """
        This is torrent specific, and 
        """
        path = os.path.join('.cache', asset)
        if os.path.isfile(path):
            return open(path).read()
        else:
            return self.download(asset)

    def assemble(self):
        if self.fetch_asset('BLOCK1'):
            return True
```

And then this is followed by:

```python
from avocado import Test

class TorrentBlockFetcherTest(Test):
    """
    Tests for the torrent block fetch functionality
    """
    def test(self):
        asset = fetch_asset('http://torrent.org/reference.torrent')
        self.assertEqual(TorrentAssemble(asset).assemble())
```

The call to TorrentAssembler.fetch_asset() would be treated as an
avocado.Test.fetch_asset().  There's a simple solution to this, which
is to rely on `avocado.core.safeloader.find_avocado_tests()`, and only
process the fetch_assets from the classes/methods that it finds.

---

2) A more complete parsing of the arguments may be necessary to
fulfill the basic use cases.  For instance, consider that the user
specifies:

     self.fetch_asset('https://archives.fedoraproject.org/pub/archive/fedora/linux/releases/29/Everything/x86_64/os/images/pxeboot/vmlinuz',
                      asset_hash='23bebd2680757891cf7adedb033532163a792495')

The parser should not discard the `asset_hash` parameter (and its value)
given that only by having both we can guarantee that we have effetcively
donwloaded a correct "version" of the asset.

---

3) Once all necessary information has been correctly parsed, the right
layer should be used to fetch the asset.  As Lukas pointed out, I see
no reason to *not* use `avocado.utils.assets.Asset` directly.  If it
currently lacks anything, such as better error reporting, that is
something that needs to be added/fixed.

---

4) As for the supported use case, this may serve as a baseline:

```python
from avocado import Test
        
class FetchKernel(Test):

    def test(self):
        kernel_url = ('https://archives.fedoraproject.org/pub/archive/fedora'
                      '/linux/releases/29/Everything/x86_64/os/images/pxeboot'
                      '/vmlinuz')
        kernel_hash = '23bebd2680757891cf7adedb033532163a792495'
        kernel_path = self.fetch_asset(kernel_url, asset_hash=kernel_hash)
```

It's basically a copy of a similar QEMU test that originated this discussion
and feature.  I've played with the AST parser, and it should be easy that,
once within the scope of the test method.

```python
import ast

module = ast.parse(open('fetch_kernel.py').read(), 'fetch_kernel.py')
# first is the import statement
_, klass = module.body
test_method = klass.body[0]

assignments = {}

for statement in test_method.body:
    if isinstance(statement, ast.Assign):
        if isinstance(statement.value, ast.Str):
            var_name = statement.targets[0].id
            assignments[var_name] = statement.value.s
```

That should be enough to support fetch_asset() calls that refer to
variables that are *strings*.  What we don't want to get into is
handling assigments that are expressions that would have to be
evaluated.

---

I guess this leaves you with enough to work on towards a v2.  Let me
know if I seem to get any of this wrong, or if I can help in any other
way.

Thanks!

willianrampazzo:
> Hello @willianrampazzo, overall I do like this quite a lot. The parsing makes sense and walks the files pretty well, although I'd suggest keeping the `args/kwargs` and calling the `asset.Asset` directly. That should give you actually per-asset granularity for reporting purposes (following 21 assets failed to fetch: details...)
> 
> Then there are some little python issues here and there, please use `pylint` before sending patches (not necessarily always, but at least from time to time) and consider the argparse suggestions.
> 
> Looking forward to the v2.

Thanks for your review @ldoktor, I will take the suggestions into consideration while working on v2.

willianrampazzo:
> 1. The strategy of using a Python AST parser doesn't make the solution
>    100% fail proof, but we should attempt to cover the basic scenarios.

Indeed, this is a situation I did not think about, thanks for that!

> 2. A more complete parsing of the arguments may be necessary to
>    fulfill the basic use cases.  For instance, consider that the user
>    specifies:
>    self.fetch_asset('https://archives.fedoraproject.org/pub/archive/fedora/linux/releases/29/Everything/x86_64/os/images/pxeboot/vmlinuz',
>    asset_hash='23bebd2680757891cf7adedb033532163a792495')
> 
> The parser should not discard the `asset_hash` parameter (and its value)
> given that only by having both we can guarantee that we have effetcively
> donwloaded a correct "version" of the asset.

This is already covered in the code, from lines 68 to 73.

> 3. Once all necessary information has been correctly parsed, the right
>    layer should be used to fetch the asset.  As Lukas pointed out, I see
>    no reason to _not_ use `avocado.utils.assets.Asset` directly.  If it
>    currently lacks anything, such as better error reporting, that is
>    something that needs to be added/fixed.

Okay, I will work on it.

> 4. As for the supported use case, this may serve as a baseline:
> 
> ```python
> from avocado import Test
>         
> class FetchKernel(Test):
> 
>     def test(self):
>         kernel_url = ('https://archives.fedoraproject.org/pub/archive/fedora'
>                       '/linux/releases/29/Everything/x86_64/os/images/pxeboot'
>                       '/vmlinuz')
>         kernel_hash = '23bebd2680757891cf7adedb033532163a792495'
>         kernel_path = self.fetch_asset(kernel_url, asset_hash=kernel_hash)
> ```
> 
> It's basically a copy of a similar QEMU test that originated this discussion
> and feature. I've played with the AST parser, and it should be easy to get
> the needed values once within the scope of the test method.
> 
> ```python
> import ast
> 
> module = ast.parse(open('fetch_kernel.py').read(), 'fetch_kernel.py')
> # first is the import statement
> _, klass = module.body
> test_method = klass.body[0]
> 
> assignments = {}
> 
> for statement in test_method.body:
>     if isinstance(statement, ast.Assign):
>         if isinstance(statement.value, ast.Str):
>             var_name = statement.targets[0].id
>             assignments[var_name] = statement.value.s
> ```
> 
> That should be enough to support fetch_asset() calls that refer to
> variables that are _strings_. What we don't want to get into is
> handling assigments that are expressions that would have to be
> evaluated.

Here we go back to the discussion we had on the mailing list about how much we want to cover in the parser. My first implementation was covering what you mentioned before I started to find complex situations and send the e-mail to the mailing list.

Considering your suggestion, an example I found when I started covering it was the multiple assignment definitions before the `fetch_asset` call without any relation to it or the reuse of the same variable name in the same test or across the tests. The parser needs to find the `fetch_asset` call, check if there are variables at args or right side of keyword args, then walk back in the code looking for those definitions, because we may have multiple definitions of the same variable into the test or across multiple tests, or those variables could be defined as global, outside of the class. Here is an example that covers some of the situations:

```
class Hello(Test):

    def test_fetch_hello(self):
        mirrors = 'https://mirrors.peers.community/mirrors/gnu/hello/'
        hello = 'hello-2.9.tar.gz'
        hello_locations = "https://mirrors.peers.community/mirrors/gnu/hello/hello-2.9.tar.gz"
        message = "nothing related to the fetch_asset"
        hello_sig_locations = "https://mirrors.peers.community/mirrors/gnu/hello/hello-2.9.tar.gz.sig"
        self.hello = self.fetch_asset(
            name=hello,
            locations=hello_locations)
        self.hello_sig = self.fetch_asset(
            name=hello_sig,
            asset_hash='f3b9fae20c35740004ae7b8de1301836dab4ac30',
            locations=hello_sig_locations)

        mirrors = 'https://mirrors.kernel.org/gnu/hello/'
        hello = 'hello-2.8.tar.gz'
        hello_locations = "https://mirrors.kernel.org/gnu/hello/hello-2.8.tar.gz"
        message = "nothing related to the fetch_asset"
        hello_sig = 'hello-2.8.tar.gz.sig'
        hello_sig_locations = "https://mirrors.kernel.org/gnu/hello/hello-2.8.tar.gz.sig"
        self.hello = self.fetch_asset(
            name=hello,
            locations=hello_locations)
        self.hello_sig = self.fetch_asset(
            name=hello_sig,
            asset_hash='f3b9fae20c35740004ae7b8de1301836dab4ac30',
            locations=hello_sig_locations)
```
Lukas also mentioned supporting lists. When I suggested supporting just pure string parameters was to make a parser that supports a minimal case, other complex cases would be supported into the configuration file, as suggested in the mailing list.

So, to make discussion shorter, my main concern here is to define how much we want to cover into the parser. With that definition, I can come with a solution that fits it.

> I guess this leaves you with enough to work on towards a v2. Let me
> know if I seem to get any of this wrong, or if I can help in any other
> way.

Thanks a lot for your suggestions as those will help to come up with a better implementation!

clebergnu:
> > 1. The strategy of using a Python AST parser doesn't make the solution
> >    100% fail proof, but we should attempt to cover the basic scenarios.
> 
> Indeed, this is a situation I did not think about, thanks for that!
> 
> > 1. A more complete parsing of the arguments may be necessary to
> >    fulfill the basic use cases.  For instance, consider that the user
> >    specifies:
> >    self.fetch_asset('https://archives.fedoraproject.org/pub/archive/fedora/linux/releases/29/Everything/x86_64/os/images/pxeboot/vmlinuz',
> >    asset_hash='23bebd2680757891cf7adedb033532163a792495')
> > 
> > The parser should not discard the `asset_hash` parameter (and its value)
> > given that only by having both we can guarantee that we have effetcively
> > donwloaded a correct "version" of the asset.
> 
> This is already covered in the code, from lines 68 to 73.
> 

You're right, I missed that.  For some reason I experimented with some `fetch_asset()` calls and it seemed that keyword parameters were not supported.

> > 1. Once all necessary information has been correctly parsed, the right
> >    layer should be used to fetch the asset.  As Lukas pointed out, I see
> >    no reason to _not_ use `avocado.utils.assets.Asset` directly.  If it
> >    currently lacks anything, such as better error reporting, that is
> >    something that needs to be added/fixed.
> 
> Okay, I will work on it.
> 
> > 1. As for the supported use case, this may serve as a baseline:
> > 
> > ```python
> > from avocado import Test
> >         
> > class FetchKernel(Test):
> > 
> >     def test(self):
> >         kernel_url = ('https://archives.fedoraproject.org/pub/archive/fedora'
> >                       '/linux/releases/29/Everything/x86_64/os/images/pxeboot'
> >                       '/vmlinuz')
> >         kernel_hash = '23bebd2680757891cf7adedb033532163a792495'
> >         kernel_path = self.fetch_asset(kernel_url, asset_hash=kernel_hash)
> > ```
> > 
> > 
> > It's basically a copy of a similar QEMU test that originated this discussion
> > and feature. I've played with the AST parser, and it should be easy to get
> > the needed values once within the scope of the test method.
> > ```python
> > import ast
> > 
> > module = ast.parse(open('fetch_kernel.py').read(), 'fetch_kernel.py')
> > # first is the import statement
> > _, klass = module.body
> > test_method = klass.body[0]
> > 
> > assignments = {}
> > 
> > for statement in test_method.body:
> >     if isinstance(statement, ast.Assign):
> >         if isinstance(statement.value, ast.Str):
> >             var_name = statement.targets[0].id
> >             assignments[var_name] = statement.value.s
> > ```
> > 
> > 
> > That should be enough to support fetch_asset() calls that refer to
> > variables that are _strings_. What we don't want to get into is
> > handling assigments that are expressions that would have to be
> > evaluated.
> 
> Here we go back to the discussion we had on the mailing list about how much we want to cover in the parser. My first implementation was covering what you mentioned before I started to find complex situations and send the e-mail to the mailing list.
> 

Maybe I wasn't clear, or just was plain wrong, about what I considered the basic use case.  Basically, what I meant was based on:

 * The use case that motivated this feature (the QEMU test)
 * The fact that we should abstain from trying to *execute* Python code to get the information we need

That's why I suggested we should stick to string assignments.  But, we won't be able to support the basic QEMU tests, unless we support variable assignments (of plain strings).

> Considering your suggestion, an example I found when I started covering it was the multiple assignment definitions before the `fetch_asset` call without any relation to it or the reuse of the same variable name in the same test or across the tests. The parser needs to find the `fetch_asset` call, check if there are variables at args or right side of keyword args, then walk back in the code looking for those definitions, because we may have multiple definitions of the same variable into the test or across multiple tests, or those variables could be defined as global, outside of the class. Here is an example that covers some of the situations:

I don't think the parser needs to find the `fetch_asset` call and then walk back.  You can parse/store all assignments within the scope of one test.  If two assignments for the same variable exist before before a `fetch_asset` call, they should just update the previously found definition (a simple loop over the statements should do just that quite naturally). 

> 
> ```
> class Hello(Test):
> 
>     def test_fetch_hello(self):
>         mirrors = 'https://mirrors.peers.community/mirrors/gnu/hello/'
>         hello = 'hello-2.9.tar.gz'
>         hello_locations = "https://mirrors.peers.community/mirrors/gnu/hello/hello-2.9.tar.gz"
>         message = "nothing related to the fetch_asset"
>         hello_sig = 'hello-2.9.tar.gz.sig'
>         hello_sig_locations = "https://mirrors.peers.community/mirrors/gnu/hello/hello-2.9.tar.gz.sig"
>         self.hello = self.fetch_asset(
>             name=hello,
>             locations=hello_locations)

I see no issues here the statement-by-statement approach.  Running this:

```python
import ast
import pprint

module = ast.parse(open('hello.py').read(), 'hello.py')
# first is the import statement
_, klass = module.body
test_method = klass.body[0]

assignments = {}

for statement in test_method.body:
    if isinstance(statement, ast.Assign):
        if isinstance(statement.value, ast.Str):
            var_name = statement.targets[0].id
            assignments[var_name] = statement.value.s
        elif isinstance(statement.value, ast.Call):
            if statement.value.func.attr == 'fetch_asset':
                print('>>> fetch_asset current assigments:')
                pprint.pprint(assignments)
```

Gets me:

```
>>> fetch_asset current assigments:
{'hello': 'hello-2.9.tar.gz',
 'hello_locations': 'https://mirrors.peers.community/mirrors/gnu/hello/hello-2.9.tar.gz',
 'hello_sig': 'hello-2.9.tar.gz.sig',
 'hello_sig_locations': 'https://mirrors.peers.community/mirrors/gnu/hello/hello-2.9.tar.gz.sig',
 'message': 'nothing related to the fetch_asset',
 'mirrors': 'https://mirrors.peers.community/mirrors/gnu/hello/'}
>>> fetch_asset current assigments:
{'hello': 'hello-2.9.tar.gz',
 'hello_locations': 'https://mirrors.peers.community/mirrors/gnu/hello/hello-2.9.tar.gz',
 'hello_sig': 'hello-2.9.tar.gz.sig',
 'hello_sig_locations': 'https://mirrors.peers.community/mirrors/gnu/hello/hello-2.9.tar.gz.sig',
 'message': 'nothing related to the fetch_asset',
 'mirrors': 'https://mirrors.peers.community/mirrors/gnu/hello/'}
>>> fetch_asset current assigments:
{'hello': 'hello-2.8.tar.gz',
 'hello_locations': 'https://mirrors.kernel.org/gnu/hello/hello-2.8.tar.gz',
 'hello_sig': 'hello-2.8.tar.gz.sig',
 'hello_sig_locations': 'https://mirrors.kernel.org/gnu/hello/hello-2.8.tar.gz.sig',
 'message': 'nothing related to the fetch_asset',
 'mirrors': 'https://mirrors.kernel.org/gnu/hello/'}
>>> fetch_asset current assigments:
{'hello': 'hello-2.8.tar.gz',
 'hello_locations': 'https://mirrors.kernel.org/gnu/hello/hello-2.8.tar.gz',
 'hello_sig': 'hello-2.8.tar.gz.sig',
 'hello_sig_locations': 'https://mirrors.kernel.org/gnu/hello/hello-2.8.tar.gz.sig',
 'message': 'nothing related to the fetch_asset',
 'mirrors': 'https://mirrors.kernel.org/gnu/hello/'}
```
Which seems to match the code.  We may abort if we find anything that is not a string for the first implementation, but that would still solve the basic real QEMU case.

>         self.hello_sig = self.fetch_asset(
>             name=hello_sig,
>             asset_hash='f3b9fae20c35740004ae7b8de1301836dab4ac30',
>             locations=hello_sig_locations)
> 
>         mirrors = 'https://mirrors.kernel.org/gnu/hello/'
>         hello = 'hello-2.8.tar.gz'
>         hello_locations = "https://mirrors.kernel.org/gnu/hello/hello-2.8.tar.gz"
>         message = "nothing related to the fetch_asset"
>         hello_sig = 'hello-2.8.tar.gz.sig'
>         hello_sig_locations = "https://mirrors.kernel.org/gnu/hello/hello-2.8.tar.gz.sig"
>         self.hello = self.fetch_asset(
>             name=hello,
>             locations=hello_locations)
>         self.hello_sig = self.fetch_asset(
>             name=hello_sig,
>             asset_hash='f3b9fae20c35740004ae7b8de1301836dab4ac30',
>             locations=hello_sig_locations)
> ```
> 
> Lukas also mentioned supporting lists. When I suggested supporting just pure string parameters was to make a parser that supports a minimal case, other complex cases would be supported into the configuration file, as suggested in the mailing list.
> 
> So, to make discussion shorter, my main concern here is to define how much we want to cover into the parser. With that definition, I can come with a solution that fits it.
> 

I guess I better defined what I think should be the scope: cover the QEMU test case.

> > I guess this leaves you with enough to work on towards a v2. Let me
> > know if I seem to get any of this wrong, or if I can help in any other
> > way.
> 
> Thanks a lot for your suggestions as those will help to come up with a better implementation!

willianrampazzo:
> Maybe I wasn't clear, or just was plain wrong, about what I considered the basic use case. Basically, what I meant was based on:
> 
> * The use case that motivated this feature (the QEMU test)
> * The fact that we should abstain from trying to _execute_ Python code to get the information we need
> 
> That's why I suggested we should stick to string assignments. But, we won't be able to support the basic QEMU tests, unless we support variable assignments (of plain strings).
> 
> > Considering your suggestion, an example I found when I started covering it was the multiple assignment definitions before the `fetch_asset` call without any relation to it or the reuse of the same variable name in the same test or across the tests. The parser needs to find the `fetch_asset` call, check if there are variables at args or right side of keyword args, then walk back in the code looking for those definitions, because we may have multiple definitions of the same variable into the test or across multiple tests, or those variables could be defined as global, outside of the class. Here is an example that covers some of the situations:
> 
> I don't think the parser needs to find the `fetch_asset` call and then walk back. You can parse/store all assignments within the scope of one test. If two assignments for the same variable exist before before a `fetch_asset` call, they should just update the previously found definition (a simple loop over the statements should do just that quite naturally).
> 
> > ```
> > class Hello(Test):
> > 
> >     def test_fetch_hello(self):
> >         mirrors = 'https://mirrors.peers.community/mirrors/gnu/hello/'
> >         hello = 'hello-2.9.tar.gz'
> >         hello_locations = "https://mirrors.peers.community/mirrors/gnu/hello/hello-2.9.tar.gz"
> >         message = "nothing related to the fetch_asset"
> >         hello_sig = 'hello-2.9.tar.gz.sig'
> >         hello_sig_locations = "https://mirrors.peers.community/mirrors/gnu/hello/hello-2.9.tar.gz.sig"
> >         self.hello = self.fetch_asset(
> >             name=hello,
> >             locations=hello_locations)
> > ```
> 
> I see no issues here the statement-by-statement approach. Running this:
> 
> ```python
> import ast
> import pprint
> 
> module = ast.parse(open('hello.py').read(), 'hello.py')
> # first is the import statement
> _, klass = module.body
> test_method = klass.body[0]
> 
> assignments = {}
> 
> for statement in test_method.body:
>     if isinstance(statement, ast.Assign):
>         if isinstance(statement.value, ast.Str):
>             var_name = statement.targets[0].id
>             assignments[var_name] = statement.value.s
>         elif isinstance(statement.value, ast.Call):
>             if statement.value.func.attr == 'fetch_asset':
>                 print('>>> fetch_asset current assigments:')
>                 pprint.pprint(assignments)
> ```
> 
> Gets me:
> 
> ```
> >>> fetch_asset current assigments:
> {'hello': 'hello-2.9.tar.gz',
>  'hello_locations': 'https://mirrors.peers.community/mirrors/gnu/hello/hello-2.9.tar.gz',
>  'hello_sig': 'hello-2.9.tar.gz.sig',
>  'hello_sig_locations': 'https://mirrors.peers.community/mirrors/gnu/hello/hello-2.9.tar.gz.sig',
>  'message': 'nothing related to the fetch_asset',
>  'mirrors': 'https://mirrors.peers.community/mirrors/gnu/hello/'}
> >>> fetch_asset current assigments:
> {'hello': 'hello-2.9.tar.gz',
>  'hello_locations': 'https://mirrors.peers.community/mirrors/gnu/hello/hello-2.9.tar.gz',
>  'hello_sig': 'hello-2.9.tar.gz.sig',
>  'hello_sig_locations': 'https://mirrors.peers.community/mirrors/gnu/hello/hello-2.9.tar.gz.sig',
>  'message': 'nothing related to the fetch_asset',
>  'mirrors': 'https://mirrors.peers.community/mirrors/gnu/hello/'}
> >>> fetch_asset current assigments:
> {'hello': 'hello-2.8.tar.gz',
>  'hello_locations': 'https://mirrors.kernel.org/gnu/hello/hello-2.8.tar.gz',
>  'hello_sig': 'hello-2.8.tar.gz.sig',
>  'hello_sig_locations': 'https://mirrors.kernel.org/gnu/hello/hello-2.8.tar.gz.sig',
>  'message': 'nothing related to the fetch_asset',
>  'mirrors': 'https://mirrors.kernel.org/gnu/hello/'}
> >>> fetch_asset current assigments:
> {'hello': 'hello-2.8.tar.gz',
>  'hello_locations': 'https://mirrors.kernel.org/gnu/hello/hello-2.8.tar.gz',
>  'hello_sig': 'hello-2.8.tar.gz.sig',
>  'hello_sig_locations': 'https://mirrors.kernel.org/gnu/hello/hello-2.8.tar.gz.sig',
>  'message': 'nothing related to the fetch_asset',
>  'mirrors': 'https://mirrors.kernel.org/gnu/hello/'}
> ```
> 
> Which seems to match the code. We may abort if we find anything that is not a string for the first implementation, but that would still solve the basic real QEMU case.
> 
> > ```
> >     self.hello_sig = self.fetch_asset(
> >         name=hello_sig,
> >         asset_hash='f3b9fae20c35740004ae7b8de1301836dab4ac30',
> >         locations=hello_sig_locations)
> > 
> >     mirrors = 'https://mirrors.kernel.org/gnu/hello/'
> >     hello = 'hello-2.8.tar.gz'
> >     hello_locations = "https://mirrors.kernel.org/gnu/hello/hello-2.8.tar.gz"
> >     message = "nothing related to the fetch_asset"
> >     hello_sig = 'hello-2.8.tar.gz.sig'
> >     hello_sig_locations = "https://mirrors.kernel.org/gnu/hello/hello-2.8.tar.gz.sig"
> >     self.hello = self.fetch_asset(
> >         name=hello,
> >         locations=hello_locations)
> >     self.hello_sig = self.fetch_asset(
> >         name=hello_sig,
> >         asset_hash='f3b9fae20c35740004ae7b8de1301836dab4ac30',
> >         locations=hello_sig_locations)
> > ```
> > 
> > 
> > ```
> > 
> > Lukas also mentioned supporting lists. When I suggested supporting just pure string parameters was to make a parser that supports a minimal case, other complex cases would be supported into the configuration file, as suggested in the mailing list.
> > 
> > So, to make discussion shorter, my main concern here is to define how much we want to cover into the parser. With that definition, I can come with a solution that fits it.
> > ```
> 
> I guess I better defined what I think should be the scope: cover the QEMU test case.

I think I got the point here. I'll work on v2 and we can improve it later.

avocado/plugins/assets.py
ldoktor: Well it's slower :-)

avocado/plugins/assets.py
ldoktor: Well you won't lose the logging messages by directly calling the `asset.Asset`. Even better you can tweak the verbosity to show more or less info...

avocado/plugins/assets.py
ldoktor: Sure, in ideal world yes, but consider hunting down a "typo"... We already use `if/elif/...`, let's add the 2 lines of `else: raise ...` with really simple error message as in 99.99% cases it won't hit the user. But it might...

avocado/plugins/assets.py
ldoktor: Well it's the "fetcher" that makes me confused. I'd understand calling it "fetch" as that is an action. But "fetcher" is not an action, therefor I'd expect some generator or class factory in return... Anyway as I mentioned, I am not native so I can get wrong ideas...

avocado/plugins/assets.py
ldoktor: Btw this is my strongest point here, I don't like the idea of generating a test just to fetch the assets. I mean I won't block this if someone else supports it, but using API directly is IMO much nicer.

avocado/plugins/assets.py
willianrampazzo: Just a heads up, I added the `else` statement in the v2 implementation, but the parser handles unknown subcommands into the `config` method, so the `else` won't make any effect here.

willianrampazzo:
Closing to continue the discussion on v2.

willianrampazzo:

PR #3334, opened 2019-10-04
Adding function to create ifcfg- file for IPV6 configuration.

vaishnavibhat:
function to vreate configuration file for IPV6 address.

### DIFF
--- a/avocado/utils/configure_network.py
+++ b/avocado/utils/configure_network.py
@@ -48,7 +48,10 @@ def set_ip(ipaddr, netmask, interface):
             network_conf.write("DEVICE=%s \n" % interface)
             network_conf.write("ONBOOT=yes \n")
             network_conf.write("IPADDR=%s \n" % ipaddr)
-            network_conf.write("NETMASK=%s" % netmask)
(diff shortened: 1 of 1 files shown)


avocado/utils/configure_network.py
```diff
@@ -75,6 +75,53 @@ def set_ip(ipaddr, netmask, interface):
             raise NWException("ifup fails: %s" % ex)
 
 
+def set_ipv6(ipaddr, netmask, interface):
```
PraveenPenguin: can this be achieved in `set_ip` module as well having an extra argument where it decide as it is ipv4 or ipv6

vaishnavibhat: Resending the patch by enabling ipv6 in the existing function  set_ip(ipaddr, netmask, interface)
IPV6 is enabled in all the cases when this utility is being called. Test can proceed when both ipv4 and ipv6 address are assigned, so following this standard.

PraveenPenguin:
@vaishnavibhat can you please add @manvanthar suggestion from #3361 and also resolve Travis failure

PR #3337, opened 2019-10-10
Add initial support to 'assets fetch' command with a simple parser [v2.1]

willianrampazzo:
Implements 'assets fetch' command that uses a parser on instrumented test source to find 'fetch_asset' calls composed of simple strings as parameters, or at least one level of a variable in the same context
with a string assignment, and fetch those assets without running the test.

---

Changes from v1 (#3333)

- Executing asset.Asset instead of an avocado test.
- Handling one level of variable string assignment.
- Fix lots of pylint messages.
- Changed the way the call is built to fit asset.Asset call.

ps. fixed minor issue from v2. Still missing functional/unit tests as a v3 is expected.

### DIFF
--- a/avocado/plugins/assets.py
+++ b/avocado/plugins/assets.py
@@ -1,0 +1,245 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: Red Hat Inc. 2019
+# Authors: Willian Rampazzo <willianr@redhat.com>
+
+"""
+Assets subcommand
+"""
+
+import ast
+import os
+
+from avocado.core import data_dir
+from avocado.core import safeloader
+
+from avocado.core.output import LOG_UI
+from avocado.core.plugin_interfaces import CLICmd
+
+from avocado.utils import asset
+from avocado.utils import data_structures
+
+
+class FetchAssetHandler(ast.NodeVisitor):  # pylint: disable=R0902
+    """
+    Handles the parse of instrumented tests for `fetch_asset` statements
+    and fetch it.
+    """
+
+    def __init__(self, file_name):
+        self.file_name = file_name
+        self.pattern = 'fetch_asset'
+        self.asgmts = {}
+        self.calls = []
+
+        # hold current class and current method to make sure we have the
+        # correct context for the assignment statement.
+        # at this time, module constants and class attributes are discarded.
+        self.current_klass = None
+        self.current_method = None
+
+        # check if we have valid instrumented tests
+        # discards disabled tests
+        self.tests = safeloader.find_avocado_tests(self.file_name)[0]
+
+        # create Abstract Syntax Tree from test source file
+        with open(self.file_name) as source_file:
+            self.tree = ast.parse(source_file.read(), self.file_name)
+
+        # build list of keyword arguments from calls that match pattern
+        self.visit(self.tree)
+    # __init__()
+
+    def _parse_args(self, node):
+        args = []
+        # variables to make lines shorter
+        cur_klass = self.current_klass
+        cur_method = self.current_method
+
+        # parse args from call
+        for arg in node.args:
+            # handle string args
+            if isinstance(arg, ast.Str):
+                args.append(arg.s)
+            # handle variable args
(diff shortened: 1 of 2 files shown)


willianrampazzo:

avocado/plugins/assets.py
```diff
@@ -0,0 +1,245 @@
...
+"""
+Assets subcommand
+"""
+
+import ast
+import os
+
+from avocado.core import data_dir
+from avocado.core import safeloader
+
```
clebergnu: Nitpick: the predominant style is to not have spaces between blocks importing from the same "namespace", that is, among all `avocado.*` modules.

avocado/plugins/assets.py
```diff
@@ -0,0 +1,245 @@
...
+from avocado.core.output import LOG_UI
+from avocado.core.plugin_interfaces import CLICmd
+
+from avocado.utils import asset
+from avocado.utils import data_structures
+
+
+class FetchAssetHandler(ast.NodeVisitor):  # pylint: disable=R0902
+    """
+    Handles the parse of instrumented tests for `fetch_asset` statements
```
clebergnu: I believe this should read "Handles the parsing ..." instead.

avocado/plugins/assets.py
```diff
@@ -0,0 +1,245 @@
...
+        fetch_subcommand_parser = subcommands.add_parser(
+            'fetch',
+            help='Fetch assets from test source or config file if it\'s not'
+            ' already in the cache')
+        fetch_subcommand_parser.add_argument('references', nargs='+',
+                                             metavar='TEST_REFERENCE',
+                                             help='List of test references')
+    # configure()
+
+    def run(self, config):
```
clebergnu: It's important for this command to return a meaningful exit code to its callers.  I see this command being used before the `avocado run` invocation, in which a failure to fetch the assets would not attempt to run the tests (for minimum changes of false negatives due to network issues, download timeouts, etc).

Take a look at `avocado.core.exit_codes` for some documented exit statuses.

avocado/plugins/assets.py
```diff
@@ -0,0 +1,245 @@
...
+    Implements the avocado 'assets' subcommand
+    """
+    name = 'assets'
+    description = 'Manage assets'
+
+    def configure(self, parser):
+        """
+        Add the subparser for the assets action.
+
+        :param parser: Main test runner parser.
```
clebergnu: I understand this docstring is in many other plugins, but it's actually a bit misleading.  The parser is the `avocado` command line application parser, which is a lot more than a "test runner", but the main entry point into most of Avocado functionality.  I'd go with:

```
:param parser: The Avocado command line application parser
:type parser: :class:`avocado.core.parser.ArgumentParser`
```

See https://github.com/avocado-framework/avocado/pull/3339 for the resolution on the other occurrences.

avocado/plugins/assets.py
```diff
@@ -0,0 +1,245 @@
...
+from avocado.core import safeloader
+
+from avocado.core.output import LOG_UI
+from avocado.core.plugin_interfaces import CLICmd
+
+from avocado.utils import asset
+from avocado.utils import data_structures
+
+
+class FetchAssetHandler(ast.NodeVisitor):  # pylint: disable=R0902
```
clebergnu: This class is a node visitor, but you're mixing it up with, what I consider to be, unrelated code that actually fetches the asset.  IMO, the two deserve to be separated, and `fetch_assets()` may be found to be enough to be a function.

This is a design decision, so make your best judgement about this comment.

avocado/plugins/assets.py
```diff
@@ -0,0 +1,245 @@
...
+        # discards disabled tests
+        self.tests = safeloader.find_avocado_tests(self.file_name)[0]
+
+        # create Abstract Syntax Tree from test source file
+        with open(self.file_name) as source_file:
+            self.tree = ast.parse(source_file.read(), self.file_name)
+
+        # build list of keyword arguments from calls that match pattern
+        self.visit(self.tree)
+    # __init__()
```
clebergnu: The closing of methods/classes with comments is not present anywhere else in Avocado.  It's not wrong, but it's a good idea to conform to the project style as much as possible.  In short, I'd drop these.

avocado/plugins/assets.py
```diff
@@ -0,0 +1,245 @@
...
+                # look for assignments at method
+                if arg.id in self.asgmts[cur_klass][cur_method]:
+                    args.append(self.asgmts[cur_klass][cur_method][arg.id])
+                # right now we support just one level of variable as argument,
+                # with a pure string assignment, in the same context,
+                # just like `name = 'file.zip'`
+                else:
+                    return None
+
+        # start building the keywords dictionary for the call
```
clebergnu: You may want to explain that "the call" is `avocado.utils.asset.Asset()` (right?), which, is actually a class constructor.

Issue #3338, opened 2019-10-11
process.SubProcess.send_signal: fail to send signal in sudo mode


lolyu:
```python
# trap.py 
import signal
import sys
import time


def signal_handler(sig, frame):
    print("Receive signal: ", sig)
    sys.exit(0)


signal.signal(signal.SIGUSR1, signal_handler)


while True:
    time.sleep(0.1)
```
```
# python
Python 3.6.8 (default, Jul  3 2019, 10:59:07) 
[GCC 8.3.1 20190507 (Red Hat 8.3.1-4)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from avocado.utils import astring
>>> from avocado.utils import process
>>> import signal
>>> cmd = "python trap.py"
>>> pp = process.SubProcess(cmd, verbose=True, allow_output_check=None, shell=False, env=None, sudo=False, ignore_bg_processes=False, encoding=astring.ENCODING)
>>> pp.send_signal(signal.SIGUSR1)
>>> pp.get_stdout()
b''
>>> pp._popen.send_signal(signal.SIGUSR1)
>>> pp.get_stdout()
b'Receive signal:  10\n'
```
Once in `root` mode, `process.SubProcess.send_signal` fails to signal the child process. Direct calling  of `subprocess.popen.send_signal` works as expected.

lolyu:
I will try to fix this.

lolyu:
thanks all.

PR #3341, opened 2019-10-15
process.SubProcess.send_signal: fix the failure to signal itself

lolyu:
`send_signal` fails to send signal to itself in sudo mode, which only
send signals to its child processes.


lolyu:
#https://github.com/avocado-framework/avocado/issues/3338

PR #3345, opened 2019-10-22
Add initial support to 'assets fetch' command with a simple parser [v4]

willianrampazzo:
Implements 'assets fetch' command that uses a parser on instrumented test source to find 'fetch_asset' calls composed of simple strings as parameters, or at least one level of variable in the same context
with a string assignment, and fetch those assets without running the test.

---

Changes from v1 (#3333)

- Executing asset.Asset instead of an avocado test.
- Handling one level of variable string assignment.
- Fix lots of pylint messages.
- Changed the way the call is built to fit asset.Asset call.

Changes from v2 (#3337)

- Fix bug that incorrectly considered functions outside of the classes.
- Add command return codes.
- Move `fetch_assets` out of the node visitor class.
- Add functional tests.
- Add some unit tests. Some are still missing.

Changes from v3 (#3344)

- Create a new functional test class to allow the correct use of setup for the expected file in the test.
- Move `pattern` to class attribute.
- Relax return code when no supported test was found in the file.
- Improve help message.

### DIFF
--- a/avocado/plugins/assets.py
+++ b/avocado/plugins/assets.py
@@ -1,0 +1,259 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


willianrampazzo:

PR #3358, opened 2019-11-07
Implements FetchAssetJob plugin to fetch assets during JobPreTests [v2]

willianrampazzo:
This implementation adds the functionality of assets fetch command during JobPreTests. This way, all assets are fetched prior to the tests starts. If the asset fails to be downloaded, it will try again during the test execution. This is a note for the future implementation of a decorator that skips the test if an asset is not available.
    
---

Changes from v1:
- improved granularity of fetch assets functionality to be able to look at specific tests.
- used information from job.test_suite instead of config.

### DIFF
--- a/avocado/plugins/assets.py
+++ b/avocado/plugins/assets.py
@@ -28,15 +28,20 @@ from avocado.utils.asset import Asset
 from avocado.utils import data_structures
 
 
-class FetchAssetHandler(ast.NodeVisitor):
+class FetchAssetHandler(ast.NodeVisitor):  # pylint: disable=R0902
     """
     Handles the parsing of instrumented tests for `fetch_asset` statements.
     """
 
     PATTERN = 'fetch_asset'
 
-    def __init__(self, file_name):
+    def __init__(self, file_name, klass=None, method=None):
         self.file_name = file_name
+        # fetch assets from specific test using klass and method
+        self.klass = klass
+        # we need to make sure we cover the setUp method when fetching
+        # assets for a specific test
+        self.method = [method, 'setUp']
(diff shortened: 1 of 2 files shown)


willianrampazzo:
[Extend Assets plugin to run the fetch during JobPreTests](https://trello.com/c/jcqChIqN/1531-extend-assets-plugin-to-run-the-fetch-during-jobpretests)

avocado/plugins/assets.py
```diff
@@ -24,19 +24,25 @@
...
-class FetchAssetHandler(ast.NodeVisitor):
+class FetchAssetHandler(ast.NodeVisitor):  # pylint: disable=R0902
     """
     Handles the parsing of instrumented tests for `fetch_asset` statements.
     """
 
     PATTERN = 'fetch_asset'
 
-    def __init__(self, file_name):
+    def __init__(self, file_name, klass=None, method=None):
```
beraldoleal: I would suggest to follow PEP8:

"If a function argument's name clashes with a reserved keyword, it is generally better to append a single trailing underscore rather than use an abbreviation or spelling corruption. Thus class_ is better than clss. (Perhaps better is to avoid such clashes by using a synonym.)"

clebergnu: No, `klass` is really a subterfuge for `class`.

We have a lot of those uses in Avocado (144), and I'm probably the one to blame for most if not all of them.  I think it'd be quite hard to have a pylint check for that, so the best we can do is  a mass change, probably in its own commit.

willianrampazzo:
Thanks for reviewing @clebergnu, @beraldoleal!

PR #3359, opened 2019-11-08
Implements FetchAssetJob plugin to fetch assets during JobPreTests [v3]

willianrampazzo:
This implementation adds the functionality of assets fetch command during JobPreTests. This way, all assets are fetched prior to the tests starts. If the asset fails to be downloaded, it will try again during the test execution. This is a note for the future implementation of a decorator that skips the test if an asset is not available.

Signed-off-by: Willian Rampazzo willianr@redhat.com

---
Changes from v1:
- improved granularity of fetch assets functionality to be able to look at specific tests.
- used information from job.test_suite instead of config.

Changes from v2:
- improved user output message.
- adjusted one functional test with the new output.

### DIFF
--- a/avocado/plugins/assets.py
+++ b/avocado/plugins/assets.py
@@ -28,15 +28,20 @@ from avocado.utils.asset import Asset
 from avocado.utils import data_structures
 
 
-class FetchAssetHandler(ast.NodeVisitor):
+class FetchAssetHandler(ast.NodeVisitor):  # pylint: disable=R0902
     """
     Handles the parsing of instrumented tests for `fetch_asset` statements.
     """
 
     PATTERN = 'fetch_asset'
 
-    def __init__(self, file_name):
+    def __init__(self, file_name, klass=None, method=None):
         self.file_name = file_name
+        # fetch assets from specific test using klass and method
+        self.klass = klass
+        # we need to make sure we cover the setUp method when fetching
+        # assets for a specific test
+        self.method = [method, 'setUp']
         self.asgmts = {}
         self.calls = []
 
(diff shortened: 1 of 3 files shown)


willianrampazzo:
[Extend Assets plugin to run the fetch during JobPreTests](https://trello.com/c/jcqChIqN/1531-extend-assets-plugin-to-run-the-fetch-during-jobpretests)

avocado/plugins/assets.py
```diff
@@ -203,6 +215,32 @@ def fetch_assets(test_file):
...
+    """
+    Implements the assets fetch job pre tests. This has the same effect of
+    running the 'avocado assets fetch INSTRUMENTED', but it runs during the
+    test execution, before the actual test starts.
+    """
+    name = "fetchasset"
+    description = "Fetch assets before the test run"
+
+    def __init__(self, config=None):
+        pass
```
beraldoleal: `config` is been passed but not used. Is that expected?

willianrampazzo: This is a workaround as `avocado/core/extension_manager.py`, line 63, `obj = plugin(**invoke_kwds)`, calls all the plugins with a parameter. As JobPreTests class do not define any, we need to define at least one to be able to run it, or we will have an error message like this:

```
Failed to load plugin from module "avocado.plugins.assets": TypeError('FetchAssetJob() takes no arguments') :
  File "~/src/avocado/avocado.dev/avocado/core/extension_manager.py", line 63, in __init__
    obj = plugin(**invoke_kwds)
```

willianrampazzo:
> LGTM. Upon merge, I'd squash the two last commits because it'd prevent problems when bisecting regressions (that is, the second commit introduces a test failure, while the later third fixes it).

Makes sense, thanks @clebergnu!

avocado/plugins/assets.py
beraldoleal: Ok, let's move on. :)

beraldoleal:
@clebergnu yes, he did. We also talked on IRC about this.

clebergnu:
Merged a version of the last two commits squashed as explained before.  Thanks!

PR #3383, opened 2019-11-27
Add interface_link_up function to check if the link is

vaishnavibhat:
up or down using ethtool.

This utility can be used in other network tests to check if the
link of an interface is up or not.

### DIFF
--- a/avocado/utils/configure_network.py
+++ b/avocado/utils/configure_network.py
@@ -23,6 +23,7 @@ import os
 import logging
 from . import distro
 from . import process
+from . import genio
 from .ssh import Session
 
 log = logging.getLogger('avocado.test')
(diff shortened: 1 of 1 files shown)


avocado/utils/configure_network.py
```diff
@@ -191,3 +191,15 @@ def get_peer_interface(self, peer_ip):
...
         else:
             return peer_interface
+
+
+def interface_link_up(interface):
+    '''
+    Checks if the interface link is up
+    '''
+    if 'UP' or 'yes' in \
+        process.system_output("ethtool %s | tail -1"
```
PraveenPenguin: can we depend on sysfs attibute example can be 

`if up in '/sys/class/net/%s/operstate' %% interface`

avocado/utils/configure_network.py
```diff
@@ -191,3 +191,15 @@ def get_peer_interface(self, peer_ip):
...
+                              % interface, shell=True, ignore_status=True):
+        print("Interface %s link is up" % interface)
```
PraveenPenguin: it can be part of logger or it not needed as library should not print value

avocado/utils/configure_network.py
```diff
@@ -191,3 +191,15 @@ def get_peer_interface(self, peer_ip):
                 log.error("unable to get peer interface: %s", ex)
         else:
             return peer_interface
+
+
+def interface_link_up(interface):
```
beraldoleal: If possible, I would suggest using a different method name, ie:

`is_interface_link_up`, `is_interface_up`, `is_link_up` or similar.

narasimhan-v:
@PraveenPenguin Any help on this please ?

PR #3387, opened 2019-12-03
Removing self from ping_check and set_mtu_host

smruti77:
removed self form ping_check and set_mtu_host function.


PraveenPenguin:
@smruti77 thanks for the patch , @beraldoleal thanks for ACK

PR #3388, opened 2019-12-04
Add an initial pylintrc file to be used with IDEs

willianrampazzo:
While developing, some IDEs will report lint problems that are not covered by selftests/inspekt-lint.sh. The developer may disable the message in the source code without knowing it is not covered by the
lint check. The pylintrc can be used to avoid this kind of mismatch. The messages enabled/disabled in the initial file are those listed on selftests/inspekt-lint.sh.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,497 @@
+[MASTER]
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code.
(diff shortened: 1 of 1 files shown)


willianrampazzo:
> My "IDE" doesn't know how to use a `.pylintrc` file, but I trust this adds value to other users (you included of course).

I use vim with some steroids from vim-bootstrap (https://vim-bootstrap.com/) :D ... don't know if it can be called an IDE.

PR #3397, opened 2019-12-12
docs/blueprints: First blueprint: Configuration by convention

beraldoleal:
This is the outcome of the discussions on the RFC "Configuration by
convention". It is still a draft and after revised and approved we can
start working on the specific issues.


beraldoleal:
> Hi @beraldoleal, it is nice to see the entire blueprint available here! Can you please clarify what is the next step in this? I mean granted, we can just approve it (as far as I can see it follows our full discussions in the mailing list and was updated to reflect all recent comments) but how should we proceed about implementing these guidelines? Perhaps gradually or in some large pass PR-s to at least cover all core plugins? Then I assume other plugins should follow the standards set here as well. Am I correct in my assumptions on the anticipated workflow?

Hi @pevogam yes, I think that you are.

And since that the blueprint is still a draft, could be merged anytime soon and changes can be still merged after that. The idea was to not implement any "drastic change" until we have the status changed to "approved". 

IMO, the next steps should be: classify the issues by priority and create all issues on GH/Trello so we can split the work.

I already have started looking for a few issues, but I will make clear my assignments when these issues become a GH/Trello issue. (after the BP approval).

When all issues are closed then this "epic" issue (this blueprint) will be considered "implemented/done". Then we can try to rewrite this blueprint as documentation to be incorporated in our manuals.

I'm still waiting for a go-ahead here, but since that next week will be our release, perhaps this will wait until next year, not sure.

beraldoleal:
Fixed on v2.

PR #3412, opened 2019-12-20
Refactor network utility

willianrampazzo:


willianrampazzo:
This came with the code of my previous pull request #3411. I will fix it and resend as they are not related.

PR #3495, opened 2020-01-13
modifying device_exists function in multipath.py utils

Naresh-ibm:
multipath -ll command used to exit with TRUE condition in previos logic
all the tim. Hence change it properly with the help of if condition which
checks for real existance of mpath.
Now it is working fine.


PraveenPenguin:
@beraldoleal thanks for ACK

PR #3499, opened 2020-01-15
Readd build and quality badges on README.

willianrampazzo:
Reintroduce the build and quality badges to the README file.


willianrampazzo:
> Why aren't the Cirrus, codecov and Documentation (readthedocs.org) indicators also being reintroduced?
> 

I took the code from a previous commit that was missing those badges. I readded them.

> Also, a matter of style/preference: wouldn't it be better to have those at the end, so that they don't confuse users as the very first thing they see _and_ read when looking at the unrendered source?

I followed what I'm used to seeing at GitHub. Projects usually add the badges at the beginning. Anyway, I moved them to the end now.

I pushed a new commit with the changes. Do you need a v2 instead?

willianrampazzo:
Closing to submit a fixed v2.

PR #3507, opened 2020-01-17
selftests/deployment: Disable fedora-modular repos [v2]

beraldoleal:
When you have Fedora modules enabled, Avocado package will be installed
from modules instead of the ones in the COPR or Official. This change
will disable fedora-modular and fedora-updates-modular repos during this
deployment.

----
Changes from v1:

  * Converted int to str


beraldoleal:
Related to #3505

PR #3522, opened 2020-01-21
small fix: removed duplicated strings on warning messages

beraldoleal:
Removed some duplicated strings on warning messages.

### DIFF
--- a/avocado/plugins/run.py
+++ b/avocado/plugins/run.py
@@ -209,8 +209,8 @@ class Run(CLICmd):
                                                           None)
 
         warnings.warn("The following arguments will be changed to boolean soon: "
-                      "sysinfo, output-check, failfast, keep-tmp, "
-                      "ignore-missing-references, sysinfo and output-check",
+                      "sysinfo, output-check, failfast, keep-tmp "
+                      "and ignore-missing-references.",
                       FutureWarning)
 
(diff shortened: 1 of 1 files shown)


wainersm:
@beraldoleal this patch looks good to me!

Just one more comment related with those parameters.... on `avocado run --help` it says, for example:

```
  --failfast {on,off}   Enable or disable the job interruption on first failed
                        test. 'on' and 'off' will be deprecated soon.
```

Suppose I am about to use `--failfast`. At this point should I use `--failfast on` or can I use the new value? and what is the new value by the way? 

So perhaps it can mention that 'on' should be replaced with 'true' (likewise 'off'  with 'false'). And say this change is effectively in place (is it in place?).

beraldoleal:
@wainersm at this point the new values are not implemented yet. You still should use `--failfast on`.

The new, as described on RFC, for boolean values, are going to be without args, like that: `--failfast` if you would like to activate failfast.

wainersm:
> @wainersm at this point the new values are not implemented yet. You still should use `--failfast on`.
> 

ah, ok.

> The new, as described on RFC, for boolean values, are going to be without args, like that: `--failfast` if you would like to activate failfast.

Makes sense, `--failfast true` is redundant after all.

PR #3534, opened 2020-01-24
tests/nrunner: removed unnecessary double quotes

beraldoleal:
'"command"' is not a valid sequence of arguments (from the Python interpreter
perspective).


beraldoleal:
Maybe it is related to #3492

PR #3538, opened 2020-01-28
added sleep to the mtu_set_host() and mtu_set_peer()

smruti77:
added sleep to the mtu_set().because its taking time for physical adapter.

### DIFF
--- a/avocado/utils/configure_network.py
+++ b/avocado/utils/configure_network.py
@@ -19,6 +19,7 @@ Configure network when interface name and interface IP is available.
 
 import shutil
 import os
+import time
 
 import logging
 from . import distro
@@ -135,6 +136,7 @@ def set_mtu_host(interface, mtu):
         cmd = "ip add show %s" % interface
         mtuvalue = process.system_output(cmd, shell=True).decode("utf-8") \
                                                          .split()[4]
+        time.sleep(10)
         if mtuvalue == mtu:
             return True
     except Exception as ex:  # pylint: disable=W0703
@@ -167,6 +169,7 @@ class PeerInfo:
             self.session.cmd(cmd)
             cmd = "ip add show %s" % peer_interface
             mtuvalue = self.session.cmd(cmd).stdout.decode("utf-8").split()[4]
+            time.sleep(10)
             if mtuvalue == mtu:
                 return True
         except Exception as ex:  # pylint: disable=W0703


avocado/utils/configure_network.py
```diff
@@ -135,6 +136,7 @@ def set_mtu_host(interface, mtu):
         cmd = "ip add show %s" % interface
         mtuvalue = process.system_output(cmd, shell=True).decode("utf-8") \
                                                          .split()[4]
+        time.sleep(10)
```
beraldoleal: Hi @smruti77, I'm not so familiar with this module but I will try to give my two cents here.

I might be missing something, but shouldn't the `time.sleep()` be between the two commands?

Also, looks like 10 seconds is a lot just to set the MTU value. Could you try to explain in which situation this is taking so long? Specific arch? Is this valid for all scenarios?

For instance, on my machine, I don't have to wait for 10 seconds to set the MTU:

```
# time ip l set enp0s31f6 mtu 1500

real	0m0.004s
user	0m0.001s
sys	0m0.003s

```

So, maybe add this as a parameter of the method with a low default value.

Just a last comment: Have you tried to set this directly to `/sys/class/`? It is better?

```
# time echo 1500 > /sys/class/net/enp0s31f6/mtu 

real	0m0.000s
user	0m0.000s
sys	0m0.000s
```

smruti77: @beraldoleal not in all the card its taking time to set mtu,for physical card card its taking few sec to come up  interface while set mtu.i tried with for chceking interface up or down after set mtu but still its gave few percentage of packet loss. so i didn't get any option to check after mtu set so put sleep().

beraldoleal: @smruti77 ok, but did you tried to set the MTU with a different method? Is it the same?

Just trying to understand here, could you please execute the following script and paste the result?

```bash
#!/bin/bash
N=`date +%s%N`
export PS4='+[$(((`date +%s%N`-$N)/1000000))ms][${BASH_SOURCE}:${LINENO}]: ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
set -x

INTERFACE=enp0s31f6
MTU=5500

ip l set $INTERFACE mtu $MTU
ip a show $INTERFACE
```

beraldoleal: @smruti77 ok, but did you tried to set the MTU with a different method? Is it the same?

Just trying to understand here, could you please execute the following script and paste the result?

```bash
#!/bin/bash
N=`date +%s%N`
export PS4='+[$(((`date +%s%N`-$N)/1000000))ms][${BASH_SOURCE}:${LINENO}]: ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
set -x

INTERFACE=enp0s31f6
MTU=5500

ip l set $INTERFACE mtu $MTU
ip a show $INTERFACE
```

smruti77: +[1ms][./test.sh:6]: main(): INTERFACE=eth1
+[2ms][./test.sh:7]: main(): MTU=6000
+[4ms][./test.sh:9]: main(): ip l set eth1 mtu 6000
+[1222ms][./test.sh:10]: main(): ip a show eth1
21: eth1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 6000 qdisc mq state UP group default qlen 1000
    link/ether 40:f2:e9:31:27:00 brd ff:ff:ff:ff:ff:ff
    inet 178.1.1.192/24 brd 178.1.1.255 scope global eth1
       valid_lft forever preferred_lft forever
    inet6 fe80::42f2:e9ff:fe31:2700/64 scope link 
       valid_lft forever preferred_lft forever

beraldoleal: Thank you @smruti77 .

My only concern here is to add this sleep for all the cases. When you are on a system that changing the MTU it is a quick task, this should not take 10 seconds.

So my suggestion would be to add an additional method argument, with a low default value. But I'm not sure if this is ok.

@clebergnu do you have any thoughts here?

willianrampazzo: An active wait would reduce the wait damage.

smruti77: tried with wait also its not working fine.

beraldoleal: @smruti77, please, can you give us more details of why it did not work?

smruti77: @beraldoleal  not in all adapter we are facing this issue.in some adapter its taking time after mtu set its not pinging.after few sec its pinging. tried with is_interface_link_up() check still its failing. so i thought to put sleep.

beraldoleal: @smruti77 I understood that. But we are trying to find another solution here, because time.sleep(10) is to heavy for some cases.

So, my question is: Did you tried with the "timeout"? and return if the timeout is reached even that without the value? In that case, users with interfaces that setting MTU is a fast process will not be impacted. If you tried already, why this did not work? Can you give-us a code example and the output of error?

smruti77: @beraldoleal timeout also not working.

smruti77:
new PR for this #3594

Issue #3450, opened 2020-01-07
Asset management: refactor Asset class to allow extensibility


willianrampazzo:
The objective of this issue is to track the Asset class refactoring, making the Asset class simple and improving the flexibility for future extensions.

willianrampazzo:
I have completed most of the tasks I had for the Asset class. I'm closing this issue for now. Other improvements will be handled by new issues.

PR #3550, opened 2020-01-31
Assets refactor round2

willianrampazzo:
This is the second round of the Assets class refactor based on comments made at #3411 

-  Move test_utils_assets.py to the correct selftests directory;
- Refactor _verify method;
- Move the json calls inside the `with` statements;
- Change EnvironmentError exception to OSError.


willianrampazzo:
This PR is related to #3450.

avocado/utils/asset.py
```diff
@@ -285,20 +284,17 @@ def _verify(self, asset_path):
         value as the hash of the asset_file, otherwise return False.
         :rtype: bool
         """
-        if not self.asset_hash:
```
apahim: This test should be `if self.asset_hash is None`. I think it was like that... not sure why it changed.

avocado/utils/asset.py
```diff
@@ -285,20 +284,17 @@ def _verify(self, asset_path):
...
-            return True
-        if self._get_hash_from_file(asset_path) == self.asset_hash:
-            return True
-        else:
-            return False
+        return bool(not self.asset_hash or
+                    (self.asset_hash and
+                     self._get_hash_from_file(asset_path) == self.asset_hash))
```
apahim: Why is this change needed? I mean, this is way less readable and the second `self.asset_hash` check inside the `boll()` is redundant. I'd just remove the else (while keeping the final return), but if you insist in changing this, you should consider using [any()](https://docs.python.org/3.6/library/functions.html#any).

avocado/utils/asset.py
willianrampazzo: Okay, I'll revert on v2.

avocado/utils/asset.py
willianrampazzo: Okay, I'll just remove the `else` on v2.

willianrampazzo:
Sending v2.

PR #3555, opened 2020-02-03
Assets refactor round2 [v2]

willianrampazzo:
This is the second round of the Assets class refactor based on comments made at #3411

Move test_utils_assets.py to the correct selftests directory;
Refactor _verify method;
Move the json calls inside the with statements;
Change EnvironmentError exception to OSError.


willianrampazzo:
This PR is related to #3450.

avocado/utils/asset.py
```diff
@@ -285,12 +285,10 @@ def _verify(self, asset_path):
         value as the hash of the asset_file, otherwise return False.
         :rtype: bool
         """
-        if not self.asset_hash:
+        if self.asset_hash is None or (
+                self._get_hash_from_file(asset_path) == self.asset_hash):
             return True
-        if self._get_hash_from_file(asset_path) == self.asset_hash:
```
apahim: I still don't get it. Why are you touching this?

willianrampazzo: It has the same effect, but I can revert, no problem.

apahim: To be clear, I think there are two options here:
- Option 1 (more readable): fix the first check and remove the else:
```python
    def _verify(self, asset_path):
        if self.asset_hash is None:
            return True
        if self._get_hash_from_file(asset_path) == self.asset_hash:
            return True
        return False
```
- Option 2 (a bit faster): refactor to one expression to rule them all:
```python
    def _verify(self, asset_path):
        return (self.asset_hash is None or
                self._get_hash_from_file(asset_path) == self.asset_hash)
```

willianrampazzo: I may have missed something here, but what is the difference between option 1 and the current code in this PR?

apahim: Readability & not touching the code without a reason. But maybe I don't know the reason... what are you trying to achieve?

willianrampazzo: IMO, a short-circuit statement, is more elegant than the current code, but that may not be a good reason for touching the code. I'll send a v3 without this change, or with option 2 that you suggested, what do you think?

apahim: Well, "elegancy" might be a good reason, even though we disagree on which one is more :)
I'm ok with keeping the current code in the PR then.
By the way, the `if foo is None` thing is a bigger problem than this method alone... It was changed everywhere in this module and you might want to take the opportunity to revert them all?

willianrampazzo: > By the way, the `if foo is None` thing is a bigger problem than this method alone... It was changed everywhere in this module and you might want to take the opportunity to revert them all?

Yep, will push a new PR with those changes in some minutes. Thanks for pointing that out.

beraldoleal: From Zen of Python (PEP 20):

> Beautiful is better than ugly.
> Explicit is better than implicit.
> Simple is better than complex.
> Complex is better than complicated.
> Flat is better than nested.
> Readability counts.
> If the implementation is hard to explain, it's a bad idea.
> If the implementation is easy to explain, it may be a good idea.

willianrampazzo:
Sending v3.

PR #3562, opened 2020-02-04
docs: Improve contributor usability with some explicit requirements

beraldoleal:
Today, in order to install selfttests-requirements we need make and libvirt-devel. Without this, make requirements-selftests will fail. This will add this instruction to our contributor guide.

### DIFF
--- a/docs/source/guides/contributor/chapters/environment.rst
+++ b/docs/source/guides/contributor/chapters/environment.rst
@@ -5,6 +5,17 @@ Development environment
 
 .. warning:: TODO: Needs improvment here. i.e: virtualenvs, GPG, etc.
 
+Installing dependencies
+-----------------------
+
+You need to install few dependencies before start coding::
+
+ $ sudo dnf install gcc libvirt-devel
+
+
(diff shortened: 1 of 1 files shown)


requirements-selftests.txt
```diff
@@ -12,6 +12,9 @@ funcsigs>=0.4
 aexpect==1.5.1
 psutil==5.4.7
 
+# To test optional_plugins/varianter_yaml_to_mux
+pyaml==19.12.0
```
willianrampazzo: I was testing your patch and noticed that, maybe, it would also be a good idea to change `optional_plugins/varianter_yaml_to_mux/setup.py` with this same package added here.

Or we just add this PR and submit a new one later.

beraldoleal: Hummm nice catch. I will remove this yaml from this PR and open a new issue for a better investigation on pyaml/PyYaml problem.

beraldoleal:
@willianrampazzo I removed the commit about pyaml, it is safer. Let me know what you think and feel free to accept it.

Issue #3473, opened 2020-01-08
Make avocado.skip* usable on classes


willianrampazzo:
`avocado.skip*` decorators can currently only be used on methods (setUp and test methods).  The similar decorators from the `unittest` standard library can also be used on classes.

Let's make `avocado.skip*` feature comparable to the `unittest.skip*` ones.

One example of where this would be used:

```
@skipUnless(ssh.SSH_CLIENT_BINARY, 'No SSH client available')
classs LinuxSSHTest(Test):
```

Instead of (currently required):

```
classs LinuxSSHTest(Test):

    @skipUnless(ssh.SSH_CLIENT_BINARY, 'No SSH client available')
    def setUp(self):
        super(LinuxSSH, self).setUp()
```

[1] - https://lists.gnu.org/archive/html/qemu-devel/2019-09/msg04518.html

PR #3566, opened 2020-02-11
Avocado.skip* usable on classes

richtja:
This PR expands the avocado.skip* decorators from methods to classes.
When a user uses skip decorator on the class, the decorator is applied
to every class method.

Reference: https://github.com/avocado-framework/avocado/issues/3473

### DIFF
--- a/avocado/core/decorators.py
+++ b/avocado/core/decorators.py
@@ -11,9 +11,9 @@
 #
 # Copyright: Red Hat Inc. 2017
 # Author: Amador Pahim <apahim@redhat.com>
-
 from functools import wraps
 import types
+import inspect
 
 from . import exceptions as core_exceptions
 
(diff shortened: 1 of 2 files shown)


avocado/core/decorators.py
```diff
@@ -73,14 +73,28 @@ def skip(message=None):
...
+                    raise core_exceptions.TestSkipError(message)
+                function = wrapper
+            function.__skip_test_decorator__ = True
+            return function
+
+        def class_decorator(cls):
+            for key in cls.__dict__:
+                value = getattr(cls, key)
+                if callable(getattr(cls, key)):
+                    wrapped = method_decorator(value)
```
willianrampazzo: I don't think I understood the need to set `value` before the `if` statement. Can't you do: 

```
if callable(getattr(cls, key)):
    wrapped = method_decorator(getattr(cls, key))
```
or, using the `value` variable:

```
value = getattr(cls, key)
if callable(value):
    wrapped = method_decorator(value)
```
Maybe I'm missing something here :)

avocado/core/decorators.py
```diff
@@ -73,14 +73,28 @@ def skip(message=None):
...
+                    setattr(cls, key, wrapped)
+            return cls
+
+        if inspect.isclass(obj):
```
willianrampazzo: Looking at _inspect_ source (https://github.com/python/cpython/blob/master/Lib/inspect.py#L72), the method _isclass_ does `isinstance(object, type)`. In this case, I don't see a reason to use _inspect_, in favor of `isinstance(obj, type)`.

avocado/core/decorators.py
richtja: Yes, you are right there is no need for setting `value`.

avocado/core/decorators.py
richtja: Ok, I will change that.

richtja:
@willianrampazzo thanks for the review, I will create a new version with fixes.

richtja:
I'm closing because of #3570

PR #3568, opened 2020-02-18
Added PMEM library support

harish-24:
Patch adds PMEM library support to manage persistent memory using ndctl and daxctl binaries. This patch allows managing pmem devices such as creation, deletion, enabling, disabling and re-configuring both namespaces and regions depending on supported hardware.

### DIFF
--- a/avocado/utils/pmem.py
+++ b/avocado/utils/pmem.py
@@ -1,0 +1,303 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright (C) IBM 2020 - Harish <harish@linux.vnet.ibm.com>
+#
+# Author: Harish <harish@linux.vnet.ibm.com>
+#
+
+
+import re
+import json
+import glob
+import logging
+
+from . import process
+from . import genio
+
+log = logging.getLogger('avocado.test')
+
+
+class PMemException(Exception):
+    """
+    Error raised for all PMem failures
+    """
+    def __init__(self, additional_text=None):
+        self.additional_text = additional_text
+
+    def __str__(self):
+        return ("Command failed.\ninfo: %s" % self.additional_text)
+
+
+class PMem:
+    """
+    PMem class which provides function to perform ndctl and daxctl operations
+
+    This class can be used only if ndctl binaries are provided before hand
+    """
+
+    def __init__(self, ndctl='ndctl', daxctl='daxctl'):
+        """
+        Initialize PMem object
+
+        :param ndctl: path to ndctl binary, defaults to ndctl
+        :param daxctl: path to daxctl binary, defaults to ndctl
+        """
+        for lib_bin in [ndctl, daxctl]:
+            if process.system('which %s' % lib_bin, shell=True, ignore_status=True):
+                raise PMemException("Cannot use library without "\
+                                    "proper binary %s" % lib_bin)
+        self.ndctl = ndctl
+        self.daxctl = daxctl
+
+    def run_ndctl_list(self, option=''):
+        """
+        Get the json of each provided options
+
+        :param option: optional arguments to ndctl list command
+        :return: By default returns entire list of json objects
+        :rtype: list of json objects
+        """
+        try:
+            json_op = json.loads(process.system_output(
+                '%s list %s' % (self.ndctl, option), shell=True))
+        except ValueError:
+            json_op = []
+        return json_op
+
+    @staticmethod
+    def run_ndctl_list_val(json_op, field):
+        """
+        Get the value of a field in given json
+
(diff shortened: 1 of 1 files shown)


avocado/utils/pmem.py
```diff
@@ -0,0 +1,298 @@
...
+    def get_slot_count(self, region):
+        """
+        Get max slot count in the index area for a  dimm backing a region
+        We use region0 - > nmem0
+
+        :param region: Region for which slot count is found
+        :return: Number of slots for given region
+        :rtype: int
+        """
+        nmem = "nmem%s" % re.findall(r'\d+', region)[0]
```
PraveenPenguin: what about if  region is not valid library should break can we have that check as well as in this case the library will fail I see try for actual cmd run my point is if re itself give a bogus value, seems me we need to handle that

avocado/utils/pmem.py
```diff
@@ -0,0 +1,298 @@
...
+        try:
+            json_op = json.loads(process.system_output(
+                '%s read-labels -j %s ' % (self.ndctl, nmem), shell=True))
+        except ValueError:
+            json_op = []
+        first_dict = json_op[0]
+        index_dict = self.run_ndctl_list_val(first_dict, 'index')[0]
+        return self.run_ndctl_list_val(index_dict, 'nslot') - 2
```
PraveenPenguin: ok

avocado/utils/pmem.py
```diff
@@ -0,0 +1,298 @@
...
+        """
+        Disable given region
+
+        :param name: name of the region to be disabled
+        :return: True on success
+        :raise: :class:`PMemException`, if command fails.
+        """
+        if process.system('%s disable-region %s' % (self.ndctl, name),
+                          shell=True, ignore_status=True):
+            raise PMemException("Failed to disable %s region(s)" % name)
```
PraveenPenguin: raising `Exception` is fine but library either return some value like in this case True/False as a success it returns true so I feel in Fail condition it should return False as it will easy to process/use this library in the Test program

avocado/utils/pmem.py
```diff
@@ -0,0 +1,298 @@
...
+        """
+        Enable given region
+
+        :param name: name of the region to be enabled
+        :return: True on success
+        :raise: :class:`PMemException`, if command fails.
+        """
+        if process.system('%s enable-region %s' % (self.ndctl, name),
+                          shell=True, ignore_status=True):
+            raise PMemException("Failed to enable %s region(s)" % name)
```
PraveenPenguin: same here

avocado/utils/pmem.py
```diff
@@ -0,0 +1,298 @@
...
+
+
+class PMem:
+    """
+    PMem class which provides function to perform ndctl and daxctl operations
+
+    This class can be used only if ndctl binaries are provided before hand
+    """
+
+    def __init__(self, ndctl='ndctl', daxctl='daxctl'):
```
PraveenPenguin: maybe we need to check if `ndctl` and `daxctl` is present in system as we have default value and user has an impression that it always works .if those libraries not their avocado  library will break feeling  we need to handle that as well, either always take parm through user or check if library path not there return  False and don't create Call object

avocado/utils/pmem.py
harish-24: region being an input to this function, is upto the caller to provide right value. When you say region is not valid you mean wrong input ? The caller always is expected to provide a proper input as the result of the function will subsequently used.

avocado/utils/pmem.py
harish-24: Returning failure can also be used. But the main idea behind raising an user-defined Exception is to easily handle it from the test case. For example if a test uses 4 to 5 functions of the library, instead of a `if` check for each of the function/cmd, isn't it better to handle the exception from the test? And if the functions are used across different tests then these have to be repeated everywhere. Because these libraries are expected to work most often than not.

avocado/utils/pmem.py
harish-24: Sure, makes sense.

avocado/utils/pmem.py
PraveenPenguin: @harish-24 yes i got your idea earlier as well , but library should return True and False that is easy to handel

avocado/utils/pmem.py
PraveenPenguin: yes it wrong region input, yes it should be always a valid one but in wrong input, the library needs to be handled

harish-24: ok.

avocado/utils/pmem.py
harish-24: @PraveenPenguin It could be any of the two solutions. It depends on the library usage. This case is very similar to `process.run` which does something similar. What is your thought on this?

harish-24:
@PraveenPenguin Pushed with changes except for one. Thanks.

avocado/utils/pmem.py
PraveenPenguin: yes but in that case, you need to always use try-expect in test to Handel somehow I did not like that as library always returns some value and in your case, you are returning `True`  in success case , In process API case i relate like a wrapper over python library (os.system) so it seems me ok

PraveenPenguin: anyway this is just a suggestion you take your call

harish-24: Sure, I would like to go with existing approach as it is more convenient from handling in a test and same is included in docstring.

PR #3569, opened 2020-02-18
adding ping6_check function to test ipv6 ping

vaishnavibhat:
ping6 test for ipv6 ping and flood ping

### DIFF
--- a/avocado/utils/configure_network.py
+++ b/avocado/utils/configure_network.py
@@ -189,6 +189,30 @@ class PeerInfo:
         else:
             return peer_interface
 
+    def ping6_check(self, interface, count, option=None, flood=False):
+        """
+        IPV6 ping check to peer
+        """
+        cmd = "ip addr show %s" % interface
+        try:
+            for line in self.session.cmd(cmd).stdout.decode("utf-8") \
+                                                    .splitlines():
+                if "inet6" in line:
+                    ipv6 = line.split()[1].split('/')[0]
+        except Exception as ex:
+            if ipv6 == "":
+                log.error("unable to get IPV6 of peer")
+                return False
+        cmd = "ping -I %s %s -c %s" % (interface, ipv6, count)
+        if flood is True:
+            cmd = "%s -f" % cmd
+        elif option is not None:
+            cmd = "%s %s" % (cmd, option)
+        if process.system(cmd, shell=True, verbose=True,
+                          ignore_status=True) != 0:
+            return False
+        return True
+
 
 def is_interface_link_up(interface):
     """


avocado/utils/configure_network.py
```diff
@@ -189,6 +189,29 @@ def get_peer_interface(self, peer_ip):
...
+        """
+        IPV6 ping check to peer
+        """
+        cmd = "ip addr show %s" % interface
+        try:
+            for line in self.session.cmd(cmd).stdout.decode("utf-8") \
+                                                    .splitlines():
+                if "inet6" in line:
+                    ipv6 = line.split()[1].split('/')[0]
+        except Exception as ex:
```
beraldoleal: Suggestion: This exception is too generic. Can we try to make it more specific? (or with multiple except statements)

avocado/utils/configure_network.py
```diff
@@ -189,6 +189,29 @@ def get_peer_interface(self, peer_ip):
         else:
             return peer_interface
 
+    def ping6_check(interface, count, option=None, flood=False):
+        """
+        IPV6 ping check to peer
+        """
+        cmd = "ip addr show %s" % interface
+        try:
+            for line in self.session.cmd(cmd).stdout.decode("utf-8") \
```
beraldoleal: Maybe (it is just a guess/suggestion) if you use `/proc/net/if_inet6` would be easier and without the need for decode and for.

avocado/utils/configure_network.py
```diff
@@ -189,6 +189,29 @@ def get_peer_interface(self, peer_ip):
...
+                                                    .splitlines():
+                if "inet6" in line:
+                    ipv6 = line.split()[1].split('/')[0]
+        except Exception as ex:
+            if ipv6 == "":
+                log.error("unable to get IPV6 of peer")
```
beraldoleal: If it is not ipv6 available, the we should return False, right?

avocado/utils/configure_network.py
```diff
@@ -189,6 +189,29 @@ def get_peer_interface(self, peer_ip):
         else:
             return peer_interface
 
+    def ping6_check(interface, count, option=None, flood=False):
```
beraldoleal: The first parameter here should be self.

avocado/utils/configure_network.py
vaishnavibhat: The  /proc/net/if_inet6  is giving the IPV6 in the below format which is needs to be further modified to fe80::42f2:e9ff:fe31:XXXX for the ping6 command to work. Hence reading the ip command output. 
 #cat /proc/net/if_inet6 
fe8000000000000042f2e9fffe312XXX 14 40 20 80     eth2
fe8000000000000042f2e9fffe312XXX 13 40 20 80     eth1
fe8000000000000074b761fffe7a4XXX 02 40 20 80     eth0
00000000000000000000000000000001 01 80 10 80       lo

beraldoleal: @vaishnavibhat yes. But maybe you could use `ipaddress` library:

```python
>> raw = int('00000000000000000000000000000001', 16)
>> ipv6 = ipaddress.IPv6Address(raw).exploded
>> print(ipv6)
0000:0000:0000:0000:0000:0000:0000:0001
```

Both solutions are valid and have drawbacks and advantages. I'm just trying to keep the

beraldoleal: @vaishnavibhat sorry about the confusion. I was thinking about this and there is no way to get both: ipv4 and ipv6 in the same way via the filesystem. :( 

Yes, there are files with that information but we would need different methods to get them, so no consistency. Said that I think that we should keep using `iproute` utility's commands.

But, I'm still trying to keep the code clean and readable and trying to avoid the loop and decode.

Looking for the `ip` manpage I found the `-6` and `-j` (json) options. :) Simple snippet here:
(just a test)

```python3
import json
from avocado.utils.ssh import Session

session = Session('localhost', '22', 'user', password='123')
cmd = 'ip -6 -j address show lo'
output = json.loads(session.cmd(cmd).stdout)
print(output[0]['addr_info'][0]['local']) 
```

Is that better? What do you think? Any drawbacks?

beraldoleal: Also, keep in mind that there is another PR (#3580) in the review. I think that we should join the efforts here.

beraldoleal: Yes, there is one drawback. :(

Looks like -j was only introduced on version 4.17 that is not present on RHEL 7.

vaishnavibhat: I was seeing if we have json packages installed by default across distros so that we can proceed with minimal things to install further.. json seems be installed in the lates distro version, but like you mentioned if this cannot be used for RH7 version, then would like to follow the older approach to get the ipv6 address.

beraldoleal: @vaishnavibhat ok, but still we need to create cleaner code. Please see #3584.

beraldoleal:
@vaishnavibhat,  I'm closing this in favor of pull/#3636, please feel free to add your comments there.

PR #3570, opened 2020-02-18
Avocado.skip* usable on classes [v2]

richtja:
This commit expand the avocado.skip* decorators from methods to classes.
When a user uses skip decorator on the class, the decorator is applied
to every class method.

Reference: https://github.com/avocado-framework/avocado/issues/3473

---
Changes from v1 (#3566):
- Deletion of unused variable `value`
- Usage of `isinstance(obj, type)`

### DIFF
--- a/avocado/core/decorators.py
+++ b/avocado/core/decorators.py
@@ -73,14 +73,27 @@ def skip(message=None):
     """
     Decorator to skip a test.
     """
-    def decorator(function):
-        if not isinstance(function, type):
-            @wraps(function)
-            def wrapper(*args, **kwargs):  # pylint: disable=W0613
-                raise core_exceptions.TestSkipError(message)
-            function = wrapper
-        function.__skip_test_decorator__ = True
-        return function
+    def decorator(obj):
+        def method_decorator(function):
+            if not isinstance(function, type):
+                @wraps(function)
+                def wrapper(*args, **kwargs):  # pylint: disable=W0613
+                    raise core_exceptions.TestSkipError(message)
+                function = wrapper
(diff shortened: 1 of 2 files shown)


apahim:
While I like the functionality, I think there's room for discussion regarding concepts here.
When implementing the skip decorators, the consensus was that a test method that is skipped should have no code executed at all. For a test method, no code executed means no `setUp()`, no `testMethod()` and no `tearDown()`.
To extend the `skip` concept to the test classes, I'd expect the test class also not to have any code executed at all. In that case, no `__init__()`, no `log_dir`/`data_dir` initialization, nothing.
To achieve that, looping all the methods of the class and marking them as `skip` would not be enough. This decorator should, instead, avoid the class of being instantiated, so it would act in the loader, not in the test.

willianrampazzo:
> To achieve that, looping all the methods of the class and marking them as `skip` would not be enough. This decorator should, instead, avoid the class of being instantiated, so it would probably act in the loader, not in the test.

I agree that the implementation you described would fulfill the requirements you described and would be more efficient.

richtja:
@apahim, yes, you are right that skipping class should skip execution of all class code. But I think that to do that, we have to use a completely different approach than with skipping methods. And I am not sure if it's possible to accomplish that with python decorators. We were discussing that, on a meeting today, and it came up an idea that we can use this solution for now, because it handles most of the requirements, and also start looking for a more sophisticated solution that will be skipping whole class code. What do you think?

richtja:
I created issue #3579 that focus on skipping the whole classes.

Issue #3460, opened 2020-01-08
Move sysinfo implementation to the pre/post job/test hooks


beraldoleal:
PR https://github.com/avocado-framework/avocado/pull/1134, implements job pre/post hooks.

The sysinfo functionality looks like something that can be turned into a pre/post plugin, instead of having the fixed hooks in the core app.

beraldoleal:
More information on our (deprecated) trello card:

https://trello.com/c/6T2bX5sg/660-move-sysinfo-implementation-to-the-pre-post-job-test-hooks

beraldoleal:
@richtja I think this could be closed, since it is a very old issue, and you do have already some issues covering this. Please feel free to reopen if needed.

PR #3590, opened 2020-02-21
Way to provide sysinfo in case of test failure

narasimhan-v:
sysinfo collectibles were being collected in case of all test
results. But we need to collect some data in case of test failures
only. Introduced a way to do that.

### DIFF
--- a/avocado/core/sysinfo.py
+++ b/avocado/core/sysinfo.py
@@ -412,6 +412,20 @@ class SysInfo:
             log.debug('File %s does not exist.', commands_file)
             self.commands = []
 
+        fail_commands_file = settings.get_value('sysinfo.collectibles',
+                                                'fail_commands',
+                                                key_type='path',
+                                                default='')
+
+        if os.path.isfile(fail_commands_file):
+            log.info('Commands configured by file: %s', fail_commands_file)
+            self.fail_commands = genio.read_all_lines(fail_commands_file)
+            self.fail_commands = [
+                cmd for cmd in self.fail_commands if cmd not in self.commands]
+        else:
+            log.debug('File %s does not exist.', fail_commands_file)
+            self.fail_commands = []
+
         files_file = settings.get_value('sysinfo.collectibles',
                                         'files',
                                         key_type='path',
@@ -422,6 +436,19 @@ class SysInfo:
         else:
             log.debug('File %s does not exist.', files_file)
             self.files = []
+
+        fail_files_file = settings.get_value('sysinfo.collectibles',
+                                             'fail_files',
+                                             key_type='path',
+                                             default='')
+        if os.path.isfile(fail_files_file):
+            log.info('Files configured by file: %s', fail_files_file)
+            self.fail_files = genio.read_all_lines(fail_files_file)
+            self.fail_files = [
+                files for files in self.fail_files if files not in self.files]
+        else:
+            log.debug('File %s does not exist.', fail_files_file)
+            self.fail_files = []
 
         if profiler is None:
             self.profiler = settings.get_value('sysinfo.collect',
@@ -457,10 +484,13 @@ class SysInfo:
         self.start_test_collectibles = set()
         self.end_test_collectibles = set()
 
+        self.end_test_fail_collectibles = set()
+
         self.hook_mapping = {'start_job': self.start_job_collectibles,
                              'end_job': self.end_job_collectibles,
                              'start_test': self.start_test_collectibles,
-                             'end_test': self.end_test_collectibles}
+                             'end_test': self.end_test_collectibles,
+                             'end_fail_test': self.end_test_fail_collectibles}
 
         self.pre_dir = utils_path.init_dir(self.basedir, 'pre')
         self.post_dir = utils_path.init_dir(self.basedir, 'post')
@@ -500,6 +530,13 @@ class SysInfo:
             if add_per_test:
                 self.start_test_collectibles.add(Logfile(filename))
                 self.end_test_collectibles.add(Logfile(filename))
+
+        if add_per_test:
+            for fail_filename in self.fail_files:
+                self.end_test_fail_collectibles.add(Logfile(fail_filename))
+            for fail_cmd in self.fail_commands:
+                self.end_test_fail_collectibles.add(Command(fail_cmd))
+
 
         # As the system log path is not standardized between distros,
         # we have to probe and find out the correct path.
@@ -614,10 +651,14 @@ class SysInfo:
         if self.log_packages:
             self._log_installed_packages(self.pre_dir)
 
-    def end_test_hook(self):
+    def end_test_hook(self, status=""):
         """
         Logging hook called after a test finishes.
         """
+        if status == "FAIL":
+            for log_hook in self.end_test_fail_collectibles:
+                log_hook.run(self.post_dir)
+            return
         for log_hook in self.end_test_collectibles:
             log_hook.run(self.post_dir)
         # Stop daemon(s) started previously
--- a/avocado/core/test.py
+++ b/avocado/core/test.py
@@ -1047,6 +1047,8 @@ class Test(unittest.TestCase, TestData):
         :warning message: This parameter will changed name to "msg" in the next
                           LTS release because of lint W0221
         """
+        if self.__sysinfo_enabled:
+            self.__sysinfo_logger.end_test_hook(status="FAIL")
         raise exceptions.TestFail(message)
 
     def error(self, message=None):  # pylint: disable=W0221
--- a/avocado/etc/avocado/avocado.conf
+++ b/avocado/etc/avocado/avocado.conf
@@ -29,8 +29,12 @@ per_test = False
 [sysinfo.collectibles]
 # File with list of commands that will be executed and have their output collected
 commands = etc/avocado/sysinfo/commands
+# File with list of commands that will be executed and have their output collected, in case of test failure
+fail_commands = etc/avocado/sysinfo/fail_commands
 # File with list of files that will be collected verbatim
 files = etc/avocado/sysinfo/files
+# File with list of files that will be collected verbatim, in case of test failure
+fail_files = etc/avocado/sysinfo/fail_files
 # File with list of commands that will run alongside the job/test
 profilers = etc/avocado/sysinfo/profilers
 


narasimhan-v:
```
JOB ID     : 186b9da5883e148f3ab25adc070eb81ddb11788d
JOB LOG    : /root/avocado/job-results/job-2020-02-21T02.08-186b9da/job.log
 (1/1) ../avocado-misc-tests/io/disk/fail.py:Ioping.test: FAIL: test run fails (0.71 s)
RESULTS    : PASS 0 | ERROR 0 | FAIL 1 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB TIME   : 1.11 s

# cat /root/avocado/avocado/etc/avocado/sysinfo/fail_files
/proc/uptime
/proc/modules
/proc/devices

# ls /root/avocado/job-results/job-2020-02-21T02.08-186b9da/test-results/1-.._avocado-misc-tests_io_disk_fail.py_Ioping.test/sysinfo/pre
 cmdline               dmesg           'ifconfig -a'    lscpu          mounts                     sched_features   version
 cpuinfo              'fdisk -l'        interrupts     'lspci -vvnn'  'numactl --hardware show'   slabinfo
 current_clocksource  'gcc --version'  'ip link'        meminfo        partitions                'uname -a'
'df -mP'               hostname        'ld --version'   modules        scaling_governor           uptime

# ls /root/avocado/job-results/job-2020-02-21T02.08-186b9da/test-results/1-.._avocado-misc-tests_io_disk_fail.py_Ioping.test/sysinfo/post/
 cmdline              'df -mP'    'gcc --version'  'ip link'       'lspci -vvnn'   mounts                     sched_features   version
 cpuinfo               dmesg       hostname         journalctl.gz   meminfo       'numactl --hardware show'   slabinfo
 current_clocksource  'echo 1'    'ifconfig -a'    'ld --version'   messages.gz    partitions                'uname -a'
 devices              'fdisk -l'   interrupts       lscpu           modules        scaling_governor           uptime 
```

avocado/core/sysinfo.py
```diff
@@ -614,10 +647,14 @@ def start_test_hook(self):
...
 
-    def end_test_hook(self):
+    def end_test_hook(self, status=""):
         """
         Logging hook called after a test finishes.
         """
+        if status == "FAIL":
+            for log_hook in self.end_test_fail_collectibles:
+                log_hook.run(self.post_dir)
+            return
```
sathnaga: I guess it is better to keep the default collectibles run for all status and additionally fail_collectibles  when test failed.

avocado/core/sysinfo.py
```diff
@@ -457,10 +480,13 @@ def __init__(self, basedir=None, log_packages=None, profiler=None):
         self.start_test_collectibles = set()
         self.end_test_collectibles = set()
 
+        self.end_test_fail_collectibles = set()
```
sathnaga: let's remove the common files present in `end_test_collectibles` and `end_test_fail_collectibles`, users are free to add same files and commands in both places, it is better to find only the additional newer ones compared to `end_test_collectibles` and add to `end_test_fail_collectibles`.

so we can just be sure only additional files and commands needed for failed cases and without breaking the flow of all status..

E:g:- expecting like below... 
>>> a={2,3,4} # end_test_collectibles
>>> b={2,5,6} # end_test_fail_collectibles
>>> b = b-a
>>> b
{5, 6}

avocado/core/sysinfo.py
narasimhan-v: @sathnaga that is what I am doing here, right ?

avocado/core/sysinfo.py
narasimhan-v: Sure, will remove duplicates from end_test_fail_collectibles

narasimhan-v: Sure, will remove duplicates from end_test_fail_collectibles

beraldoleal:
I just would like to add that we have plans to move sysinfo to hooks (please, see:  #3460).

But I'm not sure how this will impact on that and if we should do this now or after the migration.

@clebergnu , any comments?

PraveenPenguin:
> I just would like to add that we have plans to move sysinfo to hooks (please, see: #3460).
> 
> But I'm not sure how this will impact on that and if we should do this now or after the migration.
> 
> @clebergnu , any comments?

Hi, @beraldoleal we can have this in as we need this for our uses case, though I did not do a thorough review. I think  can be targeted when we migrate this can also be taken care

avocado/core/sysinfo.py
narasimhan-v: done it immediately after generating the list. please check.

avocado/core/sysinfo.py
```diff
@@ -412,6 +412,20 @@ def __init__(self, basedir=None, log_packages=None, profiler=None):
...
+                                                key_type='path',
+                                                default='')
+
+        if os.path.isfile(fail_commands_file):
+            log.info('Commands configured by file: %s', fail_commands_file)
+            self.fail_commands = genio.read_all_lines(fail_commands_file)
+        else:
+            log.debug('File %s does not exist.', fail_commands_file)
+            self.fail_commands = []
+        self.fail_commands = [
```
sathnaga: let's move this inside `if`

avocado/core/sysinfo.py
```diff
@@ -423,6 +437,20 @@ def __init__(self, basedir=None, log_packages=None, profiler=None):
...
+                                             key_type='path',
+                                             default='')
+        if os.path.isfile(fail_files_file):
+            log.info('Files configured by file: %s', fail_files_file)
+            self.fail_files = genio.read_all_lines(fail_files_file)
+        else:
+            log.debug('File %s does not exist.', fail_files_file)
+            self.fail_files = []
+
+        self.fail_files = [
```
sathnaga: let's move this inside `if`

avocado/core/sysinfo.py
sathnaga: lets not restrict the default run, just add our hook when enabled.

Now that we have taken out common files, both collection are needed.

narasimhan-v: @sathnaga In my implementation, the hook gets called twice, once for all cases, and one for fail cases.
In all cases, it runs for all generic collectibles.
In fail cases, it runs for fail collectibles alone.

Can you please explain your thoughts a bit more ?

avocado/core/sysinfo.py
narasimhan-v: Sure

avocado/core/sysinfo.py
sathnaga: that's what am trying to understand, why two times to call, is this not suffice, https://github.com/avocado-framework/avocado/blob/214d2191d5035647c9aa4370f888923f16819466/avocado/core/test.py#L950, just that we need export the status, or use global status variable if any?

narasimhan-v: @sathnaga Right. As of today, sysinfo does not have the context of status at all. And from the existing sysinfo hook call, status does not exist. So, I had to take this approach of explicitly calling the hook with status in fail(), by not disturbing any of the existing workflow.

beraldoleal:
> > I just would like to add that we have plans to move sysinfo to hooks (please, see: #3460).
> > But I'm not sure how this will impact on that and if we should do this now or after the migration.
> > @clebergnu , any comments?
> 
> Hi, @beraldoleal we can have this in as we need this for our uses case, though I did not do a thorough review. I think can be targeted when we migrate this can also be taken care

I understand that we don't need to move the implementation to a hook right now since that you need for your uses cases. But I think that once is a proposal to be merged into master, this **must be** reviewed properly.

narasimhan-v:
> As for the implementation I'd suggest moving the sysinfo collection to `finally` in `run_avocado` (around line `1016` in `avocado/core/test.py`) and using the `__status` to distinguish between current situation.

I was planning to move the sysinfo collection, but then there were plans of an overhaul into a plugin.
So need comments on what all can be done now, or can we do the overhaul into a plugin and this together ?

> 
> As for the configuration, how about creating dirs in `/etc/avocado/sysinfo` called `on_failure`, `on_error`, `on_skip`, ... and then using the `commands, files, ...` files in there or fallback to the `/etc/avocado/sysinfo` files when missing?
> 
> Also one question, do you want to extend, or replace the set of collected sysinfo? What I mean do you want to run the stock sysinfo collection and then run the extra per-status sysinfo, or do you want to run both? I'd probably prefer to always run the `/etc/avocado/sysinfo/*` collections and then extend it of the per status collections (eg. `/etc/avocado/sysinfo/on_fail/*` files, or even `on_pass` in case of a passed test). That way we'd have a shared minimum set and extended set per all statuses.

Yes, thanks for the suggestion @ldoktor that would be better.

avocado/core/sysinfo.py
```diff
@@ -412,6 +412,20 @@ def __init__(self, basedir=None, log_packages=None, profiler=None):
             log.debug('File %s does not exist.', commands_file)
             self.commands = []
 
+        fail_commands_file = settings.get_value('sysinfo.collectibles',
+                                                'fail_commands',
+                                                key_type='path',
+                                                default='')
```
beraldoleal: We have introduced the new `core.future.settings` module to handle configuration options. The current settings will be deprecated soon.

While accepting this current setting is ok, we should try to migrate to the new model, where default values are no more defined when "getting" (`config.get()`) values, but when "registering" (`settings.register_option()`) values.

Please, take a look into other examples of `core.future.settings` in the source code, and let me know if you have any questions.

avocado/core/sysinfo.py
```diff
@@ -412,6 +412,20 @@ def __init__(self, basedir=None, log_packages=None, profiler=None):
             log.debug('File %s does not exist.', commands_file)
             self.commands = []
 
+        fail_commands_file = settings.get_value('sysinfo.collectibles',
+                                                'fail_commands',
```
beraldoleal: I suggest adding `_file` to this option to make it more intuitive for the users.

avocado/core/sysinfo.py
```diff
@@ -423,6 +437,19 @@ def __init__(self, basedir=None, log_packages=None, profiler=None):
             log.debug('File %s does not exist.', files_file)
             self.files = []
 
+        fail_files_file = settings.get_value('sysinfo.collectibles',
+                                             'fail_files',
+                                             key_type='path',
+                                             default='')
```
beraldoleal: Same here.

PraveenPenguin:
@narasimhan-v can we  address @beraldoleal comment

narasimhan-v:
> @narasimhan-v can we address @beraldoleal comment

Sure, I can respin a patch. I was actually waiting for @ldoktor for more comments on the plugin overhaul part.

avocado/core/sysinfo.py
narasimhan-v: @beraldoleal I am looking to send a v2 of this with all comments addressed, so for my clarity, you mean: `fail_commands_file = settings.get_value('sysinfo.collectibles', 'fail_commands', ...` ?

avocado/core/sysinfo.py
narasimhan-v: @beraldoleal Looks like we are still in the older method in case of other sysinfo, do you want me to migrate existing ones also to new register method ? Or can that be taken as a separate patch ?

narasimhan-v:
@beraldoleal @PraveenPenguin Please review the v2 at https://github.com/avocado-framework/avocado/pull/3874.

@ldoktor I am planning to address your comments on the per_status in a separate enhancement, after getting comments via an RFC.

PR #3594, opened 2020-02-24
fixed configure_network  mtu set host and peer for timing issue

smruti77:
added to check mtu value and link up after mtu set through /sys/class/net/
to avoid timing issue.

### DIFF
--- a/avocado/utils/configure_network.py
+++ b/avocado/utils/configure_network.py
@@ -25,6 +25,7 @@ from . import distro
 from . import process
 from . import genio
 from .ssh import Session
+from . import wait
 
 log = logging.getLogger('avocado.test')
 
@@ -132,10 +133,8 @@ def set_mtu_host(interface, mtu):
     except process.CmdError as ex:
         raise NWException("MTU size can not be set: %s" % ex)
     try:
-        cmd = "ip add show %s" % interface
-        mtuvalue = process.system_output(cmd, shell=True).decode("utf-8") \
-                                                         .split()[4]
-        if mtuvalue == mtu:
+        output = process.system_output("cat /sys/class/net/%s/mtu" % interface).decode("utf-8")
+        if mtu in output and wait.wait_for(is_interface_link_up, timeout=120, args=[interface]):
             return True
     except Exception as ex:  # pylint: disable=W0703
         log.error("setting MTU value in host failed: %s", ex)
@@ -165,9 +164,9 @@ class PeerInfo:
         cmd = "ip link set %s mtu %s" % (peer_interface, mtu)
         try:
             self.session.cmd(cmd)
-            cmd = "ip add show %s" % peer_interface
-            mtuvalue = self.session.cmd(cmd).stdout.decode("utf-8").split()[4]
-            if mtuvalue == mtu:
+            output = self.session.cmd("cat /sys/class/net/%s/mtu" % peer_interface)
+            if mtu in output.stdout.decode('utf-8') and wait.wait_for(self.is_link_up_peer, timeout=120,
+                                                                      args=[peer_interface]):
                 return True
         except Exception as ex:  # pylint: disable=W0703
             log.error("setting MTU value in peer failed: %s", ex)
@@ -188,6 +187,18 @@ class PeerInfo:
                 log.error("unable to get peer interface: %s", ex)
         else:
             return peer_interface
+
+    def is_link_up_peer(self, peer_interface):
+        """
+        check if peer interface link is up
+        :param peer_interface: name of the peer interface
+        :retrun: True if the interface link is up, False otherwise.
+        """
+        output = self.session.cmd("cat /sys/class/net/%s/operstate" % peer_interface)
+        if 'up' in output.stdout.decode('utf-8'):
+            log.info("Interface %s link is up", peer_interface)
+            return True
+        return False
 
 
 def is_interface_link_up(interface):


smruti77:
[multiport.configure.log](https://github.com/avocado-framework/avocado/files/4243819/multiport.configure.log)

avocado/utils/configure_network.py
```diff
@@ -189,6 +187,16 @@ def get_peer_interface(self, peer_ip):
         else:
             return peer_interface
 
+    def is_interface_link_up_peer(self, peer_interface):
```
beraldoleal: IIUC, this method is generic enought to be renamed to `is_interface_link_up(self, interface)` or `is_link_up(self, interface)`, right? 
We don't need the "peer" here, right?

avocado/utils/configure_network.py
```diff
@@ -165,9 +163,9 @@ def set_mtu_peer(self, peer_interface, mtu):
         cmd = "ip link set %s mtu %s" % (peer_interface, mtu)
         try:
             self.session.cmd(cmd)
-            cmd = "ip add show %s" % peer_interface
-            mtuvalue = self.session.cmd(cmd).stdout.decode("utf-8").split()[4]
-            if mtuvalue == mtu:
+            output = self.session.cmd("cat /sys/class/net/%s/mtu" % peer_interface)
+            if mtu in output.stdout.decode('utf-8') and wait.wait_for(self.is_interface_link_up_peer,
+                                                                      timeout=120, args=[peer_interface]):
```
beraldoleal: Much better! Thanks!

avocado/utils/configure_network.py
```diff
@@ -132,10 +133,7 @@ def set_mtu_host(interface, mtu):
     except process.CmdError as ex:
         raise NWException("MTU size can not be set: %s" % ex)
     try:
-        cmd = "ip add show %s" % interface
-        mtuvalue = process.system_output(cmd, shell=True).decode("utf-8") \
-                                                         .split()[4]
-        if mtuvalue == mtu:
+        if mtu in genio.read_file("/sys/class/net/%s/mtu" % interface) and is_interface_link_up(interface):
```
beraldoleal: Why here, you are using `genio.read_file` and on another method, you are using `self.session.cmd`? If possible, let's try to use the same method.

avocado/utils/configure_network.py
```diff
@@ -189,6 +187,16 @@ def get_peer_interface(self, peer_ip):
         else:
             return peer_interface
 
+    def is_interface_link_up_peer(self, peer_interface):
+        """
+        check if peer interface link is up
```
beraldoleal: Also, I would like to suggest a better (multi-line) docstring here:

> Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description. 

avocado/utils/configure_network.py
smruti77: is_interface_link_up() this one already their in file for host machine, i did for the peer machine this one.

avocado/utils/configure_network.py
smruti77: ok i will change, genio.read_file i couldn't able to use in remote machine s i did like that.

avocado/utils/configure_network.py
beraldoleal: Hi @smruti77, I think that I understood your point. In this case, maybe we need a small refactoring here. Sorry if this is taking so long to be accepted, but we need to make sure that it is gonna be good for everyone.

IIUC, we should use more inheritance here and have a clear separation between a Host and an Interface. For the user perspective the usability should be the same:

```python3
local_host = LocalHost()
remote_host = RemoteHost(host='10.0.0.1', port='22', username='foo', password='key')
local_host.interfaces
remote_host.interfaces
```

IMO, the same API should be used to set and/or change interface attributes (regardless if is a remote or a local connection).

Also, how this current implementation will support multiple interfaces? Do I need to open a new connection for each interface? (I'm still not sure how we will handle this).

I'm just thinking as the API user. The usability should matter here. In my view, something like this could be more elegant (from the user perspective):

```python3
interfaces = []
for name in ['eth0', 'eth1', 'lo']:
   interfaces.append(Interface(name))
local_host.interfaces = interfaces
remote_host.interfaces = interfaces
```
Or instead of setting manually the interfaces on the host, even better: this could be dynamic during the Host's `__init__()` method, so:

```python3
local_host = Host()
for interface in local_host.interfaces:
   print(interface.name)
   interface.set_mtu(200)
```

In this second example, the host will discover all interfaces and make available via API.
Of course, in both cases, we need an `Interface` class.

Why I'm suggesting this big change? Because there is already a refactoring in the course (see PR #3605) by @PraveenPenguin . And I think that we should take this opportunity to do a better organization here.

@smruti77, @PraveenPenguin, if you prefer, I'm available on IRC (bleal) and we could make a call to better organize this effort.

beraldoleal: This is just an example (probably not functional and with some syntax errors):

```python3
class Interface:
    def __init__(self, name, if_type='Ethernet', remote=False):
        raise Exception("Must be implemented")

    def set_ipaddr(self, ipaddr, netmask, version=4, persistent=False):
        if version==4:
            return self.set_ipv4_addr(ipaddr, netmask, persistent):
        elif version==6:
            return self.set_ipv6_addr(ipaddr, netmask, persistent):

    def set_ipv4_addr(self, ipaddr, netmask, persistent=False):
        raise Exception("Must be implemented calling save when need")

    def set_ipv6_addr(self, ipaddr, netmask, persistent=False):
        raise Exception("Must be implemented calling save when need")

    def set_mtu(self, mtu, persistent=False):
        raise Exception("Must be implemented calling save when need")
    
    def get_mtu(self):
        raise Exception("Must be implemented")

    def get_ipaddr(self, version=4):
        raise Exception("Must be implemented")

    def get_link_state(self):
        raise Exception("Must be implemented")
       
    def is_link_up(self):
        if self.get_link_state() in ['up', 'UP']:
            return True
        else:
            return False

    def save(self):
        raise Exception("Must be implemented")


class Host:
    def __init__(self, host=None, port=22, username=None, key=None, password=None):
        self.host = host
        self.port = port
        self.username = username
        self.key = key
        self.password = password
        self.session = None
        self.interfaces = []
        self._connect()
        self._populate_interfaces()

    def _connect(self):
        if (self.host and self.port and self.username):
            try:
                self.session = Session(host, port, username, key, password)
            except Exception as ex:
                raise Exception("Could not connect to host: {}".format(ex)) 

    def is_remote(self):
        if self.session:
            return True
        else:
            return False

    def _populate_interfaces(self):
        is_remote = self.is_remote()
        if is_remote:
            names = ...
        else:
            names = os.listdir('/sys/class/net')

        for name in names:
            self.interfaces.append(Interface(name, remote=is_remote))

```

smruti77: > Hi @smruti77, I think that I understood your point. In this case, maybe we need a small refactoring here. Sorry if this is taking so long to be accepted, but we need to make sure that it is gonna be good for everyone.
> 
> IIUC, we should use more inheritance here and have a clear separation between a Host and an Interface. For the user perspective the usability should be the same:
> 
> ```python
> local_host = Host()
> remote_host = Host(host='10.0.0.1', port='22', username='foo', password='key')
> local_host.interfaces
> remote_host.interfaces
> ```
> 
> IMO, the same API should be used to set and/or change interface attributes (regardless if is a remote or a local connection).
> 
> Also, how this current implementation will support multiple interfaces? Do I need to open a new connection for each interface? (I'm still not sure how we will handle this).
> 
> I'm just thinking as the API user. The usability should matter here. In my view, something like this could be more elegant (from the user perspective):
> 
> ```python
> interfaces = []
> for name in ['eth0', 'eth1', 'lo']:
>    interfaces.append(Interface(name))
> local_host.interfaces = interfaces
> remote_host.interfaces = interfaces
> ```
> 
> Or instead of setting manually the interfaces on the host, even better: this could be dynamic during the Host's `__init__()` method, so:
> 
> ```python
> local_host = Host()
> for interface in local_host.interfaces:
>    print(interface.name)
>    interface.set_mtu(200)
> ```
> 
> In this second example, the host will discover all interfaces and make available via API.
> Of course, in both cases, we need an `Interface` class.
> 
> Why I'm suggesting this big change? Because there is already a refactoring in the course (see PR #3605) by @PraveenPenguin . And I think that we should take this opportunity to do a better organization here.
> 
> @smruti77, @PraveenPenguin, if you prefer, I'm available on IRC (bleal) and we could make a call to better organize this effort.

@beraldoleal i understood your point of view,this is better way to handle all those function.ok will try to do like.
and about multiple interface, here directly we can't pass multiple interface but in our own script we can handle that.

PraveenPenguin: > > Hi @smruti77, I think that I understood your point. In this case, maybe we need a small refactoring here. Sorry if this is taking so long to be accepted, but we need to make sure that it is gonna be good for everyone.
> > IIUC, we should use more inheritance here and have a clear separation between a Host and an Interface. For the user perspective the usability should be the same:
> > ```python
> > local_host = Host()
> > remote_host = Host(host='10.0.0.1', port='22', username='foo', password='key')
> > local_host.interfaces
> > remote_host.interfaces
> > ```
> > 
> > 
> > IMO, the same API should be used to set and/or change interface attributes (regardless if is a remote or a local connection).
> > Also, how this current implementation will support multiple interfaces? Do I need to open a new connection for each interface? (I'm still not sure how we will handle this).
> > I'm just thinking as the API user. The usability should matter here. In my view, something like this could be more elegant (from the user perspective):
> > ```python
> > interfaces = []
> > for name in ['eth0', 'eth1', 'lo']:
> >    interfaces.append(Interface(name))
> > local_host.interfaces = interfaces
> > remote_host.interfaces = interfaces
> > ```
> > 
> > 
> > Or instead of setting manually the interfaces on the host, even better: this could be dynamic during the Host's `__init__()` method, so:
> > ```python
> > local_host = Host()
> > for interface in local_host.interfaces:
> >    print(interface.name)
> >    interface.set_mtu(200)
> > ```
> > 
> > 
> > In this second example, the host will discover all interfaces and make available via API.
> > Of course, in both cases, we need an `Interface` class.
> > Why I'm suggesting this big change? Because there is already a refactoring in the course (see PR #3605) by @PraveenPenguin . And I think that we should take this opportunity to do a better organization here.
> > @smruti77, @PraveenPenguin, if you prefer, I'm available on IRC (bleal) and we could make a call to better organize this effort.
> 
> @beraldoleal i understood your point of view,this is better way to handle all those function.ok will try to do like.
> and about multiple interface, here directly we can't pass multiple interface but in our own script we can handle that.

@smruti77 i liked this approach as #3605   we are trying to do refactoring, if you are fine  i can combine  this effort in that PR itself

smruti77: @PraveenPenguin ok ,you can combine this on that PR.

beraldoleal:
@smruti77 , A few things were not working here. I'm closing this in favor of pull/#3636, please feel free to add your comments there.

PR #3595, opened 2020-02-24
Add funtion to get wwid for the given userfriendly mpath.

harihare:
get the wwid for the given mpath user friendly name.

Signed-off-by: Hariharan T S <hari@linux.vnet.ibm.com>

### DIFF
--- a/avocado/utils/multipath.py
+++ b/avocado/utils/multipath.py
@@ -26,6 +26,12 @@ from . import service
 from . import wait
 
 
+class MPException(Exception):
+    """
+    Base Exception Class for all exceptions
+    """
+
+
 def get_svc_name():
     """
     Gets the multipath service name based on distro.
@@ -104,6 +110,42 @@ def get_multipath_wwids():
                         sudo=True, shell=True).stdout_text
     wwids = wwids.strip("\n").replace("/", "").split("\n")
     return wwids
+
+
+def get_multipath_wwid(mpath):
+    """
+    Get the wwid binding for given mpath name
+
+    :return: Multipath wwid
+    :rtype: str
+    """
+    cmd = "multipathd show maps format '%n %w'"
+    try:
+        wwids = process.run(cmd, ignore_status=True,
+                            sudo=True, shell=True).stdout_text
+    except process.CmdError as ex:
+        raise MPException("Multipathd Command Failed : %s " % ex)
+    for wwid in wwids.splitlines():
+        if mpath in wwid:
+            return wwid.split()[1]
+
+
+def is_mpath_dev(mpath):
+    """
+    Check the give name is a multipath device name or not.
+
+    :return: True if device is multipath or False
+    :rtype: Boolean
+    """
+    cmd = "multipath -l -v 1"
+    try:
+        mpaths = process.run(cmd, ignore_status=True,
+                             sudo=True, shell=True).stdout_text
+    except process.CmdError as ex:
+        raise MPException("Multipath Command Failed : %s " % ex)
+    if mpath in mpaths.strip('\n').split("\n"):
+        return True
+    return False
 
 
 def get_paths(wwid):


avocado/utils/multipath.py
```diff
@@ -106,6 +106,19 @@ def get_multipath_wwids():
...
 
 
+def get_multipath_wwid(mpath):
+    """
+    Get the wwid binding for given mpath name
+
+    :return: Multipath wwid
+    :rtype: str
+    """
+    cmd = "egrep -i mpathd /etc/multipath/bindings"
```
PraveenPenguin: using bash in python not looks good, Better to use re module to processing of file content, even I suggest use `genio` to populate file content

avocado/utils/multipath.py
```diff
@@ -106,6 +106,19 @@ def get_multipath_wwids():
     return wwids
 
 
+def get_multipath_wwid(mpath):
+    """
+    Get the wwid binding for given mpath name
+
```
PraveenPenguin: doc string for input `parm` as well, please

avocado/utils/multipath.py
```diff
@@ -106,6 +106,19 @@ def get_multipath_wwids():
...
+    :return: Multipath wwid
+    :rtype: str
+    """
+    cmd = "egrep -i mpathd /etc/multipath/bindings"
+    wwid = process.run(cmd, ignore_status=True,
+                        sudo=True, shell=True).stdout_text
+    return wwid.split()[1]
```
beraldoleal: this should be inside a try.. except

avocado/utils/multipath.py
beraldoleal: Also, try to use a multi-line docstring following the PEP-0257:

avocado/utils/multipath.py
beraldoleal: Also, take in consideration other systems:

In the multipath documentation there is a configuration parameter called user_friendly_names with the following definition:

If set to yes, specifies that the system should use the /etc/multipath/bindings file to assign a persistent and unique alias to the multipath, in the form of mpath n. If set to no, specifies that the system should use the WWID as the alias for the multipath. In either case, what is specified here will be overridden by any device-specific aliases you specify in the multipaths section of the configuration file. The default value is no.

harihare: Check removed on bindings, as the name can be overridden with the help of multipaths section in configuration file. Code is updated to find the WWID from the name (name can be user friendly name or the wwid itself) in any case the code will return the wwid of the device.

avocado/utils/multipath.py
harihare: Check is done before the value is returned.

avocado/utils/multipath.py
harihare: Multiline comment added.

avocado/utils/multipath.py
```diff
@@ -106,6 +106,36 @@ def get_multipath_wwids():
...
+    Get the wwid binding for given mpath name
+
+    :return: Multipath wwid
+    :rtype: str
+    """
+    cmd = "multipathd show maps format '%%n %%w' | grep %s" % mpath
+    wwid = process.run(cmd, ignore_status=True,
+                        sudo=True, shell=True).stdout_text
+    if wwid:
+        return wwid.split()[1]
```
narasimhan-v, line 120: @harihare Please use native python parsing instead of bash parsing
```
cmd = "multipathd show maps format '%%n %%w'
wwid = process.run(cmd, ignore_status=True,
                   sudo=True, shell=True).stdout_text
if mpath in wwid.splitlines():
    return wwid.split()[1]
````

Comment (harihare) [reply avocado/utils/multipath.py]: Done

File: avocado/utils/multipath.py (hunk 1 above)
Comment (PraveenPenguin) [reply avocado/utils/multipath.py]: ah still i see we are using bash any way we can target this as refactor  for whole multipath API in future

**Comment by harihare** [comment]
Test log :
[root@ltcfleet2-lp4 disk]# avocado run disk_info.py -m disk_info.py.data/disk_info.yaml
/usr/local/lib/python3.6/site-packages/avocado_framework-78.0-py3.6.egg/avocado/plugins/run.py:287: FutureWarning: The following arguments will be changed to boolean soon: sysinfo, output-check, failfast, keep-tmp and ignore-missing-references.
  FutureWarning)
JOB ID     : 034ba7a1091a1997e90ca91b629ab278784becb8
JOB LOG    : /root/bootlist-output/results/job-2020-04-23T06.24-034ba7a/job.log
 (1/6) disk_info.py:DiskInfo.test_commands;ext4-c1e6: PASS (9.19 s)
 (2/6) disk_info.py:DiskInfo.test_commands;xfs-4659: PASS (4.43 s)
 (3/6) disk_info.py:DiskInfo.test_commands;btrfs-1508: CANCEL: btrfs is not supported with                                 RHEL 7.4 onwards (0.19 s)
 (4/6) disk_info.py:DiskInfo.test;ext4-c1e6: PASS (13.46 s)
 (5/6) disk_info.py:DiskInfo.test;xfs-4659: PASS (11.77 s)
 (6/6) disk_info.py:DiskInfo.test;btrfs-1508: CANCEL: btrfs is not supported with                                 RHEL 7.4 onwards (0.18 s)
RESULTS    : PASS 4 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 2
JOB TIME   : 39.55 s
JOB HTML   : /root/bootlist-output/results/job-2020-04-23T06.24-034ba7a/results.html

File: avocado/utils/multipath.py [hunk 5]
```diff
@@ -106,6 +106,37 @@ def get_multipath_wwids():
...
+def is_mpath(mpath):
+    """
+    Check the give name is a multipath device name or not.
+
+    :return: True if device is multipath or False
+    :rtype: Boolean
+    """
+    cmd = "multipath -l -v 1"
+    mpaths = process.run(cmd, ignore_status=True,
+                        sudo=True, shell=True).stdout_text
```
beraldoleal, line 133: I think that there is a trailing space here and/or an indentation problem that is making Travis complain.

avocado/utils/multipath.py
```diff
@@ -106,6 +106,37 @@ def get_multipath_wwids():
...
+    """
+    cmd = "multipathd show maps format '%n %w'"
+    wwids = process.run(cmd, ignore_status=True,
+                        sudo=True, shell=True).stdout_text
+    for wwid in wwids.splitlines(): 
+        if mpath in wwid:
+            return wwid.split()[1]
+
+
+def is_mpath(mpath):
```
beraldoleal, line 124: Just a small suggestion on a better function name: `is_mpath_dev` ?

harihare: Changed the function name to is_mpath_dev.

avocado/utils/multipath.py
harihare: fixed the indentation issue.

avocado/utils/multipath.py
```diff
@@ -106,6 +106,37 @@ def get_multipath_wwids():
...
+    """
+    Get the wwid binding for given mpath name
+
+    :return: Multipath wwid
+    :rtype: str
+    """
+    cmd = "multipathd show maps format '%n %w'"
+    wwids = process.run(cmd, ignore_status=True,
+                        sudo=True, shell=True).stdout_text
+    for wwid in wwids.splitlines():
```
beraldoleal, line 119: I think that Travis is complaining about a trailing space here. Not sure.

harihare: Even I checked there, there is not trailing space at all.  Not sure why it is.

avocado/utils/multipath.py
```diff
@@ -106,6 +106,37 @@ def get_multipath_wwids():
...
+
+    :return: True if device is multipath or False
+    :rtype: Boolean
+    """
+    cmd = "multipath -l -v 1"
+    mpaths = process.run(cmd, ignore_status=True,
+                         sudo=True, shell=True).stdout_text
+    if mpath in mpaths.strip('\n').split("\n"):
+        return True
+    else:
```
beraldoleal, line 136: You don't need the else here.

avocado/utils/multipath.py
```diff
@@ -106,6 +106,37 @@ def get_multipath_wwids():
...
 
+def get_multipath_wwid(mpath):
+    """
+    Get the wwid binding for given mpath name
+
+    :return: Multipath wwid
+    :rtype: str
+    """
+    cmd = "multipathd show maps format '%n %w'"
+    wwids = process.run(cmd, ignore_status=True,
```
beraldoleal, line 117: This should be inside a try/except? The command could be missing on the host for instance, or a permission denied could be raised.

avocado/utils/multipath.py
```diff
@@ -106,6 +106,37 @@ def get_multipath_wwids():
...
+
+def is_mpath_dev(mpath):
+    """
+    Check the give name is a multipath device name or not.
+
+    :return: True if device is multipath or False
+    :rtype: Boolean
+    """
+    cmd = "multipath -l -v 1"
+    mpaths = process.run(cmd, ignore_status=True,
```
beraldoleal, line 132: Same here.

avocado/utils/multipath.py
harihare: Removed the else part.

avocado/utils/multipath.py
harihare: Process.run has the error handling and reporting mechanism. Hence I think a separate try/except it not required here

avocado/utils/multipath.py
harihare: Process.run has the error handling and reporting mechanism. Hence I think a separate try/except it not required here

PraveenPenguin:
@beraldoleal can you please ACK if it looks ok to you without  `exception handling` as  exception handling we can address in another patch as it requires to adopt in entire multipath lib and we have more discussion on this from the owner of this utils (@narasimhan-v  ) if it really needed or not

avocado/utils/multipath.py
beraldoleal: @harihare if `process.run()` raises exceptions we should catch them on the caller and raise a more high-level exception. Otherwise, the user will receive a low-level exception and not a specif one. In this case a try...except would be useful here to raise another exception saying that the getting the wwid binding for a given mpath was failed plus the low-level message that `process.run()` sent to us.

avocado/utils/multipath.py
```diff
@@ -106,6 +106,36 @@ def get_multipath_wwids():
...
```
beraldoleal, line 132: same here.

beraldoleal:
> @beraldoleal can you please ACK if it looks ok to you without `exception handling` as exception handling we can address in another patch as it requires to adopt in entire multipath lib and we have more discussion on this from the owner of this utils (@narasimhan-v ) if it really needed or not

IMO it is better if our libraries could deliver a proper high-level exception with detailed information that came from low-level calls.

avocado/utils/multipath.py
narasimhan-v: @beraldoleal Since the whole multipath.py would need to handle exceptions in most of the functions, can we pursue that in a separate patch ?

PraveenPenguin:
> > @beraldoleal can you please ACK if it looks ok to you without `exception handling` as exception handling we can address in another patch as it requires to adopt in entire multipath lib and we have more discussion on this from the owner of this utils (@narasimhan-v ) if it really needed or not
> 
> IMO it is better if our libraries could deliver a proper high-level exception with detailed information that came from low-level calls.

sure I also agree this point @narasimhan-v can we first add exception handling and later we can have this request should be in

avocado/utils/multipath.py
beraldoleal: @narasimhan-v the whole multipath.py file would be nice for the future, sure. But since that his PR is adding a few methods I'm requesting to add the exceptions only on the methods added here.

harihare:
@beraldoleal Addressed  the review comments for raising the exceptions at the high level itself.

beraldoleal:
> Code Climate has analyzed commit [5fb04eb](https://github.com/avocado-framework/avocado/commit/5fb04eb285c6c84820f07537ecc408c51ed8272e) and detected **2 issues** on this pull request.
> 
> Here's the issue category breakdown:
> Category 	Count
> Duplication 	2
> 
> View more on [Code Climate](https://codeclimate.com/github/avocado-framework/avocado/pull/3595).

I marked those as "won't fix" on code climate, since that is a really small piece of duplication.

avocado/utils/multipath.py
```diff
@@ -25,6 +25,11 @@
 from . import service
 from . import wait
 
+class MPException(Exception):
```
beraldoleal, line 28: I think that Travis is complaining about a missing empty line here.

PR #3597, opened 2020-02-24
Misc Asset: Expose find_asset_file method.

willianrampazzo:
This exposes `find_asset_file` private method and removes the necessity of a parameter to call it, adding the necessary logic to this method itself. Also, it removes duplicated logic from `get_metadata` method after the changes on `find_asset_file`.

This is a pre-work for extending the Asset module to allow the skip of a test if an asset is not available.

### DIFF
--- a/avocado/utils/asset.py
+++ b/avocado/utils/asset.py
@@ -128,37 +128,6 @@ class Asset:
         finally:
             os.remove(temp)
 
-    def find_asset_file(self):
-        """
-        Search for the asset file in each one of the cache locations
-
-        :return: asset file if exists or None
-        :rtype: str or None
-        """
-        parsed_url = urllib.parse.urlparse(self.name)
-        basename = os.path.basename(parsed_url.path)
-        cache_relative_dir = self._get_relative_dir(parsed_url)
-        relative_dir = os.path.join(cache_relative_dir, basename)
-
-        for cache_dir in self.cache_dirs:
-            cache_dir = os.path.expanduser(cache_dir)
-            asset_file = os.path.join(cache_dir, relative_dir)
-
-            # To use a cached file, it must:
-            # - Exists.
-            # - Be valid (not expired).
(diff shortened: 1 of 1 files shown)


willianrampazzo:
> This pull request **introduces 1 alert** when merging [b0b8d8c](https://github.com/avocado-framework/avocado/commit/b0b8d8c6c9d30d2d9d544965648cc69196b8e6ba) into [bda9890](https://github.com/avocado-framework/avocado/commit/bda98905d288b789166a03a8e4ec2efa5d403f89) - [view on LGTM.com](https://lgtm.com/projects/g/avocado-framework/avocado/rev/pr-82e675b26b25b27a709b099c3caaea71ab897f29)
> 
> **new alerts:**
> 
> * 1 for Variable defined multiple times

Fixed!

avocado/utils/asset.py
```diff
@@ -351,18 +323,46 @@ def fetch(self):
...
+    def find_asset_file(self):
+        """
+        Search for the asset file in each one of the cache locations
+
+        :return: asset file if exists or None
+        :rtype: str or None
+        """
+        parsed_url = urllib.parse.urlparse(self.name)
+        basename = os.path.basename(parsed_url.path)
+        cache_relative_dir = self._get_relative_dir(parsed_url)
```
apahim: `parsed_url`, `basename` and `cache_relative_dir` seem like good candidates for class attributes (with better naming), since we are redefining them the same way in multiple methods.

apahim: not your commit's fault though, so it's ok to overlook that.

willianrampazzo: I noticed it when I submitted the PR, and my unconscious was telling me that there would be a comment about that :D. I already added this change to the list for the next refactoring round. But thanks anyway for pointing that out.

willianrampazzo:
Relates to #3600.

PR #3609, opened 2020-02-27
Extends fetch_asset core method to find assets and cancel the test.

willianrampazzo:

When the user wants just to find the asset in the cache, if the asset is not available, it is possible to cancel the test.

### DIFF
--- a/avocado/core/test.py
+++ b/avocado/core/test.py
@@ -1081,7 +1081,8 @@ class Test(unittest.TestCase, TestData):
         raise exceptions.TestCancel(message)
 
     def fetch_asset(self, name, asset_hash=None, algorithm=None,
-                    locations=None, expire=None):
+                    locations=None, expire=None, find_only=False,
+                    cancel_on_missing=False):
         """
         Method o call the utils.asset in order to fetch and asset file
         supporting hash check, caching and multiple locations.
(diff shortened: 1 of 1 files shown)


willianrampazzo:
Relates to #3600.

avocado/core/test.py
```diff
@@ -1081,7 +1081,8 @@ def cancel(self, message=None):  # pylint: disable=W0221
         raise exceptions.TestCancel(message)
 
     def fetch_asset(self, name, asset_hash=None, algorithm=None,
-                    locations=None, expire=None):
+                    locations=None, expire=None, find_only=False,
+                    cancel_on_missing=False):
```
apahim: makes sense

avocado/core/test.py
```diff
@@ -1093,13 +1094,32 @@ def fetch_asset(self, name, asset_hash=None, algorithm=None,
         :param locations: list of URLs from where the asset can be
                           fetched (optional)
         :param expire: time for the asset to expire
-        :raise EnvironmentError: When it fails to fetch the asset
-        :returns: asset file local path
+        :param find_only: whether it should only look for the asset in the
+                          cache and do not download/move it to the cache.
+        :param cancel_on_missing: whether the test should be canceled if the
+                                  asset was not found in the cache. Takes
+                                  effect only when parameter `find_only` is
```
apahim: I find that useful in any case... any reason to restrict it to only when `find_only` is enabled?

avocado/core/test.py
```diff
@@ -1093,13 +1094,32 @@ def fetch_asset(self, name, asset_hash=None, algorithm=None,
...
         if expire is not None:
             expire = data_structures.time_to_seconds(str(expire))
-        return asset.Asset(name, asset_hash, algorithm, locations,
-                           self.cache_dirs, expire).fetch()
+
+        asset_obj = asset.Asset(name, asset_hash, algorithm, locations,
+                                self.cache_dirs, expire)
+
+        if find_only:
+            asset_path = asset_obj.find_asset_file()
```
apahim: Before, the `find_asset_file` was meant to be used independently. Since @clebergnu suggested it could be a feature of the the `fetch_asset` here, it seems the knowledge of what method to call should be in the `Asset` class itself. I'd just parametrize the `Asset.fetch()` call with the `find_only` option and restrict the scope of the change here.
Anyway, not a big deal. I'm happy when @clebergnu is happy.

avocado/core/test.py
willianrampazzo: In the beginning, I thought it would be better to change as little as possible the behavior of the method, but now that you commented, it makes sense, users may want to cancel the test when `fetch` throws an exception instead of handling it on the test side. I'm going to make the change and send another version.

avocado/core/test.py
willianrampazzo: > Before, the `find_asset_file` was meant to be used independently. Since @clebergnu suggested it could be a feature of the the `fetch_asset` here, it seems the knowledge of what method to call should be in the `Asset` class itself. I'd just parametrize the `Asset.fetch()` call with the `find_only` option and restrict the scope of the change here.

I tried to still keep the Asset class a standalone and free from _test_ terminology module to be used out of a test, keeping each behavior in its own method.

> Anyway, not a big deal. I'm happy when @clebergnu is happy.

That's the objective :D

apahim: Fair enough, thanks!

avocado/core/test.py
apahim: cool!

willianrampazzo:
Closing in favor of #3611.

PR #3617, opened 2020-03-02
Extends fetch_asset core method to find assets and cancel the test. [v3]

willianrampazzo:
    
---

Changes from v2:
- Remove the explicit calls to the cancel exception.
- Fix string formating for the log message.
- Add functional tests.

Changes from v1:

### DIFF
--- a/avocado/core/test.py
+++ b/avocado/core/test.py
@@ -1081,7 +1081,8 @@ class Test(unittest.TestCase, TestData):
         raise exceptions.TestCancel(message)
 
     def fetch_asset(self, name, asset_hash=None, algorithm=None,
-                    locations=None, expire=None):
+                    locations=None, expire=None, find_only=False,
+                    cancel_on_missing=False):
         """
         Method o call the utils.asset in order to fetch and asset file
         supporting hash check, caching and multiple locations.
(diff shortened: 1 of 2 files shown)


avocado/core/test.py
```diff
@@ -1093,13 +1094,39 @@ def fetch_asset(self, name, asset_hash=None, algorithm=None,
...
+            if asset_path is None and cancel_on_missing:
+                self.cancel(missing_asset_message)
+            return asset_path
+
+        try:
+            asset_path = asset_obj.fetch()
+        except OSError as e:
+            if cancel_on_missing:
+                self.cancel(missing_asset_message)
+            raise e
```
apahim: This looks very good. Wondering if we shouldn't have the same effect for missing file on both `find_only=True` and `find_only=False` when `cancel_on_missing` is `False`. As of now, the former gives users a `None`, while the latter gives users an `OSError` exception. Wdyt?

avocado/core/test.py
```diff
@@ -1093,13 +1094,39 @@ def fetch_asset(self, name, asset_hash=None, algorithm=None,
...
+        return asset_path
```
apahim: Please add docs.

avocado/core/test.py
willianrampazzo: This is an interesting point. My only concern about this suggestion is that it would change the actual behavior of this method when `find_only=False` and `cancel_on_missing=False`. Users may, with the current upstream code, wait for an exception on their code when the `fetch_asset` method fails to add the asset to the cache. We always have the option to let the users know we will be changing this behavior in the upcoming release and go for the change. 

Another point is that as the method raises an exception coming from the Asset class when there is a problem fetching the asset, I'm afraid that, maybe, returning `None` masks the problem and confuses the users while debugging when `find_only=False` and `cancel_on_missing=False`.

avocado/core/test.py
willianrampazzo: Okay, will add it in the next version, thanks.

avocado/core/test.py
willianrampazzo: @apahim do you have any thoughts here? @clebergnu?

apahim: I'd avoid breaking the API. As a user, with `cancel_on_missing=False`, I'd expect an `OSError` when the file is not found, no matter if `find_only` is `True` or `False`.

willianrampazzo: Ah, I see. From your previous comment, I thought the suggestion was to return `None` in either case.

Okay, I don't have a broad opinion about returning `None` or throwing `OSError` when `find_only=True`. Thinking like a user, I would prefer to deal with `None` than handling an exception, but that is my option.

beraldoleal: My two cents here:

When you use `None` as return on fails, is harder to see why the method has failed. Because of that, I usually prefer exceptions.

willianrampazzo: Okay, let's go for the exception in v4. Thanks, folks!

willianrampazzo:
Relates to #3600.

willianrampazzo:
Closing in favor of #3620.

PR #3620, opened 2020-03-03
Extends fetch_asset core method to find assets and cancel the test. [v4]

willianrampazzo:
    
---
Changes from v3:
- Changed the behavior of `fetch_asset_file` on Asset class to raise an exception instead of returning `None`.
- Handle the exception from `fetch_asset_file` on `fetch_asset` core method.
- Refactor `get_metadata` to reduce code.
- Add comments.

Changes from v2:
- Remove the explicit calls to the cancel exception.
- Fix string formating for the log message.
- Add functional tests.

Changes from v1:

### DIFF
--- a/avocado/core/test.py
+++ b/avocado/core/test.py
@@ -1081,7 +1081,8 @@ class Test(unittest.TestCase, TestData):
         raise exceptions.TestCancel(message)
 
     def fetch_asset(self, name, asset_hash=None, algorithm=None,
-                    locations=None, expire=None):
+                    locations=None, expire=None, find_only=False,
+                    cancel_on_missing=False):
         """
         Method o call the utils.asset in order to fetch and asset file
         supporting hash check, caching and multiple locations.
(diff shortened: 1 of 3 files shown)


willianrampazzo:
Relates to #3600.

avocado/utils/asset.py
```diff
@@ -361,13 +367,15 @@ def get_metadata(self):
...
-            if os.path.isfile(metadata_file):
-                with open(metadata_file, "r") as f:
-                    metadata = json.load(f)
-                    return metadata
-        return None
+        try:
+            asset_file = self.find_asset_file()
+        except OSError:
+            LOG.info("Metadata not available.")
+            return None
```
willianrampazzo: There is a possible improvement here to change the behavior of `get_metadata` and throw an exception when the metadata is not available to follow the pattern from the class, but this is not related to the objective of this PR, so I'm taking note to change it as soon as #3600 is closed

avocado/core/test.py
```diff
@@ -1093,13 +1094,51 @@ def fetch_asset(self, name, asset_hash=None, algorithm=None,
...
+            # return the path to the asset when fetch succeeded
+            asset_path = asset_obj.fetch()
+            return asset_path
+        except OSError as e:
+            # if there was a problem fetching the asset
+            if cancel_on_missing:
+                # cancel when requested
+                self.cancel(missing_asset_message)
+            # otherwise re-throw OSError
+            raise e
```
apahim: Nice, thanks! Now users will have no surprises when using the new feature :)
Do you plan to de-duplicate the duplicated block?

willianrampazzo: Working on it right now, stay tuned for the next minutes ;)

willianrampazzo:
Closed in favor of #3622.

PR #3625, opened 2020-03-04
Misc Asset: update get_metadata to throw an exception.

willianrampazzo:
Other public methods from Asset class were changed to throw an exception instead of returning `None` when the action expected from the method fails. The only method still returning `None` is `get_metadata`. This changes the behavior to throw an exception instead of returning `None`.

### DIFF
--- a/avocado/utils/asset.py
+++ b/avocado/utils/asset.py
@@ -370,8 +370,7 @@ class Asset:
         try:
             asset_file = self.find_asset_file()
         except OSError:
-            LOG.info("Metadata not available.")
-            return None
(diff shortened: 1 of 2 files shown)


willianrampazzo:
Relates to #3450.

avocado/utils/asset.py
```diff
@@ -370,8 +370,7 @@ def get_metadata(self):
         try:
             asset_file = self.find_asset_file()
         except OSError:
-            LOG.info("Metadata not available.")
-            return None
+            raise OSError("Metadata not available.")
```
beraldoleal: GOOD!

PR #3635, opened 2020-03-06
Asset refactoring round 3

willianrampazzo:
Changes introduced by this third round of refactoring in the Asset class:

- move repeated operations to the constructor.
- change `_get_relative_dir` method to used attribute.
- avoid handling strings on `_get_relative_dir`.
- rename class attribute `parsed_url` and `basename`.
- remove unnecessary attribute `self.cache_relative_dir`.


willianrampazzo:
Relates to #3450.

PR #3638, opened 2020-03-09
utils/network: Big refactoring on network library to improve consistency

beraldoleal:
Multiple pull requests were trying to add more features to this library.
But because of a lack of convention, it was getting hard to understand
this and we were losing usability. Besides a complete rewrite here, this
pull request is a) adding a complete set of methods to handle network
interfaces parameters; b) adding a wait when setting MTU values on
interfaces; and c) better support for IPv6.
    
Thanks for @anascko, @vaishnavibhat, @PraveenPenguin, and @smruti77

### DIFF
--- a/avocado/utils/network/exceptions.py
+++ b/avocado/utils/network/exceptions.py
@@ -1,0 +1,4 @@
+class NWException(Exception):
+    """
+    Base Exception Class for all exceptions
+    """
--- a/avocado/utils/network/hosts.py
+++ b/avocado/utils/network/hosts.py
@@ -1,0 +1,86 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+#
+# Copyright: 2019-2020 Red Hat Inc.
+# Authors : Beraldo Leal <bleal@redhat.com>
+
+"""
+This module provides an useful API for hosts in a network.
+"""
+
+from ..ssh import Session
+
+from .common import _run_command
+from .interfaces import NetworkInterface
+from .exceptions import NWException
+
+
+class Host:
+    """
+    This class represents a Host.
+
+    A host can be local or remote. If you pass port, username, key or password,
+    a connection will attempt to be created.
+
+    During the initialization, all interfaces will be detected and available
+    via `interfaces` attribute.
+
+    So, for instance you could have a local and a remote host::
+
+        local = Host(host='foo', port=22, username='foo', password='bar')
+        remote = Host(host='bar')
+
+    You can iterate over the network interfaces of any host::
+
+        for i in remote.interfaces:
+            print(i.name, i.is_link_up())
+    """
+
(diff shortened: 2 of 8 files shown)


avocado/utils/network/exceptions.py
```diff
@@ -0,0 +1,4 @@
+class NWException(Exception):
```
willianrampazzo: Just wondering if it is possible to have a more specific exception here instead os `Exception`.

apahim:
Hi, thanks for this version.

While testing the happy path on a remote machine, I got:
```
>>> h = Host(host='remote.external.com', username='root', port=22, password=password)
>>> for item in h.interfaces:
...     print(item.name)
... 
eth0
lo
wlan0
```
which is good. Exercising a sad path, with an invalid port, I got:

```
>>> h = Host(host='remote.external.com', username='root', port=-1, password=password)
>>> h.interfaces
[]
```

which is also good. Exercising another sad path, also with an invalid port, I got:

```
>>> h = Host(host='remote.external.com', username='root', port=0, password=password)
>>> for item in h.interfaces:
...     print(item.name)
... 
br-7c16cfa0b398
docker0
enp0s31f6
ens1u1
lo
tun0
virbr0
virbr0-nic
wlp0s20f3
```
Which are my local machine interfaces. Is that expected?

beraldoleal:
> Which are my local machine interfaces. Is that expected?
No is not, I will do a double check. Thanks @apahim

avocado/utils/network/exceptions.py
beraldoleal: This is something that was already here. I did not create this.

Can you think of a more specific for all the network problems that we have here?

I don't see any problem with this. NWException says that it is our exceptions, and this is good because the user will know that the problem is with Avocado itself.

avocado/utils/network/interfaces.py
```diff
@@ -0,0 +1,323 @@
...
+        """
+        cmd = "ping -I {} {} -c {}".format(self.name, peer_ip, count)
+        if options is not None:
+            cmd = "{} {}".format(cmd, options)
+        try:
+            _run_command(cmd, self.remote_session)
+        except Exception as ex:
+            raise NWException("Failed to ping: {}".format(ex))
+
+    def save(self, ipaddr, netmask):
```
anascko: @beraldoleal 
Hi Beraldo,
I just have a small question.
In case a system uses NetworkManager maybe we should check it. And after changing the ifcfg-files to run the "nmcli" reload for it?

beraldoleal: @anascko, yes. We can do improvments here. But since that this was already part of the old code. I would suggest to open an issue and try to solve this on a different PR.

avocado/utils/network/hosts.py
```diff
@@ -0,0 +1,86 @@
...
+
+        session = self.remote_session
+        return [NetworkInterface(if_name=name,
+                                 remote_session=session) for name in names]
+
+    def get_interface_by_ipaddr(self, ipaddr):
+        """Return an interface that has a specific ipaddr."""
+        for interface in self.interfaces:
+            if ipaddr in interface.get_ipaddrs():
+                return interface
```
smruti77: @beraldoleal not getting interface name here, can you check it once.

beraldoleal: @smruti77 with this refactoring, we are having access to the object, not only names. You could access the name with `interface.name`.

beraldoleal:
sending a v2.

avocado/utils/network/exceptions.py
willianrampazzo: Okay, no problem, I was just wondering :)

PR #3672, opened 2020-03-24
nrunner: added --parallel-tasks option to nrun [v2]

beraldoleal:
By default, nrun will try to run always in parallel, and sometimes we
need to run the tasks in serial. This will add the --parallel-tasks option to
nrun command.

----
Changes from v1:

  * Before was --serial (bool), now is --parallel-tasks (integer)


beraldoleal:
@clebergnu sorry about that. Just push forced here.

Issue #3674, opened 2020-03-26
getting issue with "ip -4 -j address show interface" command in  interfaces.py.


smruti77:
in few machine getting output with empty dictionary so tests are failing because of this.
for example:

 ip -4 -j address show enP72p1s0
[{},{},{"ifindex":9,"ifname":"enP72p1s0","flags":["BROADCAST","MULTICAST","UP","LOWER_UP"],"mtu":1500,"qdisc":"mq","operstate":"UP","group":"default","txqlen":1000,"addr_info":[{"family":"inet","local":"13.13.13.13","prefixlen":24,"broadcast":"13.13.13.255","scope":"global","noprefixroute":true,"label":"enP72p1s0","valid_life_time":4294967295,"preferred_life_time":4294967295}]},{}]

manvanthar:
hi @beraldoleal,  is this something  you could help with the new network util changes?

Currently its blocking our tests.

smruti77:
@beraldoleal another thing in older distro's "-j" options is not supporting their what we need do.

beraldoleal:
Hi @msruti77 and @manvanthar.

Can you please send the error message (traceback) that you are getting?

beraldoleal:
About the old versions without `-j` support, this was discussed with Cleber and during our meetings:  Avocado #69 is supporting RHEL 7. `-j` it was introduced in Avocado #77. For this version, we support RHEL 8 and beyond (and new fedora/centos versions).

What is the distro that you are running?

smruti77:
@beraldoleal ,"ERROR: argument of type 'NoneType' is not iterable" this is the error what i am getting .its because of ,
we are taking output of "ip -4 -j address show enP72p1s0" with [0].
```
try:
            details = self._get_interface_details(version)
            addr_info = details[0].get('addr_info')
```
but in few machine we are getting output with {} so its failing.
```'ip -4 -j address show enP72p1s0'
[{},{},{"ifindex":9,"ifname":"enP72p1s0","flags":["BROADCAST","MULTICAST","UP","LOWER_UP"],"mtu":1500,"qdisc":"mq","operstate":"UP","group":"default","txqlen":1000,"addr_info":[{"family":"inet","local":"13.13.13.13","prefixlen":24,"scope":"global","label":"enP72p1s0","valid_life_time":4294967295,"preferred_life_time":4294967295}]},{}]
```

smruti77:
[...]

PR #3675, opened 2020-03-27
utils/network: fixed get_interface on different iproute output

beraldoleal:
Some versions of iproute give us a different output when getting
interface details. This small fix will handle properly both output
formats.

### DIFF
--- a/avocado/utils/network/interfaces.py
+++ b/avocado/utils/network/interfaces.py
@@ -56,8 +56,12 @@ class NetworkInterface:
         cmd = "ip -{} -j address show {}".format(version, self.name)
         output = run_command(cmd, self.host)
         try:
-            return json.loads(output)
-        except json.JSONDecodeError:
+            result = json.loads(output)
+            for item in result:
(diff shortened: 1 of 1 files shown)


beraldoleal:
This is related to #3674.

avocado/utils/network/interfaces.py
```diff
@@ -56,8 +56,12 @@ def _get_interface_details(self, version=4):
         cmd = "ip -{} -j address show {}".format(version, self.name)
         output = run_command(cmd, self.host)
         try:
-            return json.loads(output)
-        except json.JSONDecodeError:
+            result = json.loads(output)
+            for item in result:
+                if item.get('ifname') == self.name:
+                    return item
+            raise NWException("Interface not found")
```
willianrampazzo: This exception will be caught by the next line here, is it intended?

beraldoleal: yes. it was. Just to avoid duplicate lines to log.

PR #3679, opened 2020-03-31
hintfile: added the Avocado hint file concept and parser - v2

beraldoleal:
In order to facilitate the Avocado resolution test process, we are introducing the `.avocado.hint` file concept. Basically you must create this file inside your project and Avocado next runner commands will take this "hint file" into consideration. See an example of this file inside conf/avocado/.

This also changes nrun and nlist to look at this file.

##### Changes from v1:

 * Renamed the var name (`$test` -> `$testpath`);
 * Removed restriction on `$testpath`, now you can use in any of `uri`, `args` and `kwargs`;
 * Moved example file to `examples/`
 * Improved the resolution to merge `references` and `.avocado.hint` file

### DIFF
--- a/avocado/core/parser.py
+++ b/avocado/core/parser.py
@@ -18,11 +18,17 @@ Avocado application command line parsing.
 
 import argparse
 
+from configparser import ConfigParser
+from glob import glob
+
 from . import exit_codes
 from . import varianter
 from . import settings
 from .future.settings import settings as future_settings
+from .nrunner import Runnable
 from .output import BUILTIN_STREAMS, BUILTIN_STREAM_SETS, LOG_UI
+from .resolver import ReferenceResolution
+from .resolver import ReferenceResolutionResult
 from .version import VERSION
 
 PROG = 'avocado'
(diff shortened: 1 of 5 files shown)


examples/.avocado.hint.example
```diff
@@ -0,0 +1,29 @@
+# This is just a example file. Insert this file inside your project root
```
clebergnu: There's a problem with this file name, because it won't be seen by most users. I'd recommend creating a directory to host it, and a README file saying that it should be named `.avocado.hint` with emphasis on the *dot*.

avocado/core/parser.py
```diff
@@ -172,3 +178,86 @@ def finish(self):
...
+            key, value = kwarg.split('=')
+            result[key] = value
+        return result
+
+    def _get_resolutions_by_kind(self, kind, paths):
+        self.validate_kind_section(kind)
+
+        resolutions = []
+        uri = self._get_uri_from_section(kind)
+        args = self._get_args_from_section(kind)
```
clebergnu: If there's no args section (a possibility AFAICT), Avocado will crash.

examples/.avocado.hint.example
beraldoleal: @clebergnu do you mean `examples/hint/README` and `examples/hint/.avocado.hint` ?

Actually I was planning to create an entire section on our documentation and remove this example file. What do you think?

clebergnu:
There's something weird going on.  My `.avocado.hint` file is like this:

```ini
[kinds]
tap = ./tests/test-*

[tap]
uri = $testpath
args = --tap
```

The content that I have that I'd like to match:

```
$ find ./tests -name 'test-*' -type f -executable
./tests/test-bitmap
```

First issue is, a "bare" `avocado nlist` won't attempt to find the tests matching `[kinds] tap`, that is, `./tests/test-*`.

Second issue is with a test reference, `avocado nlist` is completely confused:

```
avocado nlist ./tests/test-bitmap
tap ./tests/test-qapi-events.c
```

And the `nrun` behavior follows the `nlist`, as it spawns a the source code file:

 ```
Avocado nrun ./tests/test-bitmap
Results server started at: 127.0.0.1:8888
1-./tests/test-qapi-events.c spawned
Finished spawning tasks
```

File: avocado/plugins/nlist.py [hunk 3]
```diff
@@ -70,7 +71,12 @@ def configure(self, parser):
     def run(self, config):
         references = config.get('nlist.references')
         verbose = config.get('nlist.verbose')
-        resolutions = resolver.resolve(references)
+        hint_filepath = '.avocado.hint'
+        if os.path.exists(hint_filepath):
+            hint = HintParser(hint_filepath)
+            resolutions = resolver.resolve(references, hint)
+        else:
+            resolutions = resolver.resolve(references)
```
willianrampazzo: I find useful to maintain the code, when possible, using single calls to a method. In your case, as `resolutions` is always created calling `resolver.resolve()`, what do you think about a code like:

```
hint =  None
hint_filepath = '.avocado.hint'
if os.path.exists(hint_filepath):
    hint = HintParser(hint_filepath)
resolutions = resolver.resolve(references, hint)
```

It is just a cosmetic suggestion, won't change the logic, so feel free to ignore it.

avocado/plugins/nrun.py
```diff
@@ -170,7 +171,12 @@ def spawn_task(self, task):
             stderr=asyncio.subprocess.PIPE)
 
     def run(self, config):
-        resolutions = resolver.resolve(config.get('nrun.references'))
+        hint_filepath = '.avocado.hint'
+        if os.path.exists(hint_filepath):
+            hint = HintParser(hint_filepath)
+            resolutions = resolver.resolve(config.get('nrun.references'), hint)
+        else:
+            resolutions = resolver.resolve(config.get('nrun.references'))
```
willianrampazzo: Same comment as on `nlist` file.

examples/.avocado.hint.example
clebergnu: Yes, that may be even better.  If users are able to copy and paste (I mean, the right formatting is used) then extra points to you! :)

willianrampazzo: I like the idea of having the template or example file and the documentation. Is it possible to keep both? Is there a downside in making this file visible in the project instead of using an invisible file? I mean, usually, configuration files are visible when they are built by the user.

beraldoleal:
> First issue is, a "bare" `avocado nlist` won't attempt to find the tests matching `[kinds] tap`, that is, `./tests/test-*`.

This is expected based on the current code, but I see, and should not. I will fix on v3.

> Second issue is with a test reference, `avocado nlist` is completely confused:
> 
> ```
> avocado nlist ./tests/test-bitmap
> tap ./tests/test-qapi-events.c
> ```

Ok, I will try to reproduce, but I think that I understood the problem. I will try to fix it on v3.

examples/.avocado.hint.example
beraldoleal: @willianrampazzo keeping both, documentation and example files, might add some extra effort to keep both updated. But maybe we can have an "include" solution, that will use the same file for both.

I will suggest keeping this file as it is, and when the new documentation is added we improve this.

clebergnu:
The include approach has been done before in the docs, so +1.

avocado/plugins/nlist.py
beraldoleal: It is just a style that I follow when I have only 2 calls and also to avoid the `hint = None` line.

But, in this case, I don't see any problem with following your suggestion.

examples/.avocado.hint.example
willianrampazzo: OKay, +1 for the include solution from my side too!

avocado/core/parser.py
beraldoleal: Fixed on v3.

beraldoleal:
Closing in favor of a v3

PR #3687, opened 2020-04-03
Misc Asset: return correct fail during download problems.

willianrampazzo:
When there is a problem downloading an asset, there is no need to remove the temporary file as it was never created. This issue was masking the real download problem.

### DIFF
--- a/avocado/utils/asset.py
+++ b/avocado/utils/asset.py
@@ -132,7 +132,11 @@ class Asset:
                 self._create_hash_file(asset_path)
                 return self._verify_hash(asset_path)
         finally:
-            os.remove(temp)
+            try:
+                os.remove(temp)
+            except FileNotFoundError:
+                LOG.info("Temporary asset file unavailable due to failed"
+                         " download attempt.")
 
     @staticmethod
     def _get_hash_file(asset_path):


willianrampazzo:
> There can be race condition here, between the check and the removal (attempt) the file may be gone. The better way to do this is to use a try/except block catching `FileNotFoundError` and ignoring it.

I thought about the try/except when I started working on it but I thought it would be too much for what was needed and it would duplicate the code already in the `fetch` method.

I will change it on V2. Anyway, do you mind to elaborate a bit more on the race condition? I'm curious about it and I could not catch it with a quick look in the code.

clebergnu:
On Fri, Apr 3, 2020, 10:44 PM Willian Rampazzo <notifications@github.com>
wrote:

> There can be race condition here, between the check and the removal
> (attempt) the file may be gone. The better way to do this is to use a
> try/except block catching FileNotFoundError and ignoring it.
>
> I thought about the try/except when I started working on it but I thought
> it would be too much for what was needed and it would duplicate the code
> already in the fetch method.
>
> I will change it on V2. Anyway, do you mind to elaborate a bit more on the
> race condition? I'm curious about it and I could not catch it with a quick
> look in the code.
>

The race is a possibility simply because the filesystem is a shared
resource, and the file could be deleted by whomever between the "isfile"
check and the attempt to delete it.  It has a small probability of
happening, but still, it's a good practice to avoid those known
possibilities.

willianrampazzo:
> The race is a possibility simply because the filesystem is a shared resource, and the file could be deleted by whomever between the "isfile" check and the attempt to delete it. It has a small probability of happening, but still, it's a good practice to avoid those known possibilities. —

That is true, I took my old operating system books from school yesterday to refresh some concepts and got your point. Thanks for taking the time to send the comment too.

As it is a simple fix, I'll do a force push and let you know.

willianrampazzo:
@clebergnu PR updated, let me know if there is something else needing changes.

avocado/utils/asset.py
```diff
@@ -132,7 +132,11 @@ def _download(self, url_obj, asset_path):
                 self._create_hash_file(asset_path)
                 return self._verify_hash(asset_path)
         finally:
-            os.remove(temp)
+            try:
+                os.remove(temp)
+            except FileNotFoundError:
+                LOG.info("Temporary asset file unavailable due to failed"
```
beraldoleal, line 138: maybe replace LOG.info to LOG.warn ? Not sure.

willianrampazzo: As deleting the temporary file should not impact the test because the download failed anyway and a message about it will be displayed, I though just inform is fine. I would go with a `pass`, but I noticed it is not a good practice.

PR #3691, opened 2020-04-06
Plugin Asset: introduces --ignore-errors to assets fetch command.

willianrampazzo:
When the command `avocado assets fetch` fails to download one of a list of assets, it ends with return code different from 0.

It may be useful to allow the command to end with a 0 return code, independent of any problem it had during the download execution. This allows, for example, Travis CI to continue.

### DIFF
--- a/avocado/plugins/assets.py
+++ b/avocado/plugins/assets.py
@@ -287,6 +287,15 @@ class Assets(CLICmd):
                                  parser=fetch_subcommand_parser,
                                  positional_arg=True)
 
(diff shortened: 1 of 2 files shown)


willianrampazzo:
@clebergnu one thing I was thinking here is if it is enough to set the return code to 0 or if we should also move the messages from stderr to stdout.

willianrampazzo:
@beraldoleal could you take a quick look if I did not mess the command parser? Thanks!

PR #3694, opened 2020-04-07
Improvements to the HTML report

samiraguiar:
Clean up the plugin code following HTML5 standards, turn all tooltips into
Bootstrap popovers, improve escaping of HTML code in the template and
add support for resizable columns.


samiraguiar:
I've included in this PR all the patches we've been using locally.
I didn't squash the commits because I wasn't sure whether you'd appreciate every change, especially the resizable columns part.

If you'd prefer, I could also squash them into one or maybe create separate PRs. Just let me know.

By the way, it would be great to have your input @richtja, as I've touched the custom tooltip code you had added.

samiraguiar:
@clebergnu great! But in case my changes cause any problems, please feel free to ping/mention me :slightly_smiling_face:

PR #3703, opened 2020-04-13
Fix Jinja2 dependency of MarkupSafe to support Python 3.5

willianrampazzo:
The new MarkupSafe 2.0.0 do not support Python 3.5. We need to restrict MarkupSafe version to 1.x.


willianrampazzo:
> PS: Travis is stuck on ppc64le jobs, but this is about Python 3.5, so I consider it safe enough to skip it.

Right, ppc64le is not helping today. Thanks for the review.

PR #3718, opened 2020-04-16
Add support to `requirement` docstring

willianrampazzo:
This adds support to use `requirement` docstrings to define a test requirement in JSON format.

ps. nrunner changes coming next!

### DIFF
--- a/avocado/core/safeloader.py
+++ b/avocado/core/safeloader.py
@@ -20,6 +20,7 @@ Safe (AST based) test loader module utilities
 import ast
 import collections
 import imp
+import json
 import os
 import re
 import sys
@@ -219,7 +220,8 @@ def modules_imported_as(module):
 
 #: Gets the docstring directive value from a string. Used to tweak
 #: test behavior in various ways
-DOCSTRING_DIRECTIVE_RE_RAW = r'\s*:avocado:[ \t]+([a-zA-Z0-9]+?[a-zA-Z0-9_:,\=\-\.]*)\s*$'
+DOCSTRING_DIRECTIVE_RE_RAW = r'\s*:avocado:[ \t]+(([a-zA-Z0-9]+?[a-zA-Z0-9_:,\=\-\.]*)|(r[a-zA-Z0-9]+?[a-zA-Z0-9_:,\=\{\}\"\-\. ]*))\s*$'
+# the RE will match `:avocado: tags=category` or `:avocado requirements={}`
 DOCSTRING_DIRECTIVE_RE = re.compile(DOCSTRING_DIRECTIVE_RE_RAW)
 
 
@@ -277,6 +279,25 @@ def get_docstring_directives_tags(docstring):
                 else:
                     result[tag] = None
     return result
+
+
+def get_docstring_directives_requirements(docstring):
+    """
+    Returns the test requirements from docstring patterns like
+    `:avocado: requirement={}`.
+
+    :rtype: list
+    """
+    requirements = []
+    for item in get_docstring_directives(docstring):
+        if item.startswith('requirement='):
+            _, requirement_str = item.split('requirement=', 1)
+            try:
+                requirements.append(json.loads(requirement_str))
+            except json.decoder.JSONDecodeError:
+                # ignore requirement in case of malformed dictionary
+                continue
+    return requirements
 
 
 def find_class_and_methods(path, method_pattern=None, base_class=None):
--- a/avocado/core/safeloader.py
+++ b/avocado/core/safeloader.py
@@ -342,7 +342,7 @@ def find_class_and_methods(path, method_pattern=None, base_class=None):
     return result
 
 
-def get_methods_info(statement_body, class_tags):
+def get_methods_info(statement_body, class_tags, class_requirements):
     """
     Returns information on an Avocado instrumented test method
     """
@@ -351,12 +351,16 @@ def get_methods_info(statement_body, class_tags):
         if (isinstance(st, ast.FunctionDef) and
                 st.name.startswith('test')):
             docstring = ast.get_docstring(st)
+
             mt_tags = get_docstring_directives_tags(docstring)
             mt_tags.update(class_tags)
 
-            methods = [method for method, _ in methods_info]
+            mt_requirements = get_docstring_directives_requirements(docstring)
+            mt_requirements.extend(class_requirements)
+
+            methods = [method for method, _, _ in methods_info]
             if st.name not in methods:
-                methods_info.append((st.name, mt_tags))
+                methods_info.append((st.name, mt_tags, mt_requirements))
 
     return methods_info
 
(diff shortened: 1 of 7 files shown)


avocado/core/test.py
```diff
@@ -321,7 +321,8 @@ class Test(unittest.TestCase, TestData):
     timeout = None
 
     def __init__(self, methodName='test', name=None, params=None,
-                 base_logdir=None, job=None, runner_queue=None, tags=None):
+                 base_logdir=None, job=None, runner_queue=None, tags=None,
+                 requirements=None):
```
clebergnu, line 325: Passing `tags` to a test was based on a requirement to use tags as "default" values in some situations (the arch on a qemu test, for instance).  I think it's better to keep the test unaware of the requirements, because the whole point is for the test to not even be started if the requirements are missing, and to be able to count on them by default.

Also, given that this is mostly a nrunner related feature, adding this information to the `Avocado.Test` class and to the (older) `avocado.core.loader` code may not be a good idea wrt the support we'll have to add, and then eventually drop.

avocado/core/safeloader.py
```diff
@@ -219,7 +220,8 @@ def modules_imported_as(module):
 
 #: Gets the docstring directive value from a string. Used to tweak
 #: test behavior in various ways
-DOCSTRING_DIRECTIVE_RE_RAW = r'\s*:avocado:[ \t]+([a-zA-Z0-9]+?[a-zA-Z0-9_:,\=\-\.]*)\s*$'
+DOCSTRING_DIRECTIVE_RE_RAW = r'\s*:avocado:[ \t]+(([a-zA-Z0-9]+?[a-zA-Z0-9_:,\=\-\.]*)|(r[a-zA-Z0-9]+?[a-zA-Z0-9_:,\=\{\}\"\-\. ]*))\s*$'
+# the RE will match `:avocado: tags=category` or `:avocado requirements={}`
```
clebergnu, line 224: There's a typo here, missing the `:` right after the second `:avocado` word

selftests/unit/test_safeloader.py
```diff
@@ -260,6 +266,20 @@ def test_tag_keyval_duplicate(self):
         exp = {"fast": None, "arch": set(["x86_64", "ppc64"])}
         self.assertEqual(safeloader.get_docstring_directives_tags(raw), exp)
 
+    def test_get_requirement_empty(self):
```
clebergnu, line 269: These 3 tests (at least) seem all to belong to the previous commit.

willianrampazzo:
>```
> +    :avocado: requirement=[{"rpm": "gcc"}]
>```

This was a topic of discussion during the BP002 RFC. The original proposal was the one that you tried to use, but we decided to go with:

```
:avocado: requirement={"rpm": "gcc"}
```

without using the brackets.

clebergnu:
Oops, I missed that. Thanks!

avocado/core/test.py
clebergnu: @willianrampazzo I thought more about this, and while many use test cases may not need to reference its requirements (for instance, it may need a package `foo` installed, but it will really just trust it's present and execute `foo-cmd`) there are some situations such as having a file downloaded (like the current assets) in which the test will need to reference them.

I don't have the right solution for that, but maybe something like and `id` in the requirements file:

```python
{'url': 'https://host/file.rom', 'id': 'rom'}
```

And on the test:

```python
class Asset(Test):
    def test(self):
        boot_with_rom(os.environ.get(`AVOCADO_REQUIREMENT__ROM`)
```

The point of using an environment variable, rather than adding support for the test is based on:

1) `exec-test` and similars (`tap`) won't have another way of accessing those requirements
2) current `INSTRUMENTED` tests are being kept compatible, but we may want/need to break compatibility when the current runner is replaced.

@beraldoleal do you have comments here?

willianrampazzo: From my point of view, the _requirements resolver_ will be responsible for collecting/fulfilling the requirements for a test. This action, as you previously mentioned, will allow a test to start, depending on the availability of all the requirements. Initially, thinking from a high level, I would not expect the _requirements resolver_ to provide any information to the test (the opposite later in this comment [1]).

Looking at the test execution, how the test will use those requirements is the responsibility of the test, and not the _requirements resolver_. What I mean is, if an asset was downloaded before the test starts by the _requirements resolver_, it is available in the cache. If the test starts, it means the `Asset.find_asset_file()` utility will succeed when looking for the asset.

[1] Now, if the objective is to make the requirements available to the test without any test interaction (e.g., looking for the requirements locally), I agree we can find a way to build a list of `requirement:target` to pass to the test. Still, again, I don't think it is the _requirements resolver_ responsibility to build each detail of this list; on the other hand, it may be its responsibility to consolidate these pieces of information provided by each requirement module.

> The point of using an environment variable, rather than adding support for the test is based on:
> 
> 1. `exec-test` and similars (`tap`) won't have another way of accessing those requirements

In my point of view, this is a new feature for `exec-test` and `tap` test. Today they work without it, looking for its requirements during the test execution or with a pre-test script. I don't think this is something that impacts this PR, but an extension to it. In this case, we have time to discuss possible solutions.

> 2. current `INSTRUMENTED` tests are being kept compatible, but we may want/need to break compatibility when the current runner is replaced.

Right, and it may use the new feature as I commented in 1, but today, I don't see this PR as an impacting point to it.

clebergnu: > From my point of view, the _requirements resolver_ will be responsible for collecting/fulfilling the requirements for a test. This action, as you previously mentioned, will allow a test to start, depending on the availability of all the requirements. Initially, thinking from a high level, I would not expect the _requirements resolver_ to provide any information to the test (the opposite later in this comment [1]).
> 

OK, so what I get from this is that you agree that we should not add a `requirements` parameter and attribute to the `Avocado.Test` class at this point, right?

willianrampazzo: > > From my point of view, the _requirements resolver_ will be responsible for collecting/fulfilling the requirements for a test. This action, as you previously mentioned, will allow a test to start, depending on the availability of all the requirements. Initially, thinking from a high level, I would not expect the _requirements resolver_ to provide any information to the test (the opposite later in this comment [1]).
> 
> OK, so what I get from this is that you agree that we should not add a `requirements` parameter and attribute to the `Avocado.Test` class at this point, right?

Yes, I agree considering this sentence, no, I don't agree considering sentence [1]. :D
It is just a matter of deciding which one we want to follow.

clebergnu: OK, before this gets more confusing, let me just say: I suggest you don't export the requirements to the `Avocado.Test` at this time.  Let me know what you decide on doing.

willianrampazzo: > OK, before this gets more confusing, let me just say: I suggest you don't export the requirements to the `Avocado.Test` at this time. Let me know what you decide on doing.

Okay, this (adding the requirements to the test) is a small change we can do later if we decide to go for it.

What about the loader, as you mentioned in the first comment?

willianrampazzo:
closing in favor of #3725

PR #3721, opened 2020-04-17
calculating and return the size multiple disk

Naresh-ibm:
previously the function use to calculate and return the size
of a single disk. modified it to do the same for multiple disk separated
by space

### DIFF
--- a/avocado/utils/lv_utils.py
+++ b/avocado/utils/lv_utils.py
@@ -36,22 +36,26 @@ class LVException(Exception):
     """
 
 
-def get_diskspace(disk):
+def get_diskspace(disks):
     """
-    Get the entire disk space of a given disk.
+    Get the entire disk space of a given disk(s).
 
-    :param str disk: name of the disk to find the free space of
+    :param str disk: name of the disk(s) to find the free space of
     :returns: size in bytes
-    :rtype: str
-    :raises: :py:class:`LVException` on failure to find disk space
+    :rtype: int
+    :raises: :py:class:`LVException` on failure to find disk space 
     """
-    result = process.run('fdisk -l %s' % disk,
-                         env={"LANG": "C"}, sudo=True).stdout_text
-    results = result.splitlines()
-    for line in results:
-        if line.startswith('Disk ' + disk):
-            return re.findall(r", (.*?) bytes", line)[0]
-    raise LVException('Error in finding disk space')
+    size = 0
+    for disk in disks.split(" "):
+        result = process.run('fdisk -l %s' % disk,
+                             env={"LANG": "C"}, sudo=True).stdout_text
+        results = result.splitlines()
+        for line in results:
+            if line.startswith('Disk ' + disk):
+                size = size + int(re.findall(r", (.*?) bytes", line)[0])
+    if not size:
+        raise LVException('Error in finding disk space')
+    return size
 
 
 def vg_ramdisk(disk, vg_name, ramdisk_vg_size,


Naresh-ibm:
>>> from avocado.utils import lv_utils
>>> 
>>> dir(lv_utils)
['LOGGER', 'LVException', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', 'get_diskspace', 'logging', 'lv_check', 'lv_create', 'lv_list', 'lv_mount', 'lv_reactivate', 'lv_remove', 'lv_revert', 'lv_revert_with_snapshot', 'lv_take_snapshot', 'lv_umount', 'os', 'process', 're', 'shutil', 'thin_lv_create', 'time', 'vg_check', 'vg_create', 'vg_list', 'vg_ramdisk', 'vg_ramdisk_cleanup', 'vg_reactivate', 'vg_remove']
>>> 
>>> disk = '/dev/mapper/mpathc /dev/mapper/mpathb'
>>> 
>>> lv_utils.get_diskspace(disk)
85899345920
>>> 
>>> disk = '/dev/mapper/mpathc'
>>> lv_utils.get_diskspace(disk)
42949672960
>>> 
>>> disk = '/dev/mapper/mpathb'
>>> 
>>> lv_utils.get_diskspace(disk)
42949672960
>>>

avocado/utils/lv_utils.py
```diff
@@ -38,20 +38,21 @@ class LVException(Exception):
 
 def get_diskspace(disk):
     """
-    Get the entire disk space of a given disk.
+    Get the entire disk space of a given disk(s).
 
-    :param str disk: name of the disk to find the free space of
+    :param str disk: name of the disk(s) to find the free space of
```
narasimhan-v, line 43: `name of the disk(s), comma separated, to find the free space of`

avocado/utils/lv_utils.py
```diff
@@ -38,20 +38,21 @@ class LVException(Exception):
...
+    """
+    size = 0
+    for disk in disks.split(" "):
+        result = process.run('fdisk -l %s' % disk,
+                             env={"LANG": "C"}, sudo=True).stdout_text
+        results = result.splitlines()
+        for line in results:
+            if line.startswith('Disk ' + disk):
+                size = size + int(re.findall(r", (.*?) bytes", line)[0])
+    return size
```
narasimhan-v, line 55:
```
if not size:
    raise LVException('Error in finding disk space')
return size
```

avocado/utils/lv_utils.py
Naresh-ibm: @narasimhan-v Abdul said the inputs will be of space separated, can we discus in group and conclude on this?

narasimhan-v: @Naresh-ibm Right, space separated. My mistake.

avocado/utils/lv_utils.py
```diff
@@ -38,20 +38,21 @@ class LVException(Exception):
 
 def get_diskspace(disk):
```
harish-24, line 39: Shouldn't this be changed to `disks`?

Naresh-ibm: @harish-24 Thanks for it... missed it while copying..

narasimhan-v:
@beraldoleal That would be a much better solution, except for the case that `vg_ramdisk()` also uses `get_diskspace()`. We would either need to handle the 2 different functions there as well or go with this solution. Please suggest.

beraldoleal:
@narasimhan-v you can change the internals of `vg_ramdis()` to use the new function without any problem or keep calling the `get_diskspace()`, since that will not breake the API.

We can keep both, `get_diskspace()` and `get_diskspaces()`, IMO.

Naresh-ibm:
@narasimhan-v @beraldoleal 
I did not get you guys, how this is going to affect the existing API.

earlier : give a disk name to this function and it used to return the size of the given disk.

current: give a disk/Disks to the function and it returns the  size of disk/Disks.

===>>> here if multiple disks are given, it will calculate the size of each and add all and return the total size.

if you give single disk, the function is still does the same thing which it used to do earlier.
Hence,
if your existing API is passing a single disk and getting the size, the current function also does the it without any issue.

please correct me if I need to aware of any other information or missing anything.

willianrampazzo:
> @narasimhan-v @beraldoleal
> I did not get you guys, how this is going to affect the existing API.

I think what @beraldoleal was mentioning is that you are changing `get_diskspace(disk)` to `get_diskspace(disks)`. If someone uses this module and call `get_diskspace(disk=disk_location)`, your code will break it.

beraldoleal:
> > @narasimhan-v @beraldoleal
> > I did not get you guys, how this is going to affect the existing API.
> 
> I think what @beraldoleal was mentioning is that you are changing `get_diskspace(disk)` to `get_diskspace(disks)`. If someone uses this module and call `get_diskspace(disk=disk_location)`, your code will break it.

@Naresh-ibm , this ^.

Naresh-ibm:
@narasimhan-v @beraldoleal  the existing code was using the location only and we are still using the same logic.
I just added  a for loop by braking the multiple disk location into singles and use it.

Hence the current code also uses the disks=locations instead of disk=location.

we are using the same "fdisk -l disk" command. and this command works only with location and not with just disk name.

I feel if some API is had some hard code name also this should not break it.
The function does simple thing,  give a disk location or location it gives you the size of it.
Hence I feel there is no chance of hard coding something from this in any API (AFAIK)

willianrampazzo:
> I feel if some API is had some hard code name also this should not break it.

@Naresh-ibm ^ here is the problem, the argument of this function is "hardcoded" as you mentioned. If you change the parameter name from `disk` to `disks` you break the test of other users that are calling it with passing the keyword argument as `disk=location`.

> The function does simple thing, give a disk location or location it gives you the size of it.
> Hence I feel there is no chance of hard coding something from this in any API (AFAIK)

As I mentioned, the internals does not matter, what matters are the function parameters and the return of it.

Naresh-ibm:
> > I feel if some API is had some hard code name also this should not break it.
> 
> @Naresh-ibm ^ here is the problem, the argument of this function is "hardcoded" as you mentioned. If you change the parameter name from `disk` to `disks` you break the test of other users that are calling it passing the keyword argument as `disk=location`.
>
@willianrampazzo The called function parameter name can be anything (AFAIK). As for as Framework or any self tests are using this name as hard code value, I don't think others users has the chance of hard coding it.

Do you guys has any test that is hard coded this name?
 
> > The function does simple thing, give a disk location or location it gives you the size of it.
> > Hence I feel there is no chance of hard coding something from this in any API (AFAIK)
> 
> As I mentioned, the internals does not matter, what matters are the function parameters and the return of it.

===>> if you guys has a strong reason, then I can go ahead change it to disk back use different names internally or if you guys suggest to go for different function, that can also be done.
please let me know which options you want me to go ahead.

willianrampazzo:
> @willianrampazzo The called function parameter name can be anything (AFAIK). 

I agree with you when you are creating a function you can use the parameter names you would like, but this is not the case, you are changing an already existing function with a known parameter.

> As for as Framework or any self tests are using this name as hard code value, I don't think others users has the chance of hard coding it.

Here you are inferring that every single user of Avocado Framework in the world would never use the keyword argument, which cannot be validated.

> 
> Do you guys has any test that is hard coded this name?
> 

No, but we would need to check with every other user that uses Avocado, that is the reason why @beraldoleal suggested you add a deprecation message if you really would like to go with this approach.

> ===>> if you guys has a strong reason, then I can go ahead change it to disk back use different names internally or if you guys suggest to go for different function, that can also be done.
> please let me know which options you want me to go ahead.

Changing the parameter naming is a strong reason. @beraldo already gave you the suggestions here https://github.com/avocado-framework/avocado/pull/3721#pullrequestreview-395354411 and here https://github.com/avocado-framework/avocado/pull/3721#issuecomment-615453390.

Naresh-ibm:
Hi @willianrampazzo @beraldoleal 

we are planning to keep the function paramter value as it previously and continue with the same function name as it was, so that new functionality of multiple disks handling is added without any loss of existing functionality. let me know your thought on same.

beraldoleal:
> Hi @willianrampazzo @beraldoleal
> 
> we are planning to keep the function paramter value as it previously and continue with the same function name as it was, so that new functionality of multiple disks handling is added without any loss of existing functionality. let me know your thought on same.

We need to keep the readability as good as possible. And a good name convention is important. Have a "singular" param that is in fact a multiple values is not a good convention, IMO.

What I'm missing here? I did not understood why not change the method name as I mention. What is the problem with that approach?

narasimhan-v:
@beraldoleal There is no problem with your suggestion, just that we thought to avoid deprecation if possible.

So, yes, I think we can go with `get_disks_total_space(disks)` and add a deprecation message

Naresh-ibm:
closing this in favour of #3792

PR #3732, opened 2020-04-23
future/settings: Migrate resultsdb optional plugin to use the new API

beraldoleal:
code, this change only migrates the resultsdb optional plugin options to
use the new module. This will associate --resultsdb-api,
--resultsdb-logs, and --resultsdb-note-limit to
plugins.resultsdb.api_url, plugins.resultsdb.logs_url and
plugins.resultsdb.note_size_limit respectively on the configuration file.

### DIFF
--- a/optional_plugins/resultsdb/avocado_resultsdb/__init__.py
+++ b/optional_plugins/resultsdb/avocado_resultsdb/__init__.py
@@ -22,7 +22,7 @@ import time
 import resultsdb_api
 
 from avocado.core.plugin_interfaces import CLI, ResultEvents, Result
-from avocado.core.settings import settings
+from avocado.core.future.settings import settings
 from avocado.core.output import LOG_UI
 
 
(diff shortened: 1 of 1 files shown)


optional_plugins/resultsdb/avocado_resultsdb/__init__.py
```diff
@@ -156,11 +151,13 @@ class ResultsdbResult(Result):
...
-                job.config.get('stdout_claimed_by', None) is None):
+        resultsdb_logs = job.config.get('plugins.resultsdb.logs_url')
+        stdout_claimed_by = job.config.get('stdout_claimed_by')
+        if (resultsdb_logs is not None and stdout_claimed_by is None):
             log_msg = "JOB URL    : %s/%s"
-            LOG_UI.info(log_msg, job.config.get('resultsdb_logs',
-                                                os.path.basename(job.logdir)))
+            LOG_UI.info(log_msg,
+                        resultsdb_logs,
+                        os.path.basename(job.logdir))
```
willianrampazzo, line 160: This seems to be a code refactor, maybe it deserves its own commit.

beraldoleal: What part? I'm noot sure, just a small refactoring to fit in 80 columns using the new logic.

willianrampazzo: Right, this is not related to this commit message. That is why I suggested to move it to another commit.

willianrampazzo: Anyway, it does not have an impact on the module, so no problem keeping it this way.

PR #3750, opened 2020-04-28
software_manager.py: Fix tolerant option with yum-builddep

harish-24:
software_manager.py: Add build dependency for DnfBackend

As --tolerant option is not supported with dnf-builddep, adding build dependency method to DnfBackend to fix this.

### DIFF
--- a/avocado/utils/software_manager.py
+++ b/avocado/utils/software_manager.py
@@ -725,6 +725,22 @@ class DnfBackend(YumBackend):
         """
         super(DnfBackend, self).__init__(cmd='dnf')
 
+    def build_dep(self, name):
+        """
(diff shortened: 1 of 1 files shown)


harish-24:
> Well, `yum-utils` still supports `--tolerant`, but the fact is that `yum-utils` is deprecated:
>
> And `dnf-utils` should be used instead. In `software_manager.DnfBackend` we are still relying on `yum` to implement basically everything that should really be based on `dnf`. @harish-24 could you try the following instead and let me know if that works for you?
> 

@clebergnu Yes, I agree this is a better solution. I will push with the suggested change. Thanks for this suggestion.

PR #3777, opened 2020-05-05
Add method which can be use for restore for interface file from backup

PraveenPenguin:
As we need to revert back the interface file when test completed
so added utility can be used for same

### DIFF
--- a/avocado/utils/network/interfaces.py
+++ b/avocado/utils/network/interfaces.py
@@ -314,3 +314,29 @@ class NetworkInterface:
         except Exception as ex:
             msg = 'Failed to remove ipaddr. {}'.format(ex)
             raise NWException(msg)
+
+    def move_interface_file(self, ignore_missing=True):
+        """
+        Revert interface file from backup
+
+        This method checks if a backup version  is available for given interface
+        then it copies backup file to interface file in /sysfs path
+        """
+
+        current_distro = distro_detect()
+
+        filename = "ifcfg-{}".format(self.name)
+        if current_distro.name in ['rhel', 'fedora']:
+            path = "/etc/sysconfig/network-scripts"
+        elif current_distro.name == 'SuSE':
+            path = "/etc/sysconfig/network"
+        else:
+            msg = 'Distro not supported by API. Could not save ipaddr.'
+            raise NWException(msg)
+
+        backup_file = "{}.backup-{}".format(filename, time.time())
+        if os.path.exists(backup_file):
+            shutil.move(backup_file, filename)
+        else:
+            if not ignore_missing:
+                raise NWException("%s interface not available" % self.name)


avocado/utils/network/interfaces.py
```diff
@@ -314,3 +314,29 @@ def remove_ipaddr(self, ipaddr, netmask):
         except Exception as ex:
             msg = 'Failed to remove ipaddr. {}'.format(ex)
             raise NWException(msg)
+
+    def move_interface_file(self, ignore_missing=True):
```
beraldoleal, line 318: I suggest renaming this to `restore_from_backup()`.

Also, silent operations are not good practices. If the user doesn't mind of that, a good alternative would be to raise a proper exception incase of the backup file does not exists, and the user will handle the exception.

avocado/utils/network/interfaces.py
```diff
@@ -314,3 +314,29 @@ def remove_ipaddr(self, ipaddr, netmask):
...
+        """
+        Revert interface file from backup
```
beraldoleal, line 320: All methods here in this module have the summary line on the line right after the method and are ending with a 'dot' for proper documentation rendering. Let's try to keep following this pattern here.

avocado/utils/network/interfaces.py
```diff
@@ -314,3 +314,29 @@ def remove_ipaddr(self, ipaddr, netmask):
...
+
+        current_distro = distro_detect()
+
+        filename = "ifcfg-{}".format(self.name)
+        if current_distro.name in ['rhel', 'fedora']:
+            path = "/etc/sysconfig/network-scripts"
+        elif current_distro.name == 'SuSE':
+            path = "/etc/sysconfig/network"
+        else:
+            msg = 'Distro not supported by API. Could not save ipaddr.'
```
beraldoleal, line 334: "... Could not restore the backup file"

avocado/utils/network/interfaces.py
```diff
@@ -314,3 +314,29 @@ def remove_ipaddr(self, ipaddr, netmask):
...
+            raise NWException(msg)
+
+        backup_file = "{}.backup-{}".format(filename, time.time())
+        if os.path.exists(backup_file):
+            shutil.move(backup_file, filename)
```
beraldoleal, line 339: This logic makes no sense to me: You are moving a backup_file that will never exist (because of the time.time()).

Also during a few PRs ago, we removed the `time.time()` logic in order to avoid having a dir full of files.

avocado/utils/network/interfaces.py
```diff
@@ -314,3 +314,29 @@ def remove_ipaddr(self, ipaddr, netmask):
...
+        else:
+            if not ignore_missing:
+                raise NWException("%s interface not available" % self.name)
```
beraldoleal, line 342: "Backup file not available, could not restore file."

Also, check my previous comment about `ignore_missing`. IMO, we don't need it.

avocado/utils/network/interfaces.py
PraveenPenguin: ah sure i miss that

avocado/utils/network/interfaces.py
PraveenPenguin: ah miss that

beraldoleal: @PraveenPenguin just to be clear, we don't need the `ignore_missing` flag, but still it is good to have a `NWException` always when the file is not found.

PraveenPenguin:
@beraldoleal thanks for review have next version on #3778  incorporated all review comment please have a look

PR #3789, opened 2020-05-07
Docs: fix make html warnings.

willianrampazzo:
Multiple warnings during `make html` for the documentation.


avocado/core/nrunner.py
```diff
@@ -872,7 +872,7 @@ def get_commands(self):
         """
         Return the command names, as seen on the command line application
 
-        For every method whose name starts with "command_", a the name of
+        For every method whose name starts with "command", and the name of
```
willianrampazzo, line 875: Escape the underline with `\` triggers a pylint warning. I tried other options without success, so I decided to remove the underline as the process is detailed some lines below.

PR #3792, opened 2020-05-08
calculating and return the size multiple disk

Naresh-ibm:
The function "get_diskspace()" calculate the size of a single disk.
deprecating this function intead of replacing it with new function.
And adding a new function get_disks_total_space(), which returns the
size of multiple disks combinely.

### DIFF
--- a/avocado/utils/lv_utils.py
+++ b/avocado/utils/lv_utils.py
@@ -23,7 +23,7 @@ import os
 import re
 import shutil
 import time
-
+import warnings
 from . import process
 
 
@@ -45,6 +45,8 @@ def get_diskspace(disk):
     :rtype: str
     :raises: :py:class:`LVException` on failure to find disk space
     """
+    warnings.warn("deprecated, use get_device_total_space instead",
+                  DeprecationWarning)
     result = process.run('fdisk -l %s' % disk,
                          env={"LANG": "C"}, sudo=True).stdout_text
     results = result.splitlines()
@@ -52,6 +54,39 @@ def get_diskspace(disk):
         if line.startswith('Disk ' + disk):
             return re.findall(r", (.*?) bytes", line)[0]
     raise LVException('Error in finding disk space')
+
+
+def get_device_total_space(disk):
+    """Get the total device size.
+
+    :param str device: name of the device/disk to find the total size
+    :returns: size in bytes
+    :rtype: int
+    :raises: :py:class:`LVException` on failure to find disk space
+    """
+    result = process.run('fdisk -l %s' % disk,
+                         env={"LANG": "C"}, sudo=True).stdout_text
+    results = result.splitlines()
+    for line in results:
+        if line.startswith('Disk ' + disk):
+            return int(re.findall(r", (.*?) bytes", line)[0])
+    raise LVException('Error in finding disk space')
+
+
+def get_devices_total_space(devices):
+    """Get the total size of given device(s)/disk(s).
+
+    :param list devices: list with the names of devices separated with space.
+    :returns: sizes in bytes
+    :rtype: int
+    :raises: :py:class:`LVException` on failure to find disk space
+    """
+    size = 0
+    for device in devices:
+        size = size + get_device_total_space(device)
+    if not size:
+        raise LVException('failed to get disks size')
+    return size
 
 
 def vg_ramdisk(disk, vg_name, ramdisk_vg_size,


Naresh-ibm:
closing this in favour of #3792

avocado/utils/lv_utils.py
```diff
@@ -54,6 +55,27 @@ def get_diskspace(disk):
...
+    :raises: :py:class:`LVException` on failure to find disk space
+    """
+    size = 0
+    for disk in disks.split(" "):
+        result = process.run('fdisk -l %s' % disk,
+                             env={"LANG": "C"}, sudo=True).stdout_text
+        results = result.splitlines()
+        for line in results:
+            if line.startswith('Disk ' + disk):
+                size = size + int(re.findall(r", (.*?) bytes", line)[0])
```
narasimhan-v, line 73: Please raise an exception here, or at least store the disk(s) for which size is not got and mention those in exception message

avocado/utils/lv_utils.py
```diff
@@ -54,6 +55,27 @@ def get_diskspace(disk):
     raise LVException('Error in finding disk space')
 
 
+def get_disks_total_space(disks):
+    """
+    Get the entire disk space of a given disk(s).
+    :param str disk: name of the disk(s) to find the free space of
```
narasimhan-v, line 61: separated by space

avocado/utils/lv_utils.py
```diff
@@ -54,6 +55,27 @@ def get_diskspace(disk):
...
+    :returns: size in bytes
+    :rtype: int
+    :raises: :py:class:`LVException` on failure to find disk space
+    """
+    size = 0
+    for disk in disks.split(" "):
+        result = process.run('fdisk -l %s' % disk,
+                             env={"LANG": "C"}, sudo=True).stdout_text
+        results = result.splitlines()
+        for line in results:
```
narasimhan-v, line 71: `for line in result.splitlines():`

beraldoleal:
> closing this in favour of #3792

???

avocado/utils/lv_utils.py
```diff
@@ -54,6 +55,27 @@ def get_diskspace(disk):
...
+def get_disks_total_space(disks):
+    """
+    Get the entire disk space of a given disk(s).
+    :param str disk: name of the disk(s) to find the free space of
+    :returns: size in bytes
+    :rtype: int
+    :raises: :py:class:`LVException` on failure to find disk space
+    """
+    size = 0
+    for disk in disks.split(" "):
```
beraldoleal, line 67: You should assume that `disks` is a list, don't need split.

Naresh-ibm:
> > closing this in favour of #3792
> 
> ???

instead of closing older PR, I closed this, hence re-opend it. so ignore this please.

Naresh-ibm:
> I think that we can make this method much simpler by calling the specific one (for one disk).
> 
> ```python
> def get_device_total_space(device):
>     """Get the total device size.
> 
>     :param str device: name of the device/disk to find the total size
>     :returns: size in bytes
>     :rtype: str
>     :raises: :py:class:`LVException` on failure to find disk space
>     """
>     try:
>         with open('/sys/class/block/{}/size'.format(device), 'r') as fp: 
>             blocks = int(fp.read())
>             return blocks * 512 
>     except FileNotFoundError:
>         raise LVException('Device not found.')
> 
> 
> def get_devices_total_space(devices):
>     """Get the total device size of each device in a list.
> 
>     :param list disks: list with the names of disks.
>     :returns: sizes in bytes
>     :rtype: generator
>     :raises: :py:class:`LVException` on failure to find disk space
>     """
>     for device in devices:
>         yield get_device_total_space(device)
> ```
> 
> IMO, this is a much more clean way to get the disk size. _**Also**_ since that we are creating new functions let's take the opportunity to review the names.
> 
> According to the [source code](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/linux/types.h?id=v4.4-rc6#n121), _"Linux always considers sectors to be 512 bytes long independently of the devices real block size."_
> 
> But, if for any reason, you don't feel comfortable using `/sys/class` no worries, we can keep using `fdisk`.



> I think that we can make this method much simpler by calling the specific one (for one disk).
> 
> ```python
> def get_device_total_space(device):
>     """Get the total device size.
> 
>     :param str device: name of the device/disk to find the total size
>     :returns: size in bytes
>     :rtype: str
>     :raises: :py:class:`LVException` on failure to find disk space
>     """
>     try:
>         with open('/sys/class/block/{}/size'.format(device), 'r') as fp: 
>             blocks = int(fp.read())
>             return blocks * 512 
>     except FileNotFoundError:
>         raise LVException('Device not found.')
>


you mean we will replace the get_diskspace(disk) for single disks with this function and use the below function for multiple disk?

 
> 
> def get_devices_total_space(devices):
>     """Get the total device size of each device in a list.
> 
>     :param list disks: list with the names of disks.
>     :returns: sizes in bytes
>     :rtype: generator
>     :raises: :py:class:`LVException` on failure to find disk space
>     """
>     for device in devices:
>         yield get_device_total_space(device)
> ```
> 
> IMO, this is a much more clean way to get the disk size. _**Also**_ since that we are creating new functions let's take the opportunity to review the names.
> 
> According to the [source code](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/linux/types.h?id=v4.4-rc6#n121), _"Linux always considers sectors to be 512 bytes long independently of the devices real block size."_
> 
> But, if for any reason, you don't feel comfortable using `/sys/class` no worries, we can keep using `fdisk`.

Naresh-ibm:
@beraldoleal  after a long discussion and based on your suggestion in #3721 I have sent this Patch.

beraldoleal:
> you mean we will replace the get_diskspace(disk) for single disks with this function and use the below function for multiple disk?

Yes, but keeping the old one with the deprecation message.

beraldoleal:
> @beraldoleal after a long discussion and based on your suggestion in #3721 I have sent this Patch.

Yes, I remember the comments that I made there, and my comments here are the same on #3721. Since that we are touching this, we could take this opportunity to improve this library and use better names/args, since that it is not so complicated.

avocado/utils/lv_utils.py
```diff
@@ -54,6 +56,38 @@ def get_diskspace(disk):
...
+def get_devices_total_space(devices):
+    """Get the total size of given device(s)/disk(s).
+
+    :param list devices: list with the names of devices separated with space.
+    :returns: sizes in bytes
+    :rtype: generator
+    :raises: :py:class:`LVException` on failure to find disk space
+    """
+    size = 0
+    for device in devices.split(" "):
```
beraldoleal, line 84: Maybe you missed my comment here https://github.com/avocado-framework/avocado/pull/3792#discussion_r422107637. You don't need a split, assume that this is a list already.

avocado/utils/lv_utils.py
```diff
@@ -54,6 +56,38 @@ def get_diskspace(disk):
...
+    try:
+        with open('/sys/class/block/{}/size'.format(device), 'r') as fp:
+            blocks = int(fp.read())
+            return blocks * 512
+    except FileNotFoundError:
+        raise LVException('Device not found.')
+
+
+def get_devices_total_space(devices):
+    """Get the total size of given device(s)/disk(s).
```
beraldoleal, line 76: Well, my initial idea with a method to get "multiple disk spaces" was to return the "size of each disk".

But now I got your intention. We can go with this.

avocado/utils/lv_utils.py
```diff
@@ -54,6 +56,38 @@ def get_diskspace(disk):
...
+
+    :param list devices: list with the names of devices separated with space.
+    :returns: sizes in bytes
+    :rtype: generator
+    :raises: :py:class:`LVException` on failure to find disk space
+    """
+    size = 0
+    for device in devices.split(" "):
+        size = size + int(get_device_total_space(device))
```
beraldoleal, line 85: Unless I'm missing something, you don't need to convert to int, right?

`get_device_total_space()` is already returning a `int`.

avocado/utils/lv_utils.py
```diff
@@ -54,6 +56,38 @@ def get_diskspace(disk):
     raise LVException('Error in finding disk space')
 
 
+def get_device_total_space(device):
+    """Get the total device size.
+
+    :param str device: name of the device/disk to find the total size
+    :returns: size in bytes
+    :rtype: str
```
beraldoleal, line 64: This method return an `int`, right?

avocado/utils/lv_utils.py
```diff
@@ -54,6 +56,38 @@ def get_diskspace(disk):
...
+    except FileNotFoundError:
+        raise LVException('Device not found.')
+
+
+def get_devices_total_space(devices):
+    """Get the total size of given device(s)/disk(s).
+
+    :param list devices: list with the names of devices separated with space.
+    :returns: sizes in bytes
+    :rtype: generator
```
beraldoleal, line 80: The return type here is not a generator anymore, is an `int`, right?

avocado/utils/lv_utils.py
Naresh-ibm: @beraldoleal  hope this is clear now or you are still has concern on splinting devices here.

beraldoleal: Hi @@Naresh-ibm:

No, it is not clear, yet, so yes, I do have some concerns.

We are in Python, this is a python library and whenever possible, we try to use `lists` instead of `strings` when referring to "list of things". I don't see a reason to use `"sda sdb sdc"` instead of `['sda', 'sdb', 'sdc']`.

We are creating a new method here and it is the perfect opportunity to organize the things here.

We have to design this library thinking on multiple users, not only one specific use case. A different user could think that is better to have a "list of things" as a string but comma-separated: `"sda,sdb,sdc"`. How to make all users happy?

In doubt, we try to use the natural way of the language to express a list, whenever possible.

If the user A, have the data as a string, it is super easy to call this as a list:

```python
total = get_devices_total_space(devices.split())
```

Why are you forcing a list of things to be a string?

Naresh-ibm:
@beraldoleal @narasimhan-v this is logic will not work for multipath disk as input, so will change back to old logic where we used 
'fdisk -l <disk_path>'
'
def get_device_total_space(device):
    """Get the total device size.

    :param str device: name of the device/disk to find the total size
    :returns: size in bytes
    :rtype: str
    :raises: :py:class:`LVException` on failure to find disk space
    """
    try:
        with open('/sys/class/block/{}/size'.format(device), 'r') as fp: 
            blocks = int(fp.read())
            return blocks * 512 
    except FileNotFoundError:
        raise LVException('Device not found.') '

please let me know  if any concern.

beraldoleal:
> @beraldoleal @narasimhan-v this is logic will not work for multipath disk as input, so will change back to old logic where we used
> 'fdisk -l <disk_path>'

It would be nice if you could explain why not, so I can understand. I didn't get you.

Naresh-ibm:
> > @beraldoleal @narasimhan-v this is logic will not work for multipath disk as input, so will change back to old logic where we used
> > 'fdisk -l <disk_path>'
> 
> It would be nice if you could explain why not, so I can understand. I didn't get you. Are you talking about fdisk -l or about the list vs string problem?


@beraldoleal 
the input for 'get_diskspace(disk)' is a disk path (the path is /dev/XXX/XXX).
if a user wants to pass a multipath disk like '/dev/mapper/mpathX' then the logic we are trying to use here will fail as the disk name 'mpathX ' is not present /sys/ dir path.

That is the reason the code was developed with fdisk -l /dev/sdX initially .

the /sys/ dir path will be like a static path and few of disk may not show up here(like mpathX from San storage).

we are back to square one.
with this situation if we write two function 'get_device_total_space()' and get_devices_total_space(), the code in 'get_device_total_space()' will be  a duplicate of 'get_diskspace(disk)', as we have to use the old logic due to above mentioned concern.

still we can go ahead with both new function as we are planning to depprecate the 'get_diskspace(disk)' , but  with the same old logic (using fdisk) if every agrees on it  and it is like same function with different name in which one will be deprecated after few days. 

hope you got the point here.

let me know if you still left with any question.

Thanks
Naresh

beraldoleal:
>  with this situation if we write two function 'get_device_total_space()' and get_devices_total_space(), the code in 'get_device_total_space()' will be a duplicate of 'get_diskspace(disk)', as we have to use the old logic due to above mentioned concern.

Ok, @Naresh-ibm thanks for the explanation. As I said on a previous comment, I'm ok using the fdisk logic.

But this thread is about the `list` vs `string`, that I would like to ear some objections (if any) of using `lists`.

Naresh-ibm:
> > with this situation if we write two function 'get_device_total_space()' and get_devices_total_space(), the code in 'get_device_total_space()' will be a duplicate of 'get_diskspace(disk)', as we have to use the old logic due to above mentioned concern.
> 
> Ok, @Naresh-ibm thanks for the explanation. As I said on a previous comment, I'm ok using the fdisk logic.
> 
> But this thread is about the `list` vs `string`, that I would like to ear some objections (if any) of using `lists`.


@beraldoleal   sorry for missing that point, I am ok to use list. I do not have objection on it. will change it to list.
will re-send the patch with changes now.

Naresh-ibm:
@PraveenPenguin @beraldoleal @narasimhan-v please review and merge this if there are no comments.
I have did a force push with changes

avocado/utils/lv_utils.py
narasimhan-v: This maintains the previous functionality, so it is returning `str` like before.

avocado/utils/lv_utils.py
narasimhan-v: `get_diskspace()` was returning `str`, and I think @Naresh-ibm has gone for keeping the same function intact for `get_device_total_space()`

avocado/utils/lv_utils.py
```diff
@@ -54,6 +73,22 @@ def get_diskspace(disk):
...
+def get_devices_total_space(devices):
+    """Get the total size of given device(s)/disk(s).
+
+    :param list devices: list with the names of devices separated with space.
+    :returns: sizes in bytes
+    :rtype: int
+    :raises: :py:class:`LVException` on failure to find disk space
+    """
+    size = 0
+    for device in devices.split():
```
narasimhan-v, line 85: @Naresh-ibm I thought we agreed to make it a list.

narasimhan-v:
@beraldoleal please check if this version is fine with you.

narasimhan-v:
@beraldoleal need your help in this, we have some tests dependent.

beraldoleal:
> @beraldoleal need your help in this, we have some tests dependent.

Sorry for the delay here. I'm just waiting for rpm-build here. I will merge asap.

PR #3809, opened 2020-05-12
Refactor base runner app

willianrampazzo:
This is a small refactor to make the list of commands available as a class attribute avoiding multiple resolutions and string manipulation when a command needs to be resolved.

Also, it changes a list comprehension by a list casting as it is faster.

Ps. this is related to my path to a full review and learning on the Avocado Nrunner architecture. If the code does not make sense, feel free to just close this PR.

### DIFF
--- a/avocado/core/nrunner.py
+++ b/avocado/core/nrunner.py
@@ -824,6 +824,7 @@ class BaseRunnerApp:
             prog = self.PROG_NAME
         if description is None:
             description = self.PROG_DESCRIPTION
+        self._class_commands_method = self._get_commands_method()
         self._setup_parser(prog, description)
 
     def _setup_parser(self, prog, description):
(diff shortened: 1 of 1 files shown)


avocado/core/nrunner.py
```diff
@@ -824,32 +824,32 @@ def __init__(self, echo=print, prog=None, description=None):
...
         subcommands.required = True
-        for cmd_meth in self._get_commands_method_without_prefix():
-            attr = "CMD_%s_ARGS" % cmd_meth.upper()
-            cmd = cmd_meth.replace('_', '-')
+        for cmd, meth in self._class_commands_method.items():
+            attr = "CMD_%s_ARGS" % cmd.upper().replace('-', '_')
             cmd_parser = subcommands.add_parser(
                 cmd,
-                help=self._get_command_method_help_message(cmd_meth))
+                help=self._get_command_method_help_message(meth))
```
beraldoleal, line 839: Maybe a more explicit name `method`?

willianrampazzo: Thanks @beraldoleal, I have updated this and a few other variables in this method.

PR #3817, opened 2020-05-14
Check namespaces exists before command executes

harish-24:
With recent changes to ndctl utility, the behaviour for changed for disabling/destroying namespaces with "all"

Before recent changes, it always iterates idle namespaces
```
$ ndctl disable-namespace all
disabled 2 namespaces
$ echo $?
0
```
Now it returns failure even if idle namespaces are found
```
$ ndctl disable-namespace all
error disabling namespaces: No such device or address
disabled 0 namespaces
$ echo $?
250
```
With this check, command executes only if active namespaces are available

### DIFF
--- a/avocado/utils/pmem.py
+++ b/avocado/utils/pmem.py
@@ -194,6 +194,9 @@ class PMem:
             args = '%s -r %s' % (args, region)
         if bus:
             args = '%s -b %s' % (args, bus)
+        if namespace == 'all':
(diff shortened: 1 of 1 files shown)


harish-24:
@beraldoleal I guess all the lines are needed in this method. If there is a way to simplify please suggest so. Otherwise please consider merging, Thanks for the review.

beraldoleal:
Travis is passing, merging it.

PR #3819, opened 2020-05-14
Nrunner: extend status server checks for user request to exit

willianrampazzo:

Here we check if there is a `bye` in any place of the message, instead of strictly finding only it.

### DIFF
--- a/avocado/core/nrunner.py
+++ b/avocado/core/nrunner.py
@@ -696,7 +696,7 @@ class StatusServer:
                     return True
 
             message = await reader.readline()
-            if message == b'bye\n':
+            if b'bye'in message:
                 print('Status server: exiting due to user request')
                 self.server_task.cancel()
                 await self.server_task


willianrampazzo:
> This causes a problem when any message contains the string 'bye', such as when run:
> 
> ```
> avocado-runner task-run -k exec -u /bin/true -i 1-bye -s 127.0.0.1:8080
> {'status': 'started', 'time': 1589491854.0816019, 'output_dir': '/tmp/.avocado-task-tlrmr2ck', 'id': '1-bye'}
> {'status': 'finished', 'time': 1589491854.0838618, 'returncode': 0, 'stdout': b'', 'stderr': b'', 'id': '1-bye'}
> ```
> 
> The status server will exit, even though it's not intended. Maybe the solution is to strip the message for new lines and carriage returns and check for `b'bye'`.

Yes, I agree. I thought about it but could not think of any solution that was straightforward at the time. I'll change it and send another version.

PR #3827, opened 2020-05-15
Nrunner: extend status server checks for user request to exit [V3]

willianrampazzo:

Changes from [2]:
- `strip` can handle it, no need to use `splitlines`.

Changes from [V1}:
- split the message with `splitlines`. This handles a good amount of separators.


willianrampazzo:
Will add some more commits here.

PR #3829, opened 2020-05-18
Introduces the 'avocado jobs' command with list, show and get-output-files (v2)

beraldoleal:
This is the initial implementation of the 'avocado jobs' command that list all avocado jobs, show details about a specific job and download output files generated by tests. This is a basic start and can be improved in the future. This is helpful to improve usability and to provide a place for the 'get-output-files' introduced on the next commits.

**Changes from v1:**
 * Small refactoring on `handle_output_files_command`;
 * Fixed method exit codes;
 * Using `string_to_safe_path` to replace `/` with `_`;
 * Adding metavar='sub-command';
 * Accepting a "job reference" on `avocado jobs show` command;
 * `latest` will be the default value for `avocado jobs show`;


avocado/plugins/jobs.py
```diff
@@ -0,0 +1,238 @@
...
+                results[job['job_id']] = result
+
+        subcommand = config.get('jobs_subcommand')
+        if subcommand == 'list':
+            return self.handle_list_command(results)
+        elif subcommand == 'show':
+            return self.handle_show_command(config)
+        elif subcommand == 'get-output-files':
+            return self.handle_output_files_command(config)
+        return exit_codes.AVOCADO_ALL_OK
```
willianrampazzo, line 238: I know there are lots of opinions about single versus multiple return statements in a method, but I think we have a good opportunity for readability here using a single return. Check if code here helps:

https://github.com/avocado-framework/avocado/blob/master/avocado/plugins/assets.py#L321-L350

beraldoleal: I think that is questionable. ;) When using the return there it is clear that we don't have to read the lines below, because we know that the code "stops there". So IMO, this is easy to read then introducing a new variable and making the code continue the execution.

willianrampazzo: > I think that is questionable. ;) When using the return there it is clear that we don't have to read the lines below, because we know that the code "stops there". So IMO, this is easy to read then introducing a new variable and making the code continue the execution.

I agree with you if we consider this specific small block of code. For long blocks of code, where sometimes you may just want to fix a bug in the middle of it, having multiple points of return may be more confusing than helping to debug.

So, here it is up to you as the code is small and easy to read in one screen.

beraldoleal:
@clebergnu I did a force-push here, replacing print with LOG_UI, some fancy colors are enabled now.

Fixed also the complexity of a method.

About the multiple `returns` as I explained to @willianrampazzo, I think that is ok as it is.

Please, let me know if you need a v3, or you are ok with this.

beraldoleal:
> LGTM! Waiting for the tabular output in a future PR.

Thanks @willianrampazzo , see my comment here about tabular output:

https://github.com/avocado-framework/avocado/pull/3825#discussion_r426622053

willianrampazzo:
> > LGTM! Waiting for the tabular output in a future PR.
> 
> Thanks @willianrampazzo , see my comment here about tabular output:
> 
> [#3825 (comment)](https://github.com/avocado-framework/avocado/pull/3825#discussion_r426622053)

Yep, that is the reason for my comment about waiting for it on a future PR. :)

beraldoleal:
> > > LGTM! Waiting for the tabular output in a future PR.
> > 
> > 
> > Thanks @willianrampazzo , see my comment here about tabular output:
> > [#3825 (comment)](https://github.com/avocado-framework/avocado/pull/3825#discussion_r426622053)
> 
> Yep, that is the reason for my comment about waiting for it on a future PR. :)

ahha ok, I got it.

PR #3857, opened 2020-05-27
lvutils: Use new get_devices_total_space()

narasimhan-v:
get_device_total_space() and get_devices_total_space() are new
functions, preferred over the "to be deprecated" get_diskspace(),
via fd233dbef312d27d87edbe493dbbd0100782380a.

### DIFF
--- a/avocado/utils/lv_utils.py
+++ b/avocado/utils/lv_utils.py
@@ -140,10 +140,10 @@ def vg_ramdisk(disk, vg_name, ramdisk_vg_size,
 
         LOGGER.debug("Converting and copying /dev/zero")
         if disk:
-            vg_size = get_diskspace(disk)
+            vg_size = get_devices_total_space(disk.split())
 
         # Initializing sparse file with extra few bytes
-        cmd = ("dd if=/dev/zero of=%s bs=1M count=1 seek=%s" %
+        cmd = ("dd if=/dev/zero of=%s bs=1M count=1 seek=%d" %
                (ramdisk_filename, vg_size))
         process.run(cmd)
         if not disk:


avocado/utils/lv_utils.py
```diff
@@ -140,10 +140,10 @@ def vg_ramdisk(disk, vg_name, ramdisk_vg_size,
 
         LOGGER.debug("Converting and copying /dev/zero")
         if disk:
-            vg_size = get_diskspace(disk)
+            vg_size = get_devices_total_space(disk.split())
```
beraldoleal, line 143: It is strange the use of `.split()` in this particular situation. Based on the docstring of this method:

```
    :param str disk: name of the disk in which volume groups are created
```
I'm expecting a single disk, so when you do a split over a string it will create for instance `['s', 'd', 'a']` or something like that, right?

Maybe you should use the method that receives only one disk or call like that:

```python
    vg_size = get_devices_total_space([disk])
```
What do you think?

narasimhan-v: @beraldoleal On my tries on both python2.7 and 3, have seen `split()`s behaviour only defaulting to space. 

```
>>> "abc".split()
['abc']
>>> 
```

narasimhan-v: @beraldoleal This is via the `vg_ramdisk()` function, and it can work with one or multiple disks (as seen in https://github.com/avocado-framework-tests/avocado-misc-tests/pull/1754). So, handling it in a generic way, treating even a single disk as loop.

narasimhan-v: We can have a check and then call either `get_device_total_space()` or `get_devices_total_space()`, but I do not see any significant advantage over this method.

beraldoleal: To be honest, I don't like this approach. If the method says that is ***one*** disk we should handle one disk, Tests should not assume that the code will handle multiple disk.

If we need to handle multiple disks here, then we should change the signature and docstrings here too. (`disks` as list, instead of `disk`).

narasimhan-v: I get your point. Shall I change the docstring of `vg_ramdisk()` ?

beraldoleal: So, if we are going on this path, we should do the complete job here: not only the docstring but also change the signature method. This will imply also a deprecation message and new method.

But we should take a breath and think if this makes sense or not.

For instance, `vg_ramdisk()` is a method/function that doesn't say much about what is doing. We should choose better names.

Also, I don't like the idea of having an argument that can accept string or list. We will end up with multiple codes:

```python
vg_ramdisk(['sda', 'sdb'], ...)
vg_ramdisk('sda', ...)
```

This tends to increase the complexity of the code in the future and also give us headaches during debug time. If we are going to accept multiple disks lets use **only** lists.

Maybe it is time to create a new method here, as we did previously, doing our best effort on making this right (better method name, a nice method signature, and a nice docstring). 

I know that sounds a rough review, and I apologize for that, I'm just trying to have a good code at the end that will be consistent and also have a good that it is easier to maintain.

narasimhan-v: I understand your concern @beraldoleal. This function needs redesign, just that we can not have tests failing because of this. If, hypothetically, fixing does take longer than expected, is it possible to have this as a temporary solution, and then have a proper solution implemented in a week or so ? I ask this because I am not aware of any other uses of this method outside the one test pointed.

beraldoleal: @narasimhan-v I'm here to help and I'm willing to improve this with you, But first, help me to understand something:

Look at the lines 158-161 of the current code:

```python
158     if not disk:
159         loop_device = result.stdout_text.rstrip()
160     else:
161         loop_device = disk
```
How this change that you are proposing will impact here? Will the `loop_device` have comma or spaces? Is that right? What is the name of the volume group when using multiple disks?

narasimhan-v: @beraldoleal 
The main thing to note is, all the commands like lvcreate, vgcreate, etc., work with multiple disks when separated by a space. That is what we are taking advantage of here.

So, from the testcase, we pass disks as a string (single and multiple), and the whole process works fine except when finding disk space. So that was enhanced earlier.

Now, coming to the quoted code in the lines 158-161, we create loop device when disk is not provided. So, it is always 1 device, and the code works fine.

Also, the vg and lv creation works perfectly in case of single or multiple disks, as shown in the referenced test pull request (https://github.com/avocado-framework-tests/avocado-misc-tests/pull/1754)

beraldoleal: @narasimhan-v thanks for the explanation, I also realized that yesterday.

So yes, I have problems to accept this because of the reasons that I explained before. I think that we should avoid ambiguity whenever possible. 

I will try to submit a PR with a new function for review as soon as possible.

narasimhan-v: Sure, thanks @beraldoleal

beraldoleal: @narasimhan-v I was almost done with a proposal of refactoring here, but looking at the current code, I found this:

```python3
        if not disk:
            LOGGER.debug("Creating loop device")
            process.run("losetup %s %s" %
                        (loop_device, ramdisk_filename), sudo=True)
```

So basically, if **there is disk** the loop_device is the disk, and the loop_device will not be created, also pvcreate and vgcreate will not use the `ramdisk_filename`. 

Unless I'm missing something, makes no sense to me to use this method passing `disk` or `disks` because of the current logic (when using disk):

1. Will mount the ramdisk_dir as tmpfs;
2. Will create a `ramdisk`filename` inside `ramdisk_dir` (memory);
**BUT**:
3. PV will be based on disk (not on this ramdisk);
4. The same with VG.

1 and 2 are being created for no reason? Because, afaik, 3 and 4 are using  `loop_device = disk` and not to the loop_device;

So, IIUC, when calling this function with `disk = something`, it is returning a `ramdisk_filename` and a `loop_device` that *was not* used to set up the physical volume and the volume group. Giving to the user the wrong impression.

Did you get? What I'm missing?

narasimhan-v: Actually, when this function is called without a disk name, it creates a loop device, and returns that.

Even I think this is an unwanted way, as you can find in:
https://github.com/avocado-framework-tests/avocado-misc-tests/blob/master/io/disk/lvsetup.py
https://github.com/avocado-framework-tests/avocado-misc-tests/blob/master/io/disk/fiotest.py

The main reason this exists is because, it was ported from autotest, and no one was forced to do anything about it, until recently.

narasimhan-v: A better approach would be similar to the fiotest mentioned, and have the test create a loop device in needed cases. If that is taken care, we can remove the usage of vg_ramdisk().

beraldoleal: @narasimhan-v IMO, when is called without a disk I understand the logic. But when called with a disk (or disks) makes no sense to me to use this method. Because the method name and docstring says that will "Create volume group on top of ram memory to speed up LV performance" and that is not true. Do you agree?

Besides that, there is also an option `use_tmpfs` that if is `False` will not speed up the process, for me this option should be always `true`.

Yes, I completely agree that we should deprecate this function, since that, afaik, is wrong in many ways.

What do you think ?

narasimhan-v: Completely agree. Can I take care of both the test and util function ?

beraldoleal: Just to make sure , the util function (vg_ramdisk()) is just to depracte this method with a warning, right?

And we can close this PR.

narasimhan-v: Right

beraldoleal: @narasimhan-v go for it. ;)

narasimhan-v: @beraldoleal done both, please check.

pevogam: Hi @beraldoleal and @narasimhan-v, I am just chiming in to write down a note for myself to trace back your conversation. As the original writer of this entire code (~2012) I can explain any questions you have. Thanks for your patience and for not being in a hurry to remove this critical for some of our own users code.

pevogam: So I finally managed to read this conversion in depth and see your conclusions. I completely agree with all conclusions that @beraldoleal made here, for me he fell victim of the fact that this function has slowly mutated with inappropriate changes that took it completely out of context. There are two most significant points I have to explain here:

1) The `disk` argument was added years ago by someone on top of the original `vg_ramdisk` function which was always meant to create the volumes on top of a tmpfs-based ramdisk. I think the person simply wanted to reuse the function for their own disk creation, something we can now clearly see is hacky, twisted the meaning of the function without updating its signature or docstring, and added confusing about using one-vs-many disks. I tried to dig into the git history to restore my memory but the closest commit I could find is [this migration](https://github.com/avocado-framework/avocado/commit/aa163a5a61f9135049df17d918c118a0c103ec77) from Autotest. However, if you see the [version](https://github.com/autotest/autotest/blob/2a6beda50eb0012ce059be18ff2344e8005e4f93/client/lv_utils.py#L46) of this function in Autotest itself, it doesn't have this complication. The change was done somewhere in between the two repos but seems impossible to find where and from whom as the authors of the commit introducing the LV utils might have simply copied it from somewhere without being the authors of these changes (btw @beraldoleal this is one reason why I insisted on [migrating the session ops code in aexpect](https://github.com/avocado-framework/aexpect/pull/54) in a clean way and work on top - to avoid debugging difficulties like these ^^).

2) The `use_tmpfs` argument was added by me [here](https://github.com/autotest/autotest/commit/1a3d2a537c676bce46bf77309ce34de8ddeab009) back in 2017 since creating the volume group on top of ramdisk was starting to drain our RAM too much and SSD-s became much faster. This one is entirely my fault in minimizing the amount of changes while disabling big parts of this function. Your comment about this switch contradicting the docsting is completely justified.

Now that these two major topics in your comments and their origin might be clearer, I am not sure what is the best way to move on. I think deprecating it is probably about right because even though it was quite useful functionality for many years, disks are much faster now and the functionality remains too niche for the general LV utilities. So I do understand and basically support your shared decisions.

@narasimhan-v There are many more tests using this function and other functions from the LV utils but these are not public. So assuming no tests are using this based on public records of open source projects is not enough, there are a lot of in-house CI environments that do and could do.

Sorry both for taking quite some time to respond here, I hope I have made things at least a bit clearer for you.

PR #3858, opened 2020-05-28
small fix on Makefile: removed missing target

beraldoleal:
make requirements is an invalid target, right now. This small change
just remove the instruction on makefile.


beraldoleal:
> Line 120 is also calling it. Maybe it is a good opportunity to fix that.
> 
> https://github.com/avocado-framework/avocado/blob/8a25ef64d2ce8675ea338657e4f68adb132e5c11/Makefile#L120

Good catch. Fixed with a force-push here.

PR #3886, opened 2020-06-04
NRunner: Support requirements into runnables

willianrampazzo:


willianrampazzo:
> Hi @willianrampazzo . The code looks good to me, but I'd like to see tests for the serialization of requirements, and better yet, the collection of requirements from a source (say docstrings) to the Task instance and then to the resulting representations (command line args, recipe). If you think there's already test coverage (I couldn't assert it), let me know.

Hi @clebergnu, this was just a preparation for the requirements resolver, should not have a real impact to the execution, that is the reason I have not added tests. As we reached the release freeze, I'll leave this change for the next release.

PR #3918, opened 2020-06-15
Removed deprecated selftests on old settings.py

beraldoleal:
The new model will ignore the registration of options via config file
only. All options need to be configured by `register_option()`. Besides
that we have already a test_future_settings.


beraldoleal:
I'm closing this, this in fact should be removed when settings.py is removed.

PR #3925, opened 2020-06-16
sysinfo.collect.enabled: True -> on

narasimhan-v:
enabled parameter is parsed as on / off, but it is True / False
in the avocado.conf file. This issue is exposed only if per_test
is False.

Reported-by: Satheesh Rajendran <sathnaga@linux.vnet.ibm.com>

### DIFF
--- a/avocado/etc/avocado/avocado.conf
+++ b/avocado/etc/avocado/avocado.conf
@@ -14,7 +14,7 @@ logs_dir = ~/avocado/job-results
 
 [sysinfo.collect]
 # Whether to collect system information during avocado jobs
-enabled = True
+enabled = on
 # Overall timeout to collect commands, when <=0 no timeout is enforced
 commands_timeout = -1
 # Whether to take a list of installed packages previous to avocado jobs


narasimhan-v:
https://github.com/avocado-framework/avocado/blob/master/avocado/plugins/sysinfo.py#L122

avocado/etc/avocado/avocado.conf
```diff
@@ -14,7 +14,7 @@ logs_dir = ~/avocado/job-results
 
 [sysinfo.collect]
 # Whether to collect system information during avocado jobs
-enabled = True
+enabled = on
```
beraldoleal, line 17: @narasimhan-v, 

Actually, AFAIK, this is not necessary as `register_option()` is saving the default values now.

There is a new PR (#3934) to remove the templates files. Can you please check if this will work when there is no avocado.conf files?

narasimhan-v: @beraldoleal Yes, this will work when there is no conf files

beraldoleal: Ok, so I think that we can close this and wait for #3934 to be approved. What do you think?

narasimhan-v: Sure, closing.

PR #3933, opened 2020-06-16
future.settings: added a 'reference' option to the 'config' command-line

beraldoleal:
This will add a new command "avocado config reference" that will print
on output the current configuration reference guide in rst format. And
can be used to update our documentation.  This PR also change our
documentation to include this reference. There is room for improvements,
for instance we can soon split the configuration into sections and also
give the ability to download a config file template.

### DIFF
--- a/avocado/plugins/config.py
+++ b/avocado/plugins/config.py
@@ -11,6 +11,9 @@
 #
 # Copyright: Red Hat Inc. 2013-2014
 # Author: Lucas Meneghel Rodrigues <lmr@redhat.com>
+#         Beraldo Leal <bleal@redhat.com>
+
+import textwrap
 
 from avocado.core import data_dir
 from avocado.core.future.settings import settings as future_settings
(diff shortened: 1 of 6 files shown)


avocado/plugins/config.py
```diff
@@ -39,7 +42,32 @@ def configure(self, parser):
                                         parser=parser,
                                         long_arg='--datadir')
 
+        subcommands = parser.add_subparsers(dest='config_subcommand',
+                                            metavar='sub-command')
+
+        help_msg = 'Show a configuration reference with all registered options'
+        subcommands.add_parser('reference', help=help_msg)
```
willianrampazzo, line 49: As the code is displaying the configs on the terminal, I would go with a more intuitive command name, like `show`, `avocado config show` feel more natural to me.

Anyway, this is not a big deal.

beraldoleal: @willianrampazzo IMO, `avocado config show` gives the impression that will show the current config. And it is not the case. Maybe `show-reference` but I'm trying to keep commands short.

willianrampazzo: Reading the PR description I thought the command was supposed to show the current config options, independent from where they come. 

What happens if I execute `avocado --config /tmp/foo.conf config reference`? I imagine it will use the internal default options, plus the updated options from `/tmp/foo.conf`. This is the reason I thought `show` would make more sense initially.

beraldoleal: **current configuration reference guide** maybe my bad English gave that impression. Sorry. Is the "current reference guide"

PR #3939, opened 2020-06-17
future.settings: brings back the previous parsing method for lists

beraldoleal:
This is using the ast module to eval lists as lists in python (using
single or double quotation marks). In the near future we can replace
this eval with a more inteligent and save method.


beraldoleal:
@clebergnu this is just bringing back the previous method so we can move on. But we can improve this soon.

PR #3974, opened 2020-06-30
core: Introduces the TestSuite() class and small refactoring

beraldoleal:
This is an attempt to organize and prepare the code to support multiple suites in a single Job. Each TestSuite will have its own configuration.

### DIFF
--- a/avocado/core/job.py
+++ b/avocado/core/job.py
@@ -338,7 +338,7 @@ class Job:
         """
         def soft_abort(msg):
             """ Only log the problem """
-            LOG_JOB.warning("Unable to update the latest link: {}", msg)
+            LOG_JOB.warning("Unable to update the latest link: %s", msg)
         basedir = os.path.dirname(self.logdir)
         basename = os.path.basename(self.logdir)
         proc_latest = os.path.join(basedir, "latest.%s" % os.getpid())
--- a/avocado/core/job.py
+++ b/avocado/core/job.py
@@ -451,23 +451,8 @@ class Job:
             suite[i] = [DryRunTest, suite[i][1]]
         return suite
 
-    @staticmethod
-    def _resolver_check_missing_references(references, resolutions):
-        missing = []
-        for reference in references:
-            results = [res.result for res in resolutions if
-                       res.reference == reference]
-            if ReferenceResolutionResult.SUCCESS not in results:
-                missing.append(reference)
-        if missing:
-            msg = "Could not resolve references: %s" % ",".join(missing)
-            raise exceptions.JobTestSuiteReferenceResolutionError(msg)
-
     def _make_test_suite_resolver(self, references, ignore_missing):
-        resolutions = resolve(references)
-        if not ignore_missing:
-            self._resolver_check_missing_references(references,
-                                                    resolutions)
+        resolutions = resolve(references, ignore_missing)
         return resolutions_to_tasks(resolutions, self.config)
 
     def _log_job_id(self):
--- a/avocado/core/job.py
+++ b/avocado/core/job.py
@@ -27,15 +27,14 @@ import tempfile
 import time
 import traceback
 
-from ..utils import astring, path, process, stacktrace
+from ..utils import astring, path, process
 from ..utils.data_structures import CallbackRegister, time_to_seconds
 from . import (data_dir, dispatcher, exceptions, exit_codes, jobdata,
-               loader, nrunner, output, result, tags, varianter, version)
+               output, result, varianter, version)
 from .job_id import create_unique_job_id
 from .future.settings import settings
 from .output import LOG_JOB, LOG_UI, STD_OUTPUT
-from .test import DryRunTest
-from .resolver import resolve
+from .suite import TestSuite, TestSuiteError
 
 _NEW_ISSUE_LINK = 'https://github.com/avocado-framework/avocado/issues/new'
 
@@ -365,49 +364,6 @@ class Job:
             if os.path.exists(proc_latest):
                 os.unlink(proc_latest)
 
-    def _make_test_suite(self, references, ignore_missing):
-        """
-        Prepares a test suite to be used for running tests
-
-        :param references: List of tests references to be resolved and
-                           transformed into test factories
-        :type references: list of str
-        :returns: a test suite (a list of test factories)
-        """
-        if self._test_runner_name == 'nrunner':
-            make_test_suite = self._make_test_suite_resolver
-        else:
-            make_test_suite = self._make_test_suite_loader
-        try:
-            self.test_suite = make_test_suite(references, ignore_missing)
-        except loader.LoaderError as details:
-            stacktrace.log_exc_info(sys.exc_info(), LOG_UI.getChild("debug"))
-            raise exceptions.JobTestSuiteError(details)
-
-    def _make_test_suite_loader(self, references, ignore_missing):
-        loader.loader.load_plugins(self.config)
-        try:
-            suite = loader.loader.discover(references, force=ignore_missing)
-            filter_tags = self.config.get("filter.by_tags.tags")
-            if filter_tags:
-                suite = tags.filter_test_tags(
-                    suite,
-                    filter_tags,
-                    self.config.get("filter.by_tags.include_empty"),
-                    self.config.get('filter.by_tags.include_empty_key'))
-        except loader.LoaderUnhandledReferenceError as details:
-            raise exceptions.JobTestSuiteError(details)
-
-        if not self.config.get('run.dry_run.enabled'):
-            return suite
-        for i in range(len(suite)):
-            suite[i] = [DryRunTest, suite[i][1]]
-        return suite
-
-    def _make_test_suite_resolver(self, references, ignore_missing):
-        resolutions = resolve(references, ignore_missing)
-        return resolutions_to_tasks(resolutions, self.config)
-
     def _log_job_id(self):
         LOG_JOB.info('Job ID: %s', self.unique_id)
         if self.replay_sourcejob is not None:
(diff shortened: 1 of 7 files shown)


avocado/core/job.py
```diff
@@ -547,7 +496,7 @@ def run_tests(self):
         jobdata.record(self.config, self.logdir, variant, sys.argv)
         summary = self.test_runner.run_suite(self,
                                              self.result,
-                                             self.test_suite,
+                                             self.test_suite.tests,
```
willianrampazzo, line 499: I understand this is the introduction of the `TestSuite` class and is the way it is for compatibility. As an evolution, I see the `run_suite` still receiving as an argument the `self.test_suite` and handling whatever is needed inside it.

beraldoleal: Agree.

beraldoleal:
> Despite Travis not liking it so much, I like how you are preparing to handle multiple test suites. From my point of view, the code and the approach looks good. I left one comment about the upcoming changes.

I'm working on it. thanks.

beraldoleal:
@willianrampazzo looks like travis it is happy now. ;)

willianrampazzo:
> @willianrampazzo looks like travis it is happy now. ;)

You rock!

avocado/core/job.py
```diff
@@ -35,58 +34,12 @@
...
 from .output import LOG_JOB, LOG_UI, STD_OUTPUT
-from .resolver import ReferenceResolutionResult, resolve
-from .tags import filter_test_tags_runnable
 from .test import DryRunTest
+from .resolver import resolve
 
 _NEW_ISSUE_LINK = 'https://github.com/avocado-framework/avocado/issues/new'
 
 
-def resolutions_to_tasks(resolutions, config):
```
clebergnu, line 45: You moved `resolutions_to_tasks()` to `avocado.core.utils` here, but it's still referenced in `_make_test_suite_resolver()`.

avocado/core/job.py
```diff
@@ -35,58 +34,12 @@
...
-        if resolution.result != ReferenceResolutionResult.SUCCESS:
-            continue
-        for runnable in resolution.resolutions:
-            if filter_by_tags:
-                if not filter_test_tags_runnable(runnable,
-                                                 filter_by_tags,
-                                                 include_empty,
-                                                 include_empty_key):
-                    continue
-            tasks.append(nrunner.Task(str(uuid.uuid1()), runnable,
```
clebergnu, line 85: With this line, `nrunner` does not need to be imported here anymore.

avocado/core/job.py
```diff
@@ -494,25 +450,17 @@ def _log_job_debug_info(self, variants):
...
-            else:
-                e_msg = ("No test references provided nor any other arguments "
-                         "resolved into tests. Please double check the "
-                         "executed command.")
-            raise exceptions.JobTestSuiteEmptyError(e_msg)
-
-        self.result.tests_total = len(self.test_suite)
+        try:
+            self.test_suite = TestSuite.from_config(self.config,
+                                                    runner=self._test_runner_name)
```
clebergnu, line 455: You're usually looking for a smaller number of arguments, so I have to ask: any reason for letting `from_config` itself find out what the runner is?  IMO, this would minimize the interface changes once the current runner is dropped, and hopefully we'll only have one runner (the "next" runner, with multiple spawners).

avocado/core/job.py
```diff
@@ -338,7 +338,7 @@ def _update_latest_link(self):
         """
         def soft_abort(msg):
             """ Only log the problem """
-            LOG_JOB.warning("Unable to update the latest link: {}", msg)
+            LOG_JOB.warning("Unable to update the latest link: %s", msg)
```
clebergnu, line 341: I'm having a hard time trying to understand why the previous commit caused the linter to flag this (I've bisected it to a8f60853ef0975587f5de7c445d256f3d13286f8).

Anyway, if we understand it or not, I'd still either incorporate this into the previous commit, or change the order so that future bisections running `make check` would not falsely flag a8f60853ef0975587f5de7c445d256f3d13286f8.

avocado/core/job.py
beraldoleal: There is no reason. :) You are right, makes sense. Will change on a v2.

avocado/core/suite.py
```diff
@@ -0,0 +1,141 @@
...
+        # config
+        self.config = settings.as_dict()
+        if config:
+            self.config.update(config)
+
+        self._variant = None
+        self._references = None
+
+        if config.get('run.dry_run.enabled'):
+            self.convert_to_dry_run()
```
clebergnu, line 40: It's expected that some Job API users will interact with `TestSuite`, so I believe this method should not be public at this time, given that it's limited to the loader implementation.  Also, it's being called indiscriminately here, so it causes a crash if the resolver implementation is used.  Sample job:

```python
#!/usr/bin/env python3

import sys

config = {'run.references': ['examples/tests/passtest.py:PassTest.test'],
          'run.test_runner': 'nrunner',
          'run.dry_run.enabled': True}

with Job(config) as j:
    sys.exit(j.run())
```

```
$ ./examples/jobs/passjob.py 

Avocado crashed: TypeError: 'Task' object is not subscriptable

  File "/home/cleber/src/avocado/avocado/avocado/core/job.py", line 455, in create_test_suite
    runner=self._test_runner_name)

  File "/home/cleber/src/avocado/avocado/avocado/core/suite.py", line 103, in from_config
    return cls.from_config_with_resolver(config, name)

  File "/home/cleber/src/avocado/avocado/avocado/core/suite.py", line 141, in from_config_with_resolver
    tests=tasks)

  File "/home/cleber/src/avocado/avocado/avocado/core/suite.py", line 40, in __init__
    self.convert_to_dry_run()

  File "/home/cleber/src/avocado/avocado/avocado/core/suite.py", line 92, in convert_to_dry_run
    self.tests[i] = [DryRunTest, self.tests[i][1]]

TypeError: 'Task' object is not subscriptable

Please include the traceback info and command line used on your bug report
Report bugs visiting https://github.com/avocado-framework/avocado/issues/new
```

avocado/core/suite.py
```diff
@@ -0,0 +1,141 @@
...
+                   "provided nor any other arguments resolved into tests")
+            raise TestSuiteError(msg)
+
+        if runner == 'nrunner':
+            return cls.from_config_with_resolver(config, name)
+        else:
+            return cls.from_config_with_loader(config, name)
+
+    @classmethod
+    def from_config_with_loader(cls, config, name=None):
```
clebergnu, line 108: *Maybe*, in accordance to the comment on the `convert_to_dry_run()` not being a public interface, this (and `from_config_with_resolver()`) should also become private.  Just an idea as we evolve the interface and think about the usability aspects.

avocado/core/suite.py
```diff
@@ -0,0 +1,141 @@
...
+    def size(self):
+        """The overall length/size of this test suite."""
+        if self.tests is None:
+            return 0
+        return len(self.tests)
+
+    @property
+    def status(self):
+        if self.tests is None:
+            return TestSuiteStatus.RESOLUTION_NOT_STARTED
```
clebergnu, line 75: I like the idea of a status, I only miss some tests on the concept and implementation.

beraldoleal: I took this idea from this comment: 

https://github.com/avocado-framework/avocado/blob/837883c264eb195e4be742fb5b2d05f548178b20/avocado/core/job.py#L175

So maybe I don't have a clear use case yet, just added there for future use based on this comment. Not sure if we should keep it or not.

avocado/core/suite.py
```diff
@@ -0,0 +1,141 @@
...
+from .utils import resolutions_to_tasks
+from .varianter import Varianter
+
+
+class TestSuiteError(Exception):
+    pass
+
+
+class TestSuiteStatus(Enum):
+    RESOLUTION_NOT_STARTED = 0
```
clebergnu, line 19: Nitpick: when we don't have a reason for manually settings values, such as mapping external interfaces, I find it convenient to just use an `object()` to guarantee uniqueness (because `auto` is only available on 3.6 and later https://docs.python.org/3/library/enum.html#enum.auto).

avocado/core/suite.py
clebergnu: I think it's valid information (I'd like to have it as a user of a test suite), but you're right that it may not be an urgent need.  I'll leave it to you to keep it (and add some basic tests) or drop it.

avocado/core/suite.py
beraldoleal: I like the numbers because sometimes it is easy to debug when the number it is showing instead of the string.

clebergnu:
Don't you have to access `.value` to get to,  well, the value?

beraldoleal:
> Don't you have to access `.value` to get to, well, the value?

Yes, but for debugging, I like to see a small number instead of an object pointer. IMO, `1` it is better than `<object at 0x7f5d50cac450>` and it is easier to associate the number to the Enum.

In any case, not a big deal here, I can change to `object()` since that it is already used in other parts of the code.

avocado/core/job.py
beraldoleal: For some reason when calling `TestSuite.from_config(self.config)` without the `runner`, the error msg disappeared. :-O
Makes no sense to me.

I will revert this.

avocado/core/suite.py
beraldoleal: @clebergnu thanks for this comment. I got some errors after a review. Sending a v2 with a some tests here.

beraldoleal:
Thank you @willianrampazzo and @clebergnu for your review. Closing this in favor of a v2.

PR #3976, opened 2020-07-03
passwordless scp add function to ssh.py

smruti77:
add scp_connection() and scp_cmd() to ssh.py for copy file to
remote and from remote through session.

### DIFF
--- a/avocado/utils/ssh.py
+++ b/avocado/utils/ssh.py
@@ -29,6 +29,12 @@ except path_utils.CmdNotFoundError:
     SSH_CLIENT_BINARY = None
 
 
+class NWException(Exception):
+    """
+    Base Exception Class for all exceptions
+    """
+
+
 class Session:
     """
     Represents an SSH session to a remote system, for the purpose of
@@ -214,3 +220,32 @@ class Session:
         :rtype: bool
         """
         return self._master_command('exit')
+
+    def copy_files(self, source, destination, recursive=False):
+        """
+        Copy Files to and from remote through scp session.
+
+        :param source: Source file
+        :type: str
+        :param destination: Destination file location
+        :type: str
+        :param recursive: Scp option for copy file. if set to True
+                          copy files inside directory recursively.
+        :type: bool
+        :returns: True if success and an exception if not.
+        :rtype: bool
+        """
+        try:
+            cmd = path_utils.find_command('scp')
+        except path_utils.CmdNotFoundError as exc:
+            raise exc
+        options = self._dash_o_opts_to_str(self.DEFAULT_OPTIONS)
+        if recursive:
+            options += ' -r'
+        options += " {} {}".format(source, destination)
+        try:
+            result = process.run("{} {}".format(cmd, options),
+                                 ignore_status=True)
+            return result.exit_status == 0
+        except process.CmdError as exc:
+            raise NWException("failed to copy file {}".format(exc))


avocado/utils/ssh.py
```diff
@@ -214,3 +214,37 @@ def quit(self):
...
+    def scp_connection(self):
+        try:
+            SCP_CLIENT_BINARY = path_utils.find_command('scp')
+        except path_utils.CmdNotFoundError:
+            SCP_CLIENT_BINARY = None
+        cmd = self._dash_o_opts_to_str(self.DEFAULT_OPTIONS)
+        cmd = "%s %s" % (SCP_CLIENT_BINARY, cmd)
+        return cmd
+
+    def scp_cmd(self, source, destination, reverse=False, recursive=False):
```
beraldoleal, line 227: IMO, we could have a better name for this method. Something like: `copy` or `copy_files` for instance.

Also, I think that the `reverse` option is confusing because of the name of the args. If the user would like a "reverse" behavior, let the user change the order of args during the call. (i.e: `copy_files("/tmp/foo/bar/*", "192.168.0.1:/tmp/foo/bar/", True)`)

avocado/utils/ssh.py
```diff
@@ -214,3 +214,37 @@ def quit(self):
...
+        """
+        copy file to and from remote through scp session
```
beraldoleal, line 229: This docstring will be displayed in our documentation, so could you please upper the first letter and also finish the sentence with a dot?

avocado/utils/ssh.py
```diff
@@ -214,3 +214,37 @@ def quit(self):
...
+        :type: str
+        :param reverse: wether to copy to or from remote.if set
+                        to False copy to remote. and it True copy
+                        from remote.
+        :type: bool
+        :param recursive: scp option for copy file. if set to True
+                          copy files inside directory recursively.
+        :type: boot
+        """
+        options = self.scp_connection()
```
beraldoleal, line 242: IIUC, this method is just creating the command. So maybe `scp_connection()` isn't a good name. Also, if you handling the `CmdNotFoundError` and setting the command to `None` on `scp_command()` method, you are letting the exception be raised on the `process.run()` what makes me think: why handle that there? 

So maybe we don't need the `scp_connection()` method here. My suggestion would be something, like this:

```python
def copy_files(self, source, destination, recursive=False):
     cmd = path_utils.find_command('scp')
     options = self._dash_o_opts_to_str(self.DEFAULT_OPTIONS)
     if recursive:
        options += ' -r'
     options += "{} {}".format(source, destination)
     result = process.run("{} {}".format(cmd, options), ignore_status=True)
```
(I didn't test this code, just an idea/example)

Or you could use the same example here and put `process.run()` inside a try.. except to handle the exception here.

What do you think?

avocado/utils/ssh.py
smruti77: @beraldoleal ok, i will change like this.

avocado/utils/ssh.py
smruti77: @beraldoleal ok, i will change like this.

avocado/utils/ssh.py
smruti77: ok

avocado/utils/ssh.py
```diff
@@ -214,3 +214,28 @@ def quit(self):
...
+        :type: str
+        :param destination: Remote file location
+        :type: str
+        :param recursive: Scp option for copy file. if set to True
+                          copy files inside directory recursively.
+        :type: boot
+        :returns: Whether the files copy successfully or not.
+        :rtype: bool
+        """
+        cmd = path_utils.find_command('scp')
```
beraldoleal, line 231: Since `find_command()` can raise a `CmdNotFoundError` exception you could try to capture that exception. Maybe add the find_command() line inside the try.. except to avoid having two excepts...

avocado/utils/ssh.py
```diff
@@ -214,3 +214,28 @@ def quit(self):
         :rtype: bool
         """
         return self._master_command('exit')
+
+    def copy_files(self, source, destination, recursive=False):
+        """
+        Copy Files to and from remote through scp session
```
beraldoleal, line 220: Missing a dot at the end of the line and missing a blank line after the first line.

avocado/utils/ssh.py
```diff
@@ -214,3 +214,28 @@ def quit(self):
...
+        :param source: Local file
+        :type: str
+        :param destination: Remote file location
+        :type: str
+        :param recursive: Scp option for copy file. if set to True
+                          copy files inside directory recursively.
+        :type: boot
```
beraldoleal, line 227: s/boot/bool/

avocado/utils/ssh.py
```diff
@@ -214,3 +214,28 @@ def quit(self):
         :rtype: bool
         """
         return self._master_command('exit')
+
+    def copy_files(self, source, destination, recursive=False):
+        """
+        Copy Files to and from remote through scp session
+        :param source: Local file
```
beraldoleal, line 221: I think that "local" is not accurate since I can copy from a remote host to my local machine with this logic, right?

avocado/utils/ssh.py
```diff
@@ -214,3 +214,28 @@ def quit(self):
...
+        :type: str
+        :param destination: Remote file location
```
beraldoleal, line 223: Same here with the "remote" word.

avocado/utils/ssh.py
```diff
@@ -214,3 +214,28 @@ def quit(self):
...
+        options = self._dash_o_opts_to_str(self.DEFAULT_OPTIONS)
+        if recursive:
+            options += ' -r'
+        options += " {} {}".format(source, destination)
+        try:
+            result = process.run("{} {}".format(cmd, options),
+                                 ignore_status=True)
+            return result.exit_status == 0
+        except process.CmdError as exc:
+            raise ("failed to copy file {}".format(exc))
```
beraldoleal, line 241: Did you execute this code? Because looks like this line have a syntax problem:

```
TypeError: exceptions must be old-style classes or derived from BaseException, not str
```

avocado/utils/ssh.py
```diff
@@ -214,3 +214,28 @@ def quit(self):
...
+        Copy Files to and from remote through scp session
+        :param source: Local file
+        :type: str
+        :param destination: Remote file location
+        :type: str
+        :param recursive: Scp option for copy file. if set to True
+                          copy files inside directory recursively.
+        :type: boot
+        :returns: Whether the files copy successfully or not.
+        :rtype: bool
```
beraldoleal, line 229: True if success and an exception if not.

avocado/utils/ssh.py
smruti77: @beraldoleal but better to have in process.run() too.

avocado/utils/ssh.py
smruti77: yes tested the code, i not getting any syntax error.

avocado/utils/ssh.py
smruti77: > @beraldoleal but if i handle in one try ..except for two others lines also come inside it so better to have separate.

smruti77:
@beraldoleal need to merge this,review please.

avocado/utils/ssh.py
beraldoleal:
```python3
#!/bin/env python3
try:
    a = 10 / 0
except ZeroDivisionError as exc:
    raise ("failed to copy file {}".format(exc))
```
```bash
./test.py
  File "/tmp/test2.py", line 3, in <module>
    a = 10 / 0
ZeroDivisionError: division by zero

During handling of the above exception, another exception occurred:

  File "/tmp/test2.py", line 5, in <module>
    raise ("failed to copy file {}".format(exc))
TypeError: exceptions must derive from BaseException
```

This code it is causing **another** exception on top of the original one. Maybe you need to raise a proper exception:

```python
raise MyCustomException("Failed to copy file {}".format(exc))
```

beraldoleal:
> @beraldoleal need to merge this,review please.

Please see my comments, IMO this still needs some changes.

PR #3983, opened 2020-07-08
Add infoblocks function to pmem library

harish-24:
Patch adds infoblocks function to pmem library

### DIFF
--- a/avocado/utils/pmem.py
+++ b/avocado/utils/pmem.py
@@ -319,3 +319,119 @@ class PMem:
                           shell=True, ignore_status=True):
             raise PMemException('Namespace destroy command failed')
         return True
+
+    @staticmethod
+    def _check_arg(key, kwargs):
+        if key in kwargs and kwargs[key]:
+            return True
+        return False
+
+    @staticmethod
+    def _check_add_arg(args_dict, key, kwargs, pop=False):
+        if PMem._check_arg(key, kwargs):
+            if pop:
+                return " %s" % args_dict[key] % kwargs.pop(key)
+            return " %s" % args_dict[key] % kwargs.get(key)
+        return ""
+
+    @staticmethod
+    def _filter_ns_infoblock(namespace, args_dict, kwargs):
+        args = ""
+        if namespace == "all":
+            for key in ['region', 'bus']:
+                args += PMem._check_add_arg(args_dict, key, kwargs, pop=True)
+        return args
+
+    def write_infoblock(self, namespace='', stdout=False, output=None,
+                        **kwargs):
+        """
+        Write an infoblock to the specified medium.
+
+        :param namespace: Write the infoblock to given namespace
+        :param stdout: Write the infoblock to stdout if True
+        :param output: Write the infoblock to the file path specified
+        :param **kwargs:
+        Example:
+           pmem.write_infoblock(namespace=ns_name, align=align,
+                                size=size, mode='devdax')
+
+        :return: True if command succeeds
+        :rtype: bool
+        :raise: :class:`PMemException`, if command fails.
+        """
+        if not (namespace or stdout or output):
+            raise PMemException("Specify atleast one output medium")
+
+        args_dict = {'region': '-r %s', 'bus': '-b %s', 'mode': '-m %s',
+                     'memmap': '-M %s', 'size': '-s %s', 'align': '-a %s',
+                     'uuid': '-u %s', 'parent_uuid': '-p %', 'offset': '-O %s'}
+
+        if namespace:
+            args = namespace
+        elif stdout:
+            args = "-c"
+        elif output:
+            args = "-o %s" % output
+
+        args += self._filter_ns_infoblock(namespace, args_dict, kwargs)
+        args += " %s" % args_dict['mode'] % kwargs.pop('mode', 'fsdax')
+        args += " %s" % args_dict['memmap'] % kwargs.pop('memmap', 'dev')
+
+        for key, value in kwargs.items():
+            if not value:
+                continue
+            if key in args_dict:
+                args += " %s" % args_dict[key] % value
+            else:
+                raise PMemException("Input not supported for write-infoblock")
+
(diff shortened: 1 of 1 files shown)


harish-24:
@beraldoleal Review would be helpful, Thanks.

avocado/utils/pmem.py
```diff
@@ -319,3 +319,111 @@ def destroy_namespace(self, namespace='all', region='', bus='',
...
+        if not (namespace or stdout or output):
+            raise PMemException("Specify atleast one output medium")
+        args = None
+        if namespace:
+            args = namespace
+        elif stdout:
+            args = " -c"
+        elif output:
+            args = "-o %s" % output
+        if namespace == "all":
```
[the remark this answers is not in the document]
beraldoleal: Hi @harish-24 the first thing that I can think here is the amount of args. As codeclimate pointed I would suggest refactoring this to have fewer arguments. What about have only "args" and the user decide the args to pass?

> Function `write_infoblock` has a Cognitive Complexity of 16 (exceeds 5 allowed). Consider refactoring.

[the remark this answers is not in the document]
harish-24: @beraldoleal Sure, do I need to keep the docstring here for possible inputs?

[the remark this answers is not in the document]
beraldoleal: @harish-24 would be nice to change the docstrings too, and if possible give us some usage examples.

avocado/utils/pmem.py
```diff
@@ -319,3 +319,109 @@ def destroy_namespace(self, namespace='all', region='', bus='',
                           shell=True, ignore_status=True):
             raise PMemException('Namespace destroy command failed')
         return True
+
+    def write_infoblock(self, namespace='', stdout=False, output=None,
+                        **kwargs):
```
beraldoleal, line 324: Well, while introducing `kwargs` it is a nice solution, this does not reduce the complexity of this method. The "Cognitive Complexity" of this method is still too high (18).

What I was thinking about was to introduce a raw "args" parameter, so we can avoid those if ... else.. inside this method. Something like this:

```python
write_infoblock('foo', args="-p foo -a bar -r foo-bar")
```

If you don't like this idea, another option would be to accept a dictionary with the options:

```python
write_infoblock('foo', args={'-p': 'foo', '-a': 'bar', '-r': 'foo-bar'})
```

In that way, you don't need to write the args inside this method. IMO, this would reduce the complexity here.

What do you think?

harish-24: Although this is a solution to reduce complexity, the idea behind using a method from tests it to make steps simple. IMO, this would mean each tests needs to form something like this `'-p': 'foo', '-a': 'bar', '-r': 'foo-bar'` (dict or string) which kind of makes thing redundant (though possible).

Let us say five different tests uses this function, needs to repeat the parameter with each options, which kind of makes method not of much use except for running a single command.

So is there a better solution which we can adopt here?

beraldoleal: I see your point and I understand. Lets try to find a better solution. I can see a few spots on this method that could be removed or are odd at least.

So, do you have the man page link for this ndctl that you are using? Because the one that I found has no bus option and do not mention about region and bus being used only with namespace == 'all'.

harish-24: https://pmem.io/ndctl/ndctl-write-infoblock.html
> The keyword all can be specified to operate on every namespace in the system, optionally filtered by bus id (see –bus= option), or region id (see –region= option).

harish-24: @beraldoleal Comments here would be appreciable, Thanks.

avocado/utils/pmem.py
```diff
@@ -319,3 +319,109 @@ def destroy_namespace(self, namespace='all', region='', bus='',
...
+        """
+        Write an infoblock to the specified medium, either of namespace, stdout
+        and output which are mutually exclusive to each other.
```
beraldoleal, line 327: The summary line should be only one line. Please, see PEP257.

avocado/utils/pmem.py
```diff
@@ -319,3 +319,109 @@ def destroy_namespace(self, namespace='all', region='', bus='',
...
+
+        write_cmd = "%s write-infoblock %s" % (self.ndctl, args)
+        if process.system(write_cmd, shell=True, ignore_status=True):
+            raise PMemException("write-infoblock command failed")
+        return True
+
+    def read_infoblock(self, namespace='', inp_file='', **kwargs):
+        """
+        Read an infoblock from the specified medium, either of namespace or
+        inp_file which are mutually exclusive to each other.
```
beraldoleal, line 380: Same here.

avocado/utils/pmem.py
```diff
@@ -319,3 +319,109 @@ def destroy_namespace(self, namespace='all', region='', bus='',
...
+        elif stdout:
+            args = "-c"
+        elif output:
+            args = "-o %s" % output
+
+        if namespace == "all":
+            if 'region' in kwargs and kwargs['region']:
+                args += " %s" % args_dict[key] % kwargs.pop('region')
+            if 'bus' in kwargs and kwargs['bus']:
+                args += " %s" % args_dict[key] % kwargs.pop('bus')
```
beraldoleal, line 359: This block of code is very similar of what we have on the second method. Could you try to split this and reuse to avoid duplication lines? Maybe we could reduce the complexity here that it is still too high.

harish-24:
@beraldoleal Pushed with repetition as minimal as possible, Please help in reviewing.

harish-24:
@beraldoleal Thanks for the review, pushed fixing the inspekt failure.

PraveenPenguin:
looks me Travis has timed out for one instance so we can make this in 
```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

The build has been terminated
```

PR #3993, opened 2020-07-14
[docs] Configuration reference fix

richtja:
The documentation about configuration references was incomplete. This happened
because, when the documentation is created the settings module with references
is not completely loaded yet. This commit should fix it.

Reference: #3947

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -33,11 +33,12 @@ except path.CmdNotFoundError:
 
 
 def generate_reference():
-    from avocado.plugins.config import Config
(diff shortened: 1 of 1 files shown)


richtja:
@clebergnu I can try to explain this.  `Config` gets the references from `future.settings`, but to store these references to the `Settings` It has to be called function `register_option` for each reference. And the registration for plugins is [here](https://github.com/avocado-framework/avocado/blob/master/avocado/core/app.py#L55). It means that if we don't initialize Avocado, we don't have the references. And that was the problem.

PR #3998, opened 2020-07-16
add function to get admin link state to handle unknown state.

smruti77:
if interface state is unknown,it will check for admin link
state of interfcae and return it.


avocado/utils/network/interfaces.py
```diff
@@ -144,6 +144,18 @@ def bring_up(self):
...
+        :return: admin link state
+        """
+        cmd = "ip link show {}".format(self.name)
+        try:
+            output = run_command(cmd, self.host, sudo=True)
+            state = output.splitlines()[0].split('>')[0].split(',')[-1]
+            return state
+        except CmdError as e:
+            raise NWException("failed to get link state: %s" % e)
+
```
narasimhan-v, line 158: Not needed

avocado/utils/network/interfaces.py
```diff
@@ -152,6 +164,9 @@ def is_link_up(self):
         """
         if self.get_link_state() == 'up':
             return True
+        elif self.get_link_state() == 'unknown':
+            if 'LOWER_UP' in self.get_admin_link_state():
+                return True
         return False
```
narasimhan-v, line 171: Additional debug message will be helpful

avocado/utils/network/interfaces.py
narasimhan-v: somehow I reviewed the old code, which had `return` statement

avocado/utils/network/interfaces.py
```diff
@@ -144,6 +144,18 @@ def bring_up(self):
...
             raise NWException("Failed to bring up: %s" % ex)
 
+    def get_admin_link_state(self):
+        """Get admin link state.
+        :return: admin link state
+        """
+        cmd = "ip link show {}".format(self.name)
+        try:
+            output = run_command(cmd, self.host, sudo=True)
+            state = output.splitlines()[0].split('>')[0].split(',')[-1]
```
beraldoleal, line 154: we are already using `-j` (json output) option when using `_get_interface_details()`. You could avoid the splits and this index syntax.

Also `_get_interface_details()` is already handling some exceptions and looks like it has the information that you need, that would make your method much cleaner:

```python
     def get_admin_link_state(self):
          try:
              return self._get_interface_details().get('flags')[2]
          except (NWException, IndexError):
              raise NWException("Could not get admin link state.")
```

avocado/utils/network/interfaces.py
```diff
@@ -152,6 +164,9 @@ def is_link_up(self):
         """
         if self.get_link_state() == 'up':
             return True
+        elif self.get_link_state() == 'unknown':
+            if 'LOWER_UP' in self.get_admin_link_state():
+                return True
```
beraldoleal, line 169: Here you are changing the logic of this method. Not sure if we need this change here. Please see my first comment here.

avocado/utils/network/interfaces.py
```diff
@@ -144,6 +144,18 @@ def bring_up(self):
         except Exception as ex:
             raise NWException("Failed to bring up: %s" % ex)
 
+    def get_admin_link_state(self):
```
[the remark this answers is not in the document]
smruti77: @beraldoleal sorry, its just confusion about name

avocado/utils/network/interfaces.py
smruti77: @beraldoleal actually the problem we are facing with veth interface (virtual interface), it is in unknown state, so for this we wanted to check with operational state,is the interface is  up or down through "LOWER_UP"
```
env3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UNKNOWN
```

beraldoleal: Again, IMO this is changing the behavior here.

Here is what I'm thinking: With your change, adding the method `get_operational_link_state()`, you have introduced the separation between physical (operational) and logical (administrative) checks here, and this **is very nice**.

But we should keep this separation on the other methods: `is_link_up()` could be a method to check if the link is "fully UP" (operational *and* administrative), because there is no admin neither operational in its name.

Imagine this scenario: Link is physically UP (operational) but administratively down. With your change, this will return `True`. Do you see it?

We need to decide what `is_link_up()` will check. IMO we should check for a *AND* on both: administrative and operational.

What do you think?

smruti77: @beraldoleal yes we can do that,i about to add that.

avocado/utils/network/interfaces.py
```diff
@@ -144,14 +144,32 @@ def bring_up(self):
...
-        if self.get_link_state() == 'up':
-            return True
+        link_state = self.get_admin_link_state()
+        oper_state = self.get_operational_link_state()
+        if link_state == 'up':
+            if oper_state == 'LOWER_UP':
+                return True
+        elif link_state == 'unknown':
+            if oper_state == 'LOWER_UP':
+                return True
```
beraldoleal, line 172: This is the way to go. But IMO we should split the methods, so this method should be very small:

```python
def is_link_up(self):
   return self.is_link_admin_up() and self.is_link_operational_up()
```

Also if we are renaming the `is_link_up()` method this should be deprecated first, by adding a warning message AND adding the new method. We should remove this method, only after a few minor-releases.

smruti77: ok

avocado/utils/network/interfaces.py
```diff
@@ -189,20 +200,22 @@ def get_hwaddr(self):
         except Exception as ex:
             raise NWException("Failed to get hw address: {}".format(ex))
 
-    def get_link_state(self):
-        """Method used to get the current link state of this interface.
+    def is_link_admin_up(self):
```
beraldoleal, line 203: Again, you are removing a method without a deprecation message. IMO, we don't need to remove "get_link_state()" method. Since that can be used by other methods. So your new method "is_link_admin_up()" could call this method.

beraldoleal: Another suggestion is to keep the methods here organized in alphabetical order.

avocado/utils/network/interfaces.py
```diff
@@ -144,16 +144,37 @@ def bring_up(self):
         except Exception as ex:
             raise NWException("Failed to bring up: %s" % ex)
 
-    def is_link_up(self):
-        """Check if the interface is up or not.
+    def is_link_admin_up(self):
```
narasimhan-v, line 147: `is_admin_link_up`

avocado/utils/network/interfaces.py
```diff
@@ -144,16 +144,37 @@ def bring_up(self):
...
-                 will return False.
+        :return: True or False, True if network interface state is 'up'
+                 or 'unknown' otherwise will return False.
         """
-        if self.get_link_state() == 'up':
+        if self.get_link_state() == 'up' or 'unknown':
             return True
         return False
 
+    def is_link_operational_up(self):
```
narasimhan-v, line 157: `is_operational_link_up`

avocado/utils/network/interfaces.py
```diff
@@ -144,16 +144,37 @@ def bring_up(self):
...
+    def is_admin_link_up(self):
+        """Check the admin link state is up or not.
 
-        :return: True or False. True if the current state is UP, otherwise
-                 will return False.
+        :return: True or False, True if network interface state is 'up'
+                 or 'unknown' otherwise will return False.
         """
-        if self.get_link_state() == 'up':
+        if self.get_link_state() == 'up' or 'unknown':
```
beraldoleal, line 153: This PR is almost ready, thanks for the changes.

My last comment is here. `unknown` is part of the operational state. Should not be checked here. If the state is `unknown` we *cannot* say that admin link is up. Right?

smruti77: @beraldoleal but we are handling this for specific interface like veth(virtual interface), we are facing the issue veth interface thats why we included the operational state to check to whether the interface is up or down.
```
env3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 2000 qdisc fq_codel state UNKNOWN group default qlen 1000
    link/ether 8e:b2:5d:3a:1d:03 brd ff:ff:ff:ff:ff:ff
    inet 100.10.10.204/24 scope global env3
       valid_lft forever preferred_lft forever
```

narasimhan-v: @smruti77 Can you point the RedHat documentation here so that it is easier to review the changes ?

beraldoleal: My question is: Can we say that *every* interface with `UNKNOWN` state is *administratively up*? Can we ensure that? IMO, `unknown`

Now that you have the second method, asking if the operational state is up is not enough for your tests?

narasimhan-v: @beraldoleal Previously we only checked admin state. When trying to solve the case of 'unknown', we realized that we need to check both operational and admin state.

beraldoleal: @narasimhan-v @smruti77 RFC2863: *operational states*. IMO, this is not part of *admin state*.

https://tools.ietf.org/html/rfc2863

abdhaleegit, line 153: @smruti77 get_link_state() is actually checking for an operational state why you are calling in is_admin_link_up ?
also self._get_interface_details() is acutually checking the admin state than why you are calling in is_operational_link_up ?  is this a typo ?  it should be other way round

@beraldoleal we need to consider unknown state while checking veth interface due to this limitation from https://www.suse.com/support/kb/doc/?id=000019147

beraldoleal: @abdhaleegit I understand that we need to check for the unknown state, but as I mention before and you noticed too, we need to this in the proper place to avoid confusion.

PR #4057, opened 2020-08-05
[Bug fix] `core.show` is not respected [v2]

richtja:
The logging system inside avocado doesn't expect changes in configuration
between jobs. This PR brings feature for enabling and disabling logging
configurations of the system. It saves the old configurations to the list for
later use.

Reference: #3997

---
Changes from version (#4051)

- Skip condition for `/bin/true`
- inverted logic in `del_last_configuration()`

### DIFF
--- a/avocado/core/job.py
+++ b/avocado/core/job.py
@@ -139,11 +139,11 @@ class Job:
 
     def __enter__(self):
         self.setup()
-        if not output.STD_OUTPUT.configured:
-            output.reconfigure(self.config)
+        output.reconfigure(self.config)
         return self
 
     def __exit__(self, _exc_type, _exc_value, _traceback):
+        output.del_last_configuration()
         self.cleanup()
 
     def __start_job_logging(self):
--- a/avocado/core/output.py
+++ b/avocado/core/output.py
@@ -390,10 +390,38 @@ def early_start():
     logging.root.level = logging.DEBUG
 
 
+CONFIG = []
+
+
+def del_last_configuration():
+    if len(CONFIG) == 1:
+        return
+    configuration = CONFIG.pop()
+    for logger_name in configuration:
+        disable_log_handler(logger_name)
+    configuration = CONFIG[-1]
+    for logger_name, handlers in configuration.items():
+        logger = logging.getLogger(logger_name)
+        for handler in handlers:
+            logger.addHandler(handler)
+
+
 def reconfigure(args):
     """
     Adjust logging handlers accordingly to app args and re-log messages.
     """
+    def save_handler(logger_name, handler, configuration):
+        if logger_name not in configuration:
+            configuration[logger_name] = []
+        configuration[logger_name].append(handler)
+
+    # Delete last configuration
+    if len(CONFIG) != 0:
+        last_configuration = CONFIG[-1]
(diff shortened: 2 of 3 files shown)


avocado/core/job.py
```diff
@@ -139,11 +139,11 @@ def __init__(self, config=None):
 
     def __enter__(self):
         self.setup()
-        if not output.STD_OUTPUT.configured:
```
chloerh, line 142: @richtja I'm sorry that I have to ask about this merged pr, but I believe this change is affecting our test with avocado-vt and tp-libvirt. 

After removing this line:142, 'reconfigure' function will be run along with every test twice. And when '--show early' is not provided or 'AVOCADO_LOG_EARLY' is not set, the root logger will be disabled according to 


```
        if "early" in enabled:
            handler = add_log_handler("", logging.StreamHandler,
                                      STD_OUTPUT.stdout, logging.DEBUG)
            save_handler("", handler, configuration)
            handler = add_log_handler(LOG_JOB, logging.StreamHandler,
                                      STD_OUTPUT.stdout, logging.DEBUG)
            save_handler(LOG_JOB.name, handler, configuration)
        else:
            disable_log_handler("")
```

Then all logs supposed to be logged by 'logging.xxx()' are missing.  
Since usually we don't use '--show early' options, I'm wondering if this is not expected. Would you mind taking a look? Or should I raise a new issue?  Thanks in advance.

richtja: > After removing this line:142, 'reconfigure' function will be run along with every test twice. 

Hi, @chloerh I can't find out why it is running for every test twice. Can you give me some example where this is happening and I will look at that? Thanks.

chloerh: @richtja yes.  I'm not very familiar with Avocado, and these are my understandings: 

When initializing 'AvocadoApp' , there is 
```
        try:
            self._load_cli_plugins()
            self._configure_cli_plugins()
            self.parser.finish()
            settings.merge_with_configs()
            settings.merge_with_arguments(self.parser.config)
            self.parser.config.update(settings.as_dict())
            self._run_cli_plugins()
        except SystemExit as detail:
            # If someone tries to exit Avocado, we should first close the
            # STD_OUTPUT and only then exit.
            output.reconfigure(reconfigure_settings)
            STD_OUTPUT.close()
            sys.exit(detail.code)
        except:
            # For any other exception we also need to close the STD_OUTPUT.
            output.reconfigure(reconfigure_settings)
            STD_OUTPUT.close()
            raise
        else:
            # In case of no exceptions, we just reconfigure the output.
            output.reconfigure(self.parser.config)

```

I guess `output.reconfigure(self.parser.config)` is the first time running.

And the 2nd time should be entering a Job:
```
    def __enter__(self):
        self.setup()
        output.reconfigure(self.config)
        return self
```

I have been using Avocado as a testing tool with avocado-vt and tp-libvirt, this happens when I try to run test cases of tp-libvirt. Thank you for taking look at this.

chloerh: By 'this', I mean logs could not be printed when calling 'logging.debug/info/error'

chloerh: Yes, it's working well now, thank you :+1:

PR #4063, opened 2020-08-06
future.settings: improved docstring on default value

### LINKED ISSUE #4040: Settings registration should execute the key_type under default values
When registering a new namespace with the Settings module, the `key_type` is not applied to the default value. This may cause problems if the default value is anything different from a string.

beraldoleal:
This will fix #4040

### DIFF
--- a/avocado/core/future/settings.py
+++ b/avocado/core/future/settings.py
@@ -502,8 +502,11 @@ class Settings:
             What is the key name of your option inside that section.
 
         default : typeof(key_type)
-            What is the default value of your option if the key is not present
-            inside the section on all configuration files.
+            The default value of an option. It sets the option value when the
+            key is not defined in any configuration files or via command-line.
+            The default value should be "processed". It means the value should
+            match the type of key_type. Due to some internal limitations, the
+            Settings module will not apply key_type to the default value.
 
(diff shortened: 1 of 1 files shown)


avocado/core/future/settings.py
```diff
@@ -503,7 +503,10 @@ def register_option(self, section, key, default, help_msg, key_type=str,
 
         default : typeof(key_type)
             What is the default value of your option if the key is not present
-            inside the section on all configuration files.
+            inside the section on all configuration files. You need to use a
+            "processed" value here, that is already matching the `key_type`.
+            Because of some internal limitations, do not expect that Avocado
+            will apply `key_type` to your default value here.
```
willianrampazzo, line 509: Do you mind if I do I complete the rewriting of this paragraph? For me, it is still not clear. Here is my suggestion:


The default value of an option. It sets the option value when the key is not defined in any configuration files. The default value should be "processed". It means the value should match the type of `key_type`. Due to some internal limitations, the Settings will *not* apply `key_type` to the default value.

beraldoleal: @willianrampazzo not at all. Fixed on a push-force.

Issue #4076, opened 2020-08-10
interfaces.py :  Create a new network config file if not present and move to backup 


abdhaleegit:
```
2020-08-10 05:39:07,325 stacktrace       L0039 ERROR|
2020-08-10 05:39:07,325 stacktrace       L0042 ERROR| Reproduced traceback from: /usr/local/lib/python3.6/site-packages/avocado_framework-80.0-py3.6.egg/avocado/core/test.py:804
2020-08-10 05:39:07,332 stacktrace       L0045 ERROR| Traceback (most recent call last):
2020-08-10 05:39:07,332 stacktrace       L0045 ERROR|   File "/root/avocado-fvt-wrapper/tests/avocado-misc-tests/io/net/multicast.py", line 132, in tearDown
2020-08-10 05:39:07,332 stacktrace       L0045 ERROR|     self.networkinterface.restore_from_backup()
2020-08-10 05:39:07,332 stacktrace       L0045 ERROR|   File "/usr/local/lib/python3.6/site-packages/avocado_framework-80.0-py3.6.egg/avocado/utils/network/interfaces.py", line 339, in restore_from_backup
2020-08-10 05:39:07,332 stacktrace       L0045 ERROR|     "Backup file not available, could not restore file.")
2020-08-10 05:39:07,332 stacktrace       L0045 ERROR| avocado.utils.network.exceptions.NWException: Backup file not available, could not restore file.
2020-08-10 05:39:07,332 stacktrace       L0046 ERROR|

abdhaleegit:
currently the code for ip config handles this way; if config file is their it will move to backup or else it will create new one without move operation. 

So if no config exists, create new one and also move to backup

smruti77:
#1862

beraldoleal:
@smruti77 not understood the relationship between 1862 and this. 

Also, @abdhaleegit, after the discussion here https://github.com/avocado-framework/avocado/pull/4077, and since this is a library that exports "reasons" to the upper layers, this should be handled properly with try.. except.

I'm closing this, but please, feel free to reopen the discussion.

smruti77:
@beraldoleal just pointed that PR its because this issue handle in another way, so no need to handle like this and for closing the issue.

PR #4077, opened 2020-08-10
restore_from_backup() shouldn't raise exception if backup file not there

smruti77:
handle restore_from_backup() for not raise exception if backup file not
found,it should ignore if backup file not their.

### DIFF
--- a/avocado/utils/network/interfaces.py
+++ b/avocado/utils/network/interfaces.py
@@ -311,7 +311,7 @@ class NetworkInterface:
             msg = 'Failed to remove ipaddr. {}'.format(ex)
             raise NWException(msg)
 
-    def restore_from_backup(self):
+    def restore_from_backup(self, ignore_missing=True):
         """Revert interface file from backup.
 
         This method checks if a backup version  is available for given interface
@@ -332,5 +332,6 @@ class NetworkInterface:
         if os.path.exists(backup_file):
             shutil.move(backup_file, filename)
         else:
-            raise NWException(
-                "Backup file not available, could not restore file.")
+            if not ignore_missing:
+                raise NWException(
+                    "Backup file not available, could not restore file.")


beraldoleal:
Hi @smruti77, thanks for this.

But, why should we ignore it? Not sure here.

IMO it is important to give information to the upper layers why we couldn't restore from backup.

Upper layers came from different users, some might need to ignore and some, don't.  For those who would like to ignore the message, why not using a simple `try...except` to handle/ignore those exceptions?

smruti77:
@beraldoleal if backup file is not their, test is failing because of exception, we want if file it not there it should not  raise exception it should pass. and if some one need information they can pass  "ignore_missing=False".

beraldoleal:
> @beraldoleal if backup file is not their, test is failing because of exception, we want if file it not there it should not raise exception it should pass. and if some one need information they can pass "ignore_missing=False".

Why you don't pass on your test side? With a try.. except? Since that an exception it is expected?

abdhaleegit:
This patch fixes this issue https://github.com/avocado-framework/avocado/issues/4076

beraldoleal:
> This patch fixes this issue #4076

Can you explain a little bit, how this code will fix that? Because this is just ignoring an exception. And IMO it is bad for usability.

My question is the same: Why we can use a try.. except to catch this exception?

smruti77:
> > @beraldoleal if backup file is not their, test is failing because of exception, we want if file it not there it should not raise exception it should pass. and if some one need information they can pass "ignore_missing=False".
> 
> Why you don't pass on your test side? With a try.. except? Since that an exception it is expected?

@beraldoleal we can do in our test side by using try .. except. but need to change in all files. so i thought have change in interface.py it self, like _move_file_to_backup().  in restore_from_backup() also if backup file is their it will move or else fine.and if someone need information they can use "ignore_missing=False" too.

beraldoleal:
> > > @beraldoleal if backup file is not their, test is failing because of exception, we want if file it not there it should not raise exception it should pass. and if some one need information they can pass "ignore_missing=False".
> > 
> > 
> > Why you don't pass on your test side? With a try.. except? Since that an exception it is expected?
> 
> @beraldoleal we can do in our test side by using try .. except. but need to change in all files. so i thought have change in interface.py it self, like _move_file_to_backup(). in restore_from_backup() also if backup file is their it will move or else fine.and if someone need information they can use "ignore_missing=False" too.

I understand that will be easier to add code inside Avocado utils. And I'm more than ok with that (we need to make this library as useful as we can). Actually I'm much appreciated that you and others are pushing things to Avocado. 

But we need to this on the *right way* otherwise will impact other Avocado users, developers, and contributors. We are not creating this library for *one specific user*, otherwise, you could add this directly to your side.

We designed this library as a pure and straightforward library without "workarounds" so it would be easier to maintain and also to use.

Exceptions are the proper way to notify library users about the *reason* why something went wrong. You are trying to change the way that other users will use this library, because of your need. And changing the default behavior to *hide* the reason, I don't think that is the right way to do it.

IMO it is not a good practice to add another argument to ask for *failure reasons*. We have try.. except for that.

But again, that is my option.

smruti77:
> > > > @beraldoleal if backup file is not their, test is failing because of exception, we want if file it not there it should not raise exception it should pass. and if some one need information they can pass "ignore_missing=False".
> > > 
> > > 
> > > Why you don't pass on your test side? With a try.. except? Since that an exception it is expected?
> > 
> > 
> > @beraldoleal we can do in our test side by using try .. except. but need to change in all files. so i thought have change in interface.py it self, like _move_file_to_backup(). in restore_from_backup() also if backup file is their it will move or else fine.and if someone need information they can use "ignore_missing=False" too.
> 
> I understand that will be easier to add code inside Avocado utils. And I'm more than ok with that (we need to make this library as useful as we can). Actually I'm much appreciated that you and others are pushing things to Avocado.
> 
> But we need to this on the _right way_ otherwise will impact other Avocado users, developers, and contributors. We are not creating this library for _one specific user_, otherwise, you could add this directly to your side.
> 
> We designed this library as a pure and straightforward library without "workarounds" so it would be easier to maintain and also to use.
> 
> Exceptions are the proper way to notify library users about the _reason_ why something went wrong. You are trying to change the way that other users will use this library, because of your need. And also changing the default behavior to _hide_ the reason, I don't think that is the right way to do it.
> 
> IMO it is not a good practice to add another argument to ask for _failure reasons_. We have try.. except for that.
> 
> But again, that is my option.

ok we will change in test code. thanks.

PR #4090, opened 2020-08-17
Vmimage automatic documentation [v2]

richtja:
This generates a documentation table with every tested distribution for
vmimage.

---
Changes from v1 #4088:

- `docs/source/guides/writer/libs/vmimage.rst.data` in .gitingnore file
- import in top of file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ docs/source/api/test
 docs/source/api/utils
 docs/source/api/plugins
 docs/source/api/optional-plugins
+docs/source/guides/writer/libs/data/vmimage
 
 # This file should be generated with:
 # avocado config reference > docs/source/config/reference.rst
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -6,6 +6,8 @@ import importlib
 import os
 import sys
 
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -21,6 +21,7 @@ docs/source/api/test
 docs/source/api/utils
 docs/source/api/plugins
 docs/source/api/optional-plugins
+docs/source/guides/writer/libs/vmimage.rst.data
```
beraldoleal, line 24: My only suggestion is to rename this dir. maybe `data/vmimage.rst`

richtja: Yes, you are right, It will be better for future extension.

richtja: I force-pushed the change here

.gitignore
```diff
@@ -21,6 +21,7 @@ docs/source/api/test
 docs/source/api/utils
 docs/source/api/plugins
 docs/source/api/optional-plugins
+docs/source/guides/writer/libs/data/vmimage
```
beraldoleal, line 24: I mean, without `libs`. This is just a rst file, not a library, right?

PR #4091, opened 2020-08-17
future.settings: fix custom configs after plugins load too

beraldoleal:
Some plugins have custom configs we need to parse this as soon plugins
are loaded.


pevogam:
Good job @beraldoleal, thanks for integrating this fix quickly. Indeed, we can think about further improving it once all the dust from the current release settles.

PR #4125, opened 2020-08-21
core.suite: Make config optional argument

beraldoleal:
Because of line 33. Thanks @pevogam.


richtja (approved):
LGTM

PR #4206, opened 2020-09-04
docs: small change to fix str convertion when using test parameters

beraldoleal:
If readers would like to test a different parameter then the current
code will break because will not convert string to integer. This closes


willianrampazzo (approved):

clebergnu (approved):
LGTM, thanks!

PR #4224, opened 2020-09-09
[Diff Plugin] skip file if it cannot be decoded [v2].

willianrampazzo:
The SysInfo plugin produces a `journalctl.gz` that cannot be decoded with the current Diff plugin implementation.

Let's skip files that cannot be decoded for now and propose a more robust solution later.

Changes from v1:
- Add debug message;
- Update the documentation.

### DIFF
--- a/avocado/plugins/diff.py
+++ b/avocado/plugins/diff.py
@@ -415,7 +415,13 @@ class Diff(CLICmd):
                 name_header = ['\n', '** %s **\n' % name]
                 sysinfo.extend(name_header)
                 with open(os.path.join(path, name), 'r') as sysinfo_file:
-                    sysinfo.extend(sysinfo_file.readlines())
+                    try:
+                        sysinfo.extend(sysinfo_file.readlines())
+                    except UnicodeDecodeError:
+                        msg = ("Ignoring file %s as it cannot be decoded."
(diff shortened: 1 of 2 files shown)


avocado/plugins/diff.py
```diff
@@ -415,7 +415,13 @@ def _get_sysinfo(resultsdir, pre_post):
                 name_header = ['\n', '** %s **\n' % name]
                 sysinfo.extend(name_header)
                 with open(os.path.join(path, name), 'r') as sysinfo_file:
-                    sysinfo.extend(sysinfo_file.readlines())
+                    try:
+                        sysinfo.extend(sysinfo_file.readlines())
+                    except UnicodeDecodeError:
+                        msg = ("Ignoring file %s as it cannot be decoded."
+                               % name)
+                        LOG_UI.debug(msg)
```
beraldoleal, line 423: Just out of curiosity, this is still showing by default ?

beraldoleal (approved):

clebergnu (approved):
LGTM, thanks!

avocado/plugins/diff.py
willianrampazzo: Well, yes here :)

PR #4237, opened 2020-09-15
add a function to check root device

smruti77:
add is_root_device() to disk.py to check root device.

### DIFF
--- a/avocado/utils/disk.py
+++ b/avocado/utils/disk.py
@@ -154,3 +154,20 @@ def get_filesystem_type(mount_point='/'):
             _, fs_file, fs_vfstype, _, _, _ = mount_line.split()
             if fs_file == mount_point:
                 return fs_vfstype
+
+
+def is_root_device(device):
+    """
+    check for root disk
+
+    :param device: device to check
+    :returns: True or False, True if given device is root disk
+              otherwise will return False.
+    """
+    cmd = "lsblk --j -o MOUNTPOINT,PKNAME"
+    output = process.run(cmd)
+    result = json.loads(output.stdout_text)
+    for item in result['blockdevices']:
+        if item['mountpoint'] == "/" and device == str(item['pkname']):
+            return True
+    return False


avocado/utils/disk.py
```diff
@@ -154,3 +154,22 @@ def get_filesystem_type(mount_point='/'):
...
+    :param device: device to check
+    :returns: True or False, True if given device is root disk
+              otherwise will return False.
+    """
+    cmd = "lsblk --j -o MOUNTPOINT,PKNAME"
+    output = process.run(cmd)
+    result = json.loads(output.stdout_text)
+    for item in result['blockdevices']:
+        if item['mountpoint'] == "/":
+            res = str(item['pkname'])
```
willianrampazzo, line 172: Do you expect more than one mount point with on "/"? I don't see a use case for that, do you have one in mind?

If not, what do you think about:

```
diff --git a/avocado/utils/disk.py b/avocado/utils/disk.py
index 053d3bc1..6c3a737b 100644
--- a/avocado/utils/disk.py
+++ b/avocado/utils/disk.py
@@ -167,8 +167,6 @@ def is_root_device(device):
     output = process.run(cmd)
     result = json.loads(output.stdout_text)
     for item in result['blockdevices']:
-        if item['mountpoint'] == "/":
-            res = str(item['pkname'])
-    if res in device:
-        return True
+        if item['mountpoint'] == "/" and device == str(item['pkname']):
+            return True
     return False
```

avocado/utils/disk.py
```diff
@@ -154,3 +154,22 @@ def get_filesystem_type(mount_point='/'):
...
+    if res in device:
```
beraldoleal, line 173: res is only defined inside the if/for on line 172 of your patch. If the list is empty and/or this if is not reached then will raise an exception, right?

beraldoleal (changes requested):

avocado/utils/disk.py
smruti77: this is fine for me.

avocado/utils/disk.py
smruti77: yes

willianrampazzo (approved):

avocado/utils/disk.py
```diff
@@ -154,3 +154,20 @@ def get_filesystem_type(mount_point='/'):
...
+
+    :param device: device to check
+    :returns: True or False, True if given device is root disk
+              otherwise will return False.
+    """
+    cmd = "lsblk --j -o MOUNTPOINT,PKNAME"
+    output = process.run(cmd)
+    result = json.loads(output.stdout_text)
+    for item in result['blockdevices']:
+        if item['mountpoint'] == "/" and str(item['pkname']) in device:
```
beraldoleal, line 171: The way that I'm reading the docstrings here, is that `device` is a string, not a list. Also, the `pkname` value on the output command is also a string. So we need either change the docstring to accept "devices" and make clear that is list, or change this `in` by a `==`.

beraldoleal (changes requested):

avocado/utils/disk.py
willianrampazzo: Indeed, my original proposal was:

`if item['mountpoint'] == "/" and device == str(item['pkname']):`

I thought the diff was applied here and I did not check the condition before approve.

IMHO, it does not make sense to accept a list here as argument, but a string, so the check should be changed to `==`.

willianrampazzo (changes requested):
The condition verification should match the string argument.

avocado/utils/disk.py
smruti77: @willianrampazzo  @beraldoleal actually i put in just because, while we passing the device it is like "/dev/sda" and here we are getting output of str(item['pkname']) is 'sda'. to avoid extra lines i keep it like this, "str(item['pkname']) in device".   ok i will change that.

beraldoleal (approved):
Thanks for the changes @smruti77 . LGTM.

willianrampazzo (approved):

PR #4238, opened 2020-09-16
utils.cpu: Handle "aarch64" in get_family()

PaulYuuu:
utils.cpu.get_arch() can return "aarch64" but get_family() missed this
value, add it to handle this situation.

https://github.com/avocado-framework/avocado/blob/e97540793998c4f24a16000465dd7fdd213bf2b9/avocado/utils/cpu.py#L165-L175

### DIFF
--- a/avocado/utils/cpu.py
+++ b/avocado/utils/cpu.py
@@ -212,7 +212,7 @@ def get_family():
             family = zfamily_map[get_version()].lower()
         except KeyError as err:
             logging.warning("Could not find family for %s\nError: %s", get_version(), err)
-    elif arch == 'arm':
+    elif arch == 'arm' or arch == 'aarch64':
         raise NotImplementedError
     return family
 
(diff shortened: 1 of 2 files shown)


PaulYuuu:
Hello @sathnaga, can you help to review? Thanks a lot.

beraldoleal (approved):

willianrampazzo (approved):

avocado/utils/cpu.py
```diff
@@ -212,7 +212,7 @@ def get_family():
             family = zfamily_map[get_version()].lower()
         except KeyError as err:
             logging.warning("Could not find family for %s\nError: %s", get_version(), err)
-    elif arch == 'arm':
+    elif arch == 'arm' or arch == 'aarch64':
```
beraldoleal, line 215: Actually, this could be even more generic. I was debugging one problem today with Marcelo Condatta, and this was returning None. I just sent a patch here. We don't need this `or`.

#4239

beraldoleal (changes requested):

avocado/utils/cpu.py
PaulYuuu: Thank you @beraldoleal, I will close this one.

PR #4242, opened 2020-09-18
add a function to get default interface

smruti77:
add get_default_interface() in interface.py to get default
interface name

### DIFF
--- a/avocado/utils/network/interfaces.py
+++ b/avocado/utils/network/interfaces.py
@@ -176,6 +176,19 @@ class NetworkInterface:
         """
         return self.is_admin_link_up() and self.is_operational_link_up()
 
+    def get_default_interface(self):
+        """Get the default interface.
+
+        :return: Default interface name.
+        """
+        output = run_command("ip route", self.host)
+        try:
+            for line in output.splitlines():
+                if "default" in line:
+                    return line.split()[4]
+        except Exception as ex:
+            raise NWException("could not get default interface: {}".format(ex))
+
     def get_ipaddrs(self, version=4):
         """Get the IP addresses from a network interface.
 


willianrampazzo (commented):
I noticed the `ip` command also has the `-j` option for JSON output. Doesn't that make things easy? You will not need to handle splits of multiples lines or splits of a line to get the values you need. Just a suggestion here.

avocado/utils/network/interfaces.py
```diff
@@ -176,6 +176,19 @@ def is_link_up(self):
         """
         return self.is_admin_link_up() and self.is_operational_link_up()
 
+    def get_default_interface(self):
```
beraldoleal, line 179: to avoid ambiguity, I suggest calling this  `get_default_route_iface` or `get_default_route_interface()`.

beraldoleal (changes requested):
Hi @smruti77 this is welcome, but I left a few comments for you.

avocado/utils/network/interfaces.py
```diff
@@ -176,6 +176,19 @@ def is_link_up(self):
...
+        """Get the default interface.
+
+        :return: Default interface name.
+        """
+        output = run_command("ip route", self.host)
```
PaulYuuu, line 184: You can use `ip route list default`

willianrampazzo: Indeed, `ip -j route list default` is a good option here.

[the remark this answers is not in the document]
smruti77: ok

avocado/utils/network/interfaces.py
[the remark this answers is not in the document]
smruti77: in Host() will be better.

avocado/utils/network/interfaces.py
smruti77: ok, will do that.

smruti77:
#4245

PR #4248, opened 2020-09-24
Setuptools implementation

t0xic0der:
- Replaced version check via `distutils.LooseVersion()` with the `packaging.version()`.
- Excluded `python-distuils` in Debian rules.
- Note that `packaging` is included in `setuptools`.

### DIFF
--- a/avocado/utils/kernel.py
+++ b/avocado/utils/kernel.py
@@ -22,7 +22,7 @@ import multiprocessing
 import os
 import shutil
 import tempfile
-from distutils.version import LooseVersion  # pylint: disable=E0611
+from packaging import version as versobjc
 
 from . import archive, asset, build, distro, process
 
@@ -207,4 +207,4 @@ def check_version(version):
     :type version: string
     :param version: version to be compared with current kernel version
     """
-    assert LooseVersion(os.uname()[2]) > LooseVersion(version), "Old kernel"
+    assert versobjc.parse(os.uname()[2]) > versobjc.parse(version), "Old kernel"
--- a/contrib/packages/debian/rules
+++ b/contrib/packages/debian/rules
@@ -4,7 +4,7 @@
 DEB_PYTHON_SYSTEM := pysupport
 
 include /usr/share/cdbs/1/rules/debhelper.mk
-include /usr/share/cdbs/1/class/python-distutils.mk
+#include /usr/share/cdbs/1/class/python-distutils.mk
 
 clean::
 	rm -rf build build-stamp configure-stamp build/ MANIFEST


contrib/packages/debian/rules
```diff
@@ -4,7 +4,7 @@
 DEB_PYTHON_SYSTEM := pysupport
 
 include /usr/share/cdbs/1/rules/debhelper.mk
-include /usr/share/cdbs/1/class/python-distutils.mk
+#include /usr/share/cdbs/1/class/python-distutils.mk
```
richtja, line 7: Why you comment on this and not deleted it?

t0xic0der: I wasn't as sure about the rules file as much as I was for the Python files. If they require removal, I'd do right away.

avocado/utils/kernel.py
```diff
@@ -22,7 +22,7 @@
 import os
 import shutil
 import tempfile
-from distutils.version import LooseVersion  # pylint: disable=E0611
+from packaging import version as versobjc
```
richtja, line 25: we use [isort](https://pycqa.github.io/isort/) for checking the ordering of the imports and it is complaining about this you have to change this.

richtja (changes requested):
Hi @t0xic0der LGTM, but there is small problems on the Travis which have to be fixed.  It is complaining about sign and isort. You have to sign all your commits, more info is in [documentation](https://avocado-framework.readthedocs.io/en/latest/guides/contributor/chapters/styleguides.html), and check alphabetical ordering of imports. 

You can use `make check` to run the same test as Travis locally.

t0xic0der:
Alright.  I would make those corrections in the impending pushes.

contrib/packages/debian/rules
beraldoleal: I'm not so familiar with CDBS in debian rules and how this will affect this migration. Yes, afaict, we need to remove this, but just need to make sure that debian packages are still working or if we need to add some rules for setuptools too.

beraldoleal (commented):
Hi @t0xic0der, thanks for this.

Besides the comment that I made about debian rules, and the comments made by @richtja, this looks good me. 

I just would like to make sure if others have any objections here before merging.

contrib/packages/debian/rules
t0xic0der: `setuptools` are the defacto Python packaging standard as of 3.8 so I do not think we need to explicitly specify about it in the Debian rules (unless I figured the way that file works wrongly)

avocado/utils/kernel.py
t0xic0der: Cool. I would make this change happen
.

richtja:
Closing in favor of v2 #4253

PR #4249, opened 2020-09-24
Used time.monotonic() for utils/process.py timeout

t0xic0der:
- Ran a `grep -Ril "start_time" | grep ".py"` and `grep -Ril "stop_time" | grep ".py"` to recursively scan for occurrences of "start_time" and "stop_time" (which would infer that duration is most likely being calculated in those Python files) and listed them. (Reference is taken from #3887)
```
$ grep -Ril "start_time" | grep ".py"
avocado/core/data_dir.py
avocado/utils/process.py
avocado/utils/wait.py
```
```
$ grep -Ril "stop_time" | grep ".py"
avocado/utils/process.py
```
- Made some 7 replacements in `avocado/utils/process.py`.
- Made the commit from the browser itself so it gets automatically signed.
- Please disregard #4243 and consider reviewing this.

### DIFF
--- a/avocado/utils/kernel.py
+++ b/avocado/utils/kernel.py
@@ -22,7 +22,7 @@ import multiprocessing
 import os
 import shutil
 import tempfile
-from distutils.version import LooseVersion  # pylint: disable=E0611
+from packaging import version as versobjc
 
 from . import archive, asset, build, distro, process
 
@@ -207,4 +207,4 @@ def check_version(version):
     :type version: string
     :param version: version to be compared with current kernel version
     """
-    assert LooseVersion(os.uname()[2]) > LooseVersion(version), "Old kernel"
+    assert versobjc.parse(os.uname()[2]) > versobjc.parse(version), "Old kernel"
--- a/avocado/utils/process.py
+++ b/avocado/utils/process.py
@@ -216,7 +216,7 @@ def kill_process_tree(pid, sig=None, send_sigcont=True, timeout=0):
         sig = signal.SIGKILL
 
     if timeout > 0:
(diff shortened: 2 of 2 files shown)


avocado/utils/kernel.py
```diff
@@ -22,7 +22,7 @@
 import os
 import shutil
 import tempfile
-from distutils.version import LooseVersion  # pylint: disable=E0611
+from packaging import version as versobjc
```
beraldoleal, line 25: Your PR it is talking about time.monotonic() and here we have changes about distutils. Maybe you need a rebase or fork from master to avoid having this changes here.

beraldoleal (changes requested):

avocado/utils/kernel.py
t0xic0der: Oh my bad. The changed file from another PR slipped in here. I'll restore the original kernel.py right away.

t0xic0der: (I think I would wait on #4253 being taken in correctly. Once, it happens I would know just what all I should do and what all I should avoid.)

beraldoleal: Or you could rebase this PR based on master, to avoid dependencies, since this PR is not related to #4253.

t0xic0der:
> @t0xic0der do you have intentions to update this? Is this still valid?

@beraldoleal It is in the backlog list of my personal kanban board so it would be some time before I'd be able to come back to this. You might want to wait or maybe assign this to someone else should they request assignment.

beraldoleal:
Closing this in favor of #4551.

PR #4262, opened 2020-10-08
blueprints: Adding a BP000 as draft

beraldoleal:
I will add all reviewers to the header after discussions.

### DIFF
--- a/docs/source/blueprints/BP000.rst
+++ b/docs/source/blueprints/BP000.rst
@@ -1,0 +1,319 @@
+:Number:BP000
+:Title: Blueprint specification and requirements
+:Author: Beraldo Leal <bleal@redhat.com>
+:Reviewers: Cleber Rosa <crosa@redhat.com>, Jan Richter <jarichte@redhat.com>,
+            Plamen Dimitrov <pdimitrov@pevogam.com>, Willian Rampazzo
+            <willianr@redhat.com>
+:Created: 29-Sep-2020
+:Type: Meta Blueprint
+:Status: Draft
+
+.. contents:: Table of Contents
+
+TL;DR
+*****
+
+Having better records of architectural decisions in our repository is a good
+way to socialize our technical decisions, improve the code review process,
+avoid unnecessary code, and balance the workload among other developers. We are
+already using the “blueprint” documents for a while and still learning how to
+write them. Although we have a basic idea of what is a blueprint, some may not
+understand the concept because we are missing one meta-document that describes
+the blueprint's basic notions. This document is a blueprint to specify how we
+should write blueprints from now on.
+
+Motivation
+**********
+
+Depending on the project size, having very well defined and structured
+documents about the architecture decisions seems like overkill, but could help
+projects of any size, including Avocado, save time, make better decisions and
+improve the way we socialize those decisions.
+
+Today in the Avocado project we have the good practice to submit an RFC to
+our mailing list or to use/open a GitHub issue when we have a new idea. RFCs are a
+widespread way to disclose the architecture decisions, but they are just one
+part of a longer process. During the RFC phase, we argue in favor of a
+proposal, and we are mostly concerned about collecting feedback. After this
+phase, we could go one step forward and consolidate the discussion in
+Blueprints (sometimes called ADRs - Architecture Decision Records). This could
+be the next step so we could better socialize our decisions for future readers.
+A very well defined and structured document has some advantages over an RFC,
+but it is not intended to replace it.
+
+With blueprints, we could not only, but mainly:
+
+  * Create better documents for future members and future reference, when we
+    are trying to answer those questions:
+
+a) *why* are you doing this? (the “problem” or the “motivation”);
+b) *what* are you proposing to solve the problem? (your solution)?
+c) And *how* are you going to implement the proposed solution?
+
+Depending on the type of your blueprint, the answer for the last question (c)
+could be written in pseudocode, general text or even not necessary (although
+desired) — more details on the section named Specification.
+
+  * Make sure that another peer will be able to implement a feature/requirement.
+
+Blueprints are not for you; they are for the community. If you know that a
+problem exists and know how to fix it, the most natural course of action would
+be to start coding this fix and submitting a Pull Request. While this is still
+valid for most of the cases, some important architectural changes must be
+discussed first to explain the “why”, “what” and “how” to keep everyone on the
+same page, avoid unnecessary coding, and most importantly: allow others to
+implement it in case you are not available.
+
+  * Improve the code review quality and time.
+
+Having a better understanding of the problem and the big picture is better for
+code review. It is harder to capture that from the Pull Request (PR)
+description and PR changes. Developers that are aware of the problem tend to
+review your changes with the problem in mind and hence more quickly. 
+
+  * Reduce onboarding time for new members.
+
+Having a history of how we made an architectural decision and why we are
+implementing it this way will give new members reading material to understand
+our project, avoiding unnecessary discussions and meetings to explain
+something.
+
+  * Create a common standard that will make it easier for readers.
+
+With an open RFC, authors tend to organize the ideas in different ways with
+different sections. Having a very well structured document with common sections
+will make it easier for readers to understand the problem and the solution.
+
+  * Track the progress of a significant implementation.
+
+We could use the blueprints header “status” line to track the progress of some
+features. We could even have a page parsing and listing all the blueprints with
+the title, author, status, and target milestone for that feature.
+
+  * Find the middle ground between “overthinking” and “auto-pilot.”
+
+Last but not least: we are *not trying* to overthink here and/or slow down our
+development processes. The idea is to have a lightweight document, very
+objective that will save us time in the medium and long run. We don’t have to
+overthink by trying to handle any possible scenario outside of ones we actually
+have a use case for. But we should also avoid the “auto-pilot” mode in our
+contributions.
+
+
+Specification
+*************
+
+One blueprint per topic
+-----------------------
+
+Try to follow the minimalist approach and be concise with content relevant to
+one particular topic. If you have a more general topic to discuss, you should
+set the type as “Epic Blueprint” (more below) but still try to be concise and
+focused on the subject.
+
+File format and location
+------------------------
+
+Our current documentation already uses ReStructuredText (.rst) format, so we
+will adopt .rst format here too. All blueprints will be located inside
+`docs/source/blueprints` with the filename `BPXXX.rst`, where XXX is the number
+of the blueprint. Just pick the next number available for your blueprint.
+
+Write for your audience
+-----------------------
+
+As mentioned before, your blueprint will be read by your peers, future members,
+and future yourself. Keep in mind that your audience is developers with a
+minimal understanding of the Avocado internals, but also be gentle and provide
+any context to understand the problem.
+
+Blueprints types
+----------------
+
+Currently, we have the following blueprint types:
+
+ * Architectural Blueprint: Any blueprint changing or introducing a new core
+   feature or architectural change to Avocado.
+
+ * Process Blueprint: Any blueprint that is not implementing a new core
+   feature, but changing how the project works. This could be, for instance,
+   related to the repositories or processes.
+
+ * Meta Blueprint: A blueprint about blueprints. Like this one and any future
+   blueprint that changes our blueprint's styles and methods.
+
+ * Epic Blueprint: A blueprint that is touching on multiple areas and is too
+   big to have all the documentation in one single blueprint. We could split
+   epic blueprints into smaller blueprints or issues (if they are small and
+   easy to understand).
+
+ * Component Blueprint: A blueprint with the intent to describe a new utility
+   module or a new plugin.
+
+Headers
+-------
+
+Python PEPs (Python Enhancement Proposals) uses RFC822 for describing the
+headers. This could be useful here too, especially when parsing those headers
+to display our list of blueprints with the current status.
+
+The current list of items of our blueprint headers is below:
+
+ * Number: Usually, the blueprint number in the format BPXXX
+
+ * Title: A short descriptive title, limited to 80 characters
+
+ * Author: The author or authors of blueprint
+
+ * Reviewers: All reviewers that approved and helped during the review process
+
+ * Created: Date string when the blueprint first draft was submitted. Please
+   use the following format: DD-MMM-YYYY.
+
+ * Type: One of the types described during the previous section
+
+ * Status: One of the types described during the next section
+
+Here is an example of a header::
+
+   :Number: BP001
+   :Title: Configuration by convention
+   :Author: Beraldo Leal <bleal@redhat.com>
+   :Reviewers: Cleber Rosa, Lukáš Doktor and Plamen Dimitrov
+   :Created: 06-Dec-2019
+   :Type: Epic Blueprint
+   :Status: WIP
+
+Blueprint statuses
+------------------
+
+ * Draft: All blueprints should be created in this state. This means the
+   blueprint is accepting comments, and probably there is a discussion
+   happening. Blueprints in draft mode can be part of our repository.
+
+ * Approved: Blueprint was approved after discussions, and all suggestions are
+   already incorporated on the document. Nobody has started working on this
+   yet.
+
+ * Assigned: This status is not about the blueprint itself, but about the proposal
+   that is the subject of the BP. This means that the blueprint was approved,
+   and someone is already working on implementing it. A BP status can change
+   from Draft to Assigned if the work has started already.
+
+ * Implemented: This means the BP is already implemented and delivered to the
+   Avocado’s master branch.
+
+ * Rejected: Rejected status means the idea was not implemented because it
+   wasn't approved by everyone or has some techinical limitations.
+
+ * Deprecated: Deprecated means it was approved, implemented, and at some
+   point, makes no more sense to have it. For example, anything related to the
+   legacy runner. Usually, Deprecated means that it was replaced by something
+   else.
+
+As you can see, there is no status to accommodate any future change in a
+blueprint. Blueprints should not be “voided.” Any improvement on an old
+blueprint should be presented as a new blueprint, changing the status of the
+original to “deprecated”.
+
+Sections
+--------
+
+In order to facilitate the reading and understanding of the problem, all
+blueprints must have the following sections:
+
+ * TL;DR
+
+ * Motivation
+
+ * Specification
+
+ * Backwards Compatibility
+
+ * Security Implications
+
+ * How to Teach This
+
+ * Related Issues
+
+ * References
+
+Bellow a brief description of what you should write in each section:
+
+ * TL;DR: Should be a short description of your blueprint. Like an abstract. We
+   recommend writing this at the end of your first draft. This will give you a
+   better overview of it.
+
+ * Motivation: This should be the motivation of your proposed solution, not the
+   motivation of the blueprint itself. It describes the problem. Here, the
+   proposed solution should answer “why” this is needed.
+
+ * Specification: In this section, you should describe how you are going to
+   solve the problem. You can create subsections here to organize your ideas
+   better. Please keep in mind that it is useful to mention the details, with
+   code snippets, examples, and/or references. This will save you time, making
+   sure that everyone is in agreement with the proposed solution.
+
+ * Backwards Compatibility: How is your proposal going to affect older versions
+   of Avocado? Should we deprecate some modules, classes, or methods? Are we
+   going to keep backwards compatibility or not?
+
+ * Security Implications: Do you have any concerns about security with your
+   proposed solution and what are they?
+
+ * How to Teach This: What is the best way to teach our devs and users your new
+   feature/solution?
+
+ * Related Issues: Here, you should mention Github links for both: a) current
+   open issues that are blocking while waiting for your BP and b) all open
+   issues that will render this BP as “implemented” when closed.
+
+   * Issues to address this BP
+
+   * Issues this BP will solve
+
+ * References: Any external reference for helping understand the problem and
+   your solution.
+
+Backwards Compatibility
+***********************
+
+So far, we are on our 3rth blueprint (BP003 was the last one). This BP000
+should have been released before those blueprints. So probably those three
+blueprints are not 100% compliance with this meta blueprint, and that is fine.
+We were learning on the fly. We don’t need to change any of those blueprints
+after BP000 gets approved.
+
+Security Implications
+*********************
+
+No security implications found so far.
+
+How to Teach This
+*****************
+
+Getting used to writing blueprints is not an easy task. And probably we are
+going to find non-planned issues with this process on the way.  The general
+rule of thumb is to use common sense. To make this more public, we could
+consider the following:
+
+ * If approved, BP000 should be on top of our blueprints lists for reference.
+
+ * We could also have a template inside the `blueprints` directory to help
+   people when submitting their own blueprints.
+
+ * Also, we could include pointers and instructions in our development guide
+   for this BP.
+
+ * Another good practice would be to make comments in Avocado’s source code
+   with some pointers to specific blueprints.
+
+Related Issues
+**************
+
+None yet.
+
+References
+**********
+
+Todo


pevogam:
@beraldoleal Thanks for your patience with this PR since it is quite important for most of us and needs more eyes and opinions as well as serious reading. I will make sure to take a look at it too as soon in the following week.

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,307 @@
...
+
+Try to follow the minimalist approach and be concise with content relevant to
+one particular topic. If you have a more general topic to discuss, you should
+set the type as “Epic Blueprint” (more below) but still try to be concise and
+focused on the subject.  File format and location
+
+Our current documentation already uses ReStructuredText (.rst) format, so we
+will adopt .rst format here too. All blueprints will be located inside
+`docs/source/blueprints` with the filename `BPXXX.rst`, where XXX is the number
+of the blueprint. Just pick the next number available for your blueprint.
```
pevogam, line 115: Hopefully no concurrent problems of multiple people starting a blueprint at the same time but probably it won't happen in practice.

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,307 @@
...
+Write for your audience
+
+As mentioned before, your blueprint will be read by your peers, future members,
+and yourself. Keep in mind that your audience is developers with a minimal
```
pevogam, line 119: and future yourself :)

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,307 @@
...
+understanding of the Avocado internals, but be gentle and provide any context
```
pevogam, line 120: I guess the "but" is misleading here since if one keeps in mind the audience has minimal understanding of the internals, they would *also* provide further context to understand the problem.

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,307 @@
...
+to understand the problem.  Blueprints types
+
+Currently, we have the blueprint types:
```
pevogam, line 123: "have the following blueprint types"

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,307 @@
...
+
+ * Architectural Blueprint: Any blueprint changing or introducing a new core
+   feature or architectural change to Avocado.
+
+ * Process Blueprint: Any blueprint that is not implementing a new core
+   feature, but changing how the project works. This could be, for instance,
+   related to the repositories or processes.
+
+ * Meta Blueprint: A blueprint about blueprints. Like this one and any future
+   blueprint that changes our blueprints styles and methods.
```
pevogam, line 133: " our blueprints' "

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,307 @@
...
+   
+ * Epic Blueprint: A blueprint that is touching in multiple areas and is too
```
pevogam, line 135: "touching on"

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,307 @@
...
+   big to have all the documentation in one single blueprint. We could split
+   epic blueprints into smaller blueprints or issues (if they are small and
+   easy to understand).
```
pevogam, line 138: This is confusing though - do we first create the big blueprint? If we do when do we just copy its parts into separate follow up blueprints? Does that mean the first one is kept and some of the information is redundant across blueprints?

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,307 @@
...
+
+ * Utility Blueprint: A blueprint with the intent to describe a new utility
+   module or a new plugin.
```
pevogam, line 141: I guess it could also be called (non-core and thus not architectural) "component" blueprint. Or "enhancement" blueprint. It could relate to additional plugins or components and not just utilities.

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,307 @@
...
+
+ * Number: Usually, the blueprint number in the format BPXXX
+
+ * Title: A short descriptive title, limited to 80 characters
+
+ * Author: The author or authors of blueprint 
+
+ * Reviewers: All reviewers that approved and helped during the review process
+
+ * Created: Date string when the blueprint first draft was submitted
```
pevogam, line 160: Is the date format also specified?

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,307 @@
...
+   happening. Blueprints in draft mode can be part of our repository. 
+
+ * Approved: Blueprint was approved after discussions, and all suggestions are
+   already incorporated on the document. Nobody has started working on this
+   yet.
+
+ * WIP: This status is not about the blueprint itself, but about the proposal
+   that is the subject of the BP. This means that the blueprint was approved,
+   and someone is already working on implementing it. A BP status can change
+   from Draft to WIP if the work has started already.
```
pevogam, line 190: Sounds a bit like "Assigned" which is a bit more compatible with the other statuses as a name.

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,307 @@
...
+ * WIP: This status is not about the blueprint itself, but about the proposal
+   that is the subject of the BP. This means that the blueprint was approved,
+   and someone is already working on implementing it. A BP status can change
+   from Draft to WIP if the work has started already. 
+
+ * Implemented: This means the BP is already implemented and delivered to the
+   Avocado’s master branch. 
+
+ * Rejected: Rejected status means the idea was not implemented because it
+   makes no sense or has some technical limitations.
```
pevogam, line 196: "makes no sense" is too strong, I would go for "wasn't approved by everyone or has some technical limitations".

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,307 @@
...
+
+ * Deprecated: Deprecated means it was approved, implemented, and at some
+   point, makes no more sense to have it. For example, anything related to the
+   legacy runner. Usually, Deprecated means that it was replaced by something
+   else.
+
+As you can see, there is no status to accommodate any future change in a
+blueprint. Blueprints should not be “voided.” Any improvement on an old
+blueprint should be presented as a new blueprint, changing the status of the
+original to “deprecated”.
```
pevogam, line 206: +1

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,307 @@
...
+   better. Please keep in mind that it is useful to mention the details, with
+   code snippets, examples, and/or references. This will save you time, making
+   sure that everyone is in agreement with the proposed solution.
+
+ * Backwards Compatibility: How is your proposal going to affect older versions
+   of Avocado? Should we deprecate some modules, classes, or methods? Are we
+   going to keep backwards compatibility or not?
+
+ * Security Implications: What are your concerns about security with your
+   proposed solution?
```
pevogam, line 251: Maybe better since many people might have none: "Do you have any concerns about security with your proposed solution and what are they?"

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,307 @@
...
+
+ * How to Teach This: What is the best way to teach our devs and users your new
+   feature/solution?
+
+ * Related Issues: Here, you should mention Github links for both: a) current
+   open issues that are blocking waiting for your BP and b) all open issues
```
pevogam, line 257: "blocking and waiting for ..." or "blocking while waiting for ..."

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,307 @@
...
+   that will make this BP as “implemented” when closed.
```
pevogam, line 258: "that will render this BP "implemented" when closed" if I understand correctly

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,307 @@
...
+
+   * Issues to address this BP
+
+   * Issues this BP will solve
```
pevogam, line 262: This is duplicated the above a) and b) but is more compact and better explaining the two types of referenced issues.

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,307 @@
...
+*********************
+
+No security implications found so far.
+
+How to Teach This
+*****************
+
+Getting used to writing blueprints is not an easy task. And probably we are
+going to find non-planned issues with this process on the way.  The general
+rule of thumb is to use common sense. To make this more public, we could:
```
pevogam, line 286: "we could consider the following:"

pevogam (commented):
I think I left 43 comments :grimacing: But most of them are small fixes so they should be easy to implement or refute :)

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,307 @@
...
+ * If approved, BP000 should be on top of our blueprints lists for reference.
+
+ * We could also have a template inside the `blueprints` directory to help
+   people when submitting their own blueprints.
+
+ * Also, we could include pointers and instructions in our development guide
+   for this BP.
+
+ * Another good practice would be to make comments in Avocado’s source code
+   with some pointers to specific blueprints.
```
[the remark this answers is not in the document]
beraldoleal: I see. Since we are still learning how to make blueprints, I guess the language needs to change from draft to approved. During the draft phase, the author is trying to prove something, and after that, we are informing our users/developers.

But I guess the whole idea of this section is to list all things that we need to do, in order to teach our users about this BP.

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,307 @@
...
+decisions and improve the way we socialize those decisions.
+
+Today we have inside the Avocado project the good practice to submit an RFC to
+our mailing list or using a GitHub issue when we have a new idea. RFCs are a
+widespread way to disclose the architecture decisions, but they are just one
+part of the process. During the RFC phase, we argue in favor of a proposal, and
+we are mostly concerned about collecting feedback. After this phase, we could
+go one step forward and consolidate the discussion in Blueprints (sometimes
+called ADR - Architecture Decision Records). This could be the next step so we
+could better socialize our decisions for future readers. A very well defined
```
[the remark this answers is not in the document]
beraldoleal: I think that I prefer to use socialize, since I would like to emphasize this word.

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,307 @@
...
+and structured document has some advantages over an RFC, but it is not intended
+to replace it.
```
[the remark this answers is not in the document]
beraldoleal: I guess, that we miss some sections on this BP yet. I was thinking about this. "When to write a BP" is a missing section here.

IMO, we could have only a BP if the author is familiar with the BP structure and has confident that topic is a BP suitable. But sometimes, new members don't know about BP, or are not 100% sure if it is a good idea. In this case, makes sense to send a RFC first with a more informal discussion as a preparation for a BP. Makes sense?

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,307 @@
...
+
+With blueprints, we could not only, but mainly:
+
+  * Create better documents for future members and future reference.
```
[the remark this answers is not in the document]
beraldoleal: When using RFCs as email threads, there are no sections, headers, IMO, the RFCs as we use today are just thread discussions and are not focused on future review/reading. But if our RFCs has the same content, headers and goal of  BP, then yes, it could be.

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,307 @@
...
+
+By trying to answer those questions:
+
+a) *why* are you doing this (the “problem” or the “motivation”);
+b) *what* are you proposing to solve the problem (your solution)?
+c) And *how* are you going to implement?
+
+Depending on the type of your blueprint, the answer for the last question (c)
+could be high level or even not necessary (although desired) — more details on
+the section named Specification.
```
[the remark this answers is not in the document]
beraldoleal: We don't need to get into code here. Just describing what we are going to use. Like a pseudocode or description.

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,307 @@
...
+
+With an open RFC, authors tend to organize the ideas in different ways with
+different sections. Having a very well structured document with common sections
+will make it easier for readers to understand the problem and the solution.
+
+  * Track the progress of a significant implementation.
+
+We could use the blueprints header “status” line to track the progress of some
+features. We could even have a page parsing and listing all the blueprints with
+the title, author, status, and target milestone for that feature.
```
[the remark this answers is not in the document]
beraldoleal: Yes, you are right. Maybe we could have one epic issue connected to every BP, and keep the tracking by looking at this issue. Or even a milestone.

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,307 @@
...
+
+  * Find the middle ground between an “overthinking” and “auto-pilot.”
+
+Last but not least: we are *not trying* to overthink here and/or slow down our
+development processes. The idea is to have a lightweight document, very
+objective that will save us time in the medium and long run. We don’t need to
+overthink trying to find any possible scenario that we don’t have a use case.
+But we should also avoid the “auto-pilot” mode to fix things.
```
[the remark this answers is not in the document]
beraldoleal: Makes sense.

beraldoleal:
Hi @pevogam, thanks for your detailed review. I send a v2 with your suggestions soon.

docs/source/blueprints/BP000.rst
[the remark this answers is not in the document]
pevogam: Alright, both ways work fine and we can decide which form is better in the future with some extra use.

docs/source/blueprints/BP000.rst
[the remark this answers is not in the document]
pevogam: Alright, still ok

docs/source/blueprints/BP000.rst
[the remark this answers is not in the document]
pevogam: Yes, for sure a new user would ask themselves what is the difference between an RFC and a blueprint but at least the explanation here doesn't seem full clear to me. We can simplify and go straight to the point with a sentence or two saying that we understand a blueprint as more formal document, containing and RFC within itself and additional points (...). Then we could mention that a user could start with the less formal RFC within the document instead of the general statements that RFC remain irreplaceable which made me wonder "in what aspects?".

docs/source/blueprints/BP000.rst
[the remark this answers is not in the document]
pevogam: Alright, it is good to mention this in the same specificity in the blueprint itself. This makes it much clearer.

docs/source/blueprints/BP000.rst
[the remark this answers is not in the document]
pevogam: How about "could be high level" -> "could be written in pseudocode or general text"?

docs/source/blueprints/BP000.rst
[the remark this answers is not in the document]
beraldoleal: Yes, the terms are confusing.... In this document, when I talk about RFC, I'm talking about the "open format email" that we sent to the mailing list (with RFC in the subject).

The sentence is there just to make clear that it was not my intetion to abolish RFC (as we make it today). And members could still use the mailing list to start a discussion, by sending a RFC. 

But in fact, RFC could be just any "request for comments": an email starting a discussion with no RFC in the subject; or even a PR with a blueprint in draft mode.

It is just a difference about the approach, instead of trying to make our RFC more formal, I'm just introducing a new model (blueprints) that is the formal "version" of our RFCs.

So, what I was trying to say was: "By introducing this blueprint model we don't need to abolish the RFC as we use today, users can still send RFCs"

But yes, we need to make this clear in this document.

docs/source/blueprints/BP000.rst
beraldoleal: Yes, I thought about it and decided to go with this simple approach.

If we have this, both are going to be in draft mode, and we can request changes before the merge.

docs/source/blueprints/BP000.rst
beraldoleal: Yes, I'm ok with Component.

docs/source/blueprints/BP000.rst
beraldoleal: Now, it is. Thanks.

docs/source/blueprints/BP000.rst
beraldoleal: I'm just giving the sub-sub section names here.

docs/source/blueprints/BP000.rst
pevogam: This remains here: I am not sure if we copy parts into smaller blueprints and leave the big one as it is, perhaps refer to the larger epic blueprint in the smaller ones as "follow-up" blueprints or at the epic one we mention that it will be split and have additional blueprints to follow with details, etc.

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,319 @@
...
+headers. This could be useful here too, especially when parsing those headers
+to display our list of blueprints with the current status.
+
+The current list of items of our blueprint headers is below:
+
+ * Number: Usually, the blueprint number in the format BPXXX
+
+ * Title: A short descriptive title, limited to 80 characters
+
+ * Author: The author or authors of blueprint
```
pevogam, line 166: Perhaps we can clarify here about the specific format, i.e. LAST NAME, FIRST NAME or emails or everything that is meant with "authors".

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,319 @@
...
+
+ * Reviewers: All reviewers that approved and helped during the review process
+
+ * Created: Date string when the blueprint first draft was submitted. Please
+   use the following format: DD-MMM-YYYY.
```
pevogam, line 171: Why three M? Also, I would suggest that we use YYYY-MM-DD as an ISO standard that is more inclusive of American or other date formats that would start with the month and confuse themselves about this. I guess if it is not too late since we already have some blueprints out though.

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,319 @@
...
+
+ * Type: One of the types described during the previous section
```
pevogam, line 173: With capital or small letters or else? Maybe in the end for any additional questions we could just say "For more details on the formats, check some of the already defined blueprints." or something.

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,319 @@
...
+
+ * Status: One of the types described during the next section
+
+Here is an example of a header::
+
+   :Number: BP001
+   :Title: Configuration by convention
+   :Author: Beraldo Leal <bleal@redhat.com>
+   :Reviewers: Cleber Rosa, Lukáš Doktor and Plamen Dimitrov
+   :Created: 06-Dec-2019
```
pevogam, line 183: The date format here doesn't follow the one required above. Perhaps we could fix it at least for the example here even if we can't fix it for the already approved blueprints?

pevogam: Doesn't this blueprint itself have these headers in the end? Then it is already an example about itself.

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,319 @@
...
+ * Reviewers: All reviewers that approved and helped during the review process
+
+ * Created: Date string when the blueprint first draft was submitted. Please
+   use the following format: DD-MMM-YYYY.
+
+ * Type: One of the types described during the previous section
+
+ * Status: One of the types described during the next section
+
+Here is an example of a header::
```
pevogam, line 177: Ah better.

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,319 @@
...
+
+   :Number: BP001
+   :Title: Configuration by convention
+   :Author: Beraldo Leal <bleal@redhat.com>
+   :Reviewers: Cleber Rosa, Lukáš Doktor and Plamen Dimitrov
+   :Created: 06-Dec-2019
+   :Type: Epic Blueprint
+   :Status: WIP
```
pevogam, line 185: This status also cannot be found below as I guess the blueprint is outdated for the standards to require here. Also maybe it is best to first describe the statuses and complete all headers, then include an example right at the end.

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,319 @@
...
+
+Bellow a brief description of what you should write in each section:
+
+ * TL;DR: Should be a short description of your blueprint. Like an abstract. We
+   recommend writing this at the end of your first draft. This will give you a
+   better overview of it.
+
+ * Motivation: This should be the motivation of your proposed solution, not the
+   motivation of the blueprint itself. It describes the problem. Here, the
+   proposed solution should answer “why” this is needed.
```
pevogam, line 249: Not sure why "the proposed solution should answer" and what this means. I guess you mean "Here, you should answer "why" your solution is needed."

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,319 @@
...
+
+ * Backwards Compatibility: How is your proposal going to affect older versions
+   of Avocado? Should we deprecate some modules, classes, or methods? Are we
+   going to keep backwards compatibility or not?
+
+ * Security Implications: Do you have any concerns about security with your
+   proposed solution and what are they?
+
+ * How to Teach This: What is the best way to teach our devs and users your new
+   feature/solution?
```
pevogam, line 265: I am not sure if "teach" in the second place it is mentioned is too strong. I would rather say "What is the best way to inform our devs and users about your new feature/solution?" But just mentioning.

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,319 @@
...
+
+ * Related Issues: Here, you should mention Github links for both: a) current
+   open issues that are blocking while waiting for your BP and b) all open
+   issues that will render this BP as “implemented” when closed.
+
+   * Issues to address this BP
+
+   * Issues this BP will solve
```
pevogam, line 273: I am still confused about the double presence of items here - first there are a) and b) and then there is a list of two more items, perhaps the same a) and b) mentioned again. Can there be just one list with two items?

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,319 @@
...
+   your solution.
+
+Backwards Compatibility
+***********************
+
+So far, we are on our 3rth blueprint (BP003 was the last one). This BP000
+should have been released before those blueprints. So probably those three
+blueprints are not 100% compliance with this meta blueprint, and that is fine.
+We were learning on the fly. We don’t need to change any of those blueprints
+after BP000 gets approved.
```
pevogam, line 285: I see, this answered some of my questions above.

docs/source/blueprints/BP000.rst
```diff
@@ -0,0 +1,319 @@
...
+Security Implications
+*********************
+
+No security implications found so far.
+
+How to Teach This
+*****************
+
+Getting used to writing blueprints is not an easy task. And probably we are
+going to find non-planned issues with this process on the way.  The general
```
pevogam, line 296: I guess you mean "unplanned" or "not intended" or "unintentional"

pevogam (commented):
Left 20 more comments, big and small :) Sorry for making you wait.

docs/source/blueprints/BP000.rst
beraldoleal: MMM is equivalent the first three characters of the month name since it is more readable for humans. This is the same format as PEPs.

docs/source/blueprints/BP000.rst
beraldoleal: Why not? "06-Dec-2019" == DD-MMM-YYYY. What I'm missing here?

docs/source/blueprints/BP000.rst
beraldoleal: Not sure if I understood. What capital letter are you talking about?

docs/source/blueprints/BP000.rst
beraldoleal: Nice catch, thanks.

beraldoleal:
@pevogam sorry it took me a while to address your issues. I was busy with other tasks.

I fixed *most* of your comments, I would say 99%. Since this had enought time for review, I will send a v2 so we can move forward with this ASAP.

I hope all your major concerns where fixed on v2. Please let me known there.

beraldoleal:
Replaced by #4310.

docs/source/blueprints/BP000.rst
pevogam: Alright, fair enough.

docs/source/blueprints/BP000.rst
pevogam: I guess this is now outdated by your previous comments. Scratch that.

docs/source/blueprints/BP000.rst
pevogam: I mean what is the specific format of "Type" but I guess people will hopefully default to one-to-one copying of the previously written types exactly in the way they are written (capitalization, wording, etc.).

pevogam:
Alright, will now be able to move to v2. I am not sure why we have to move from one PR to the next for the same feature but I am discussing workflow again which is not part of this. Moving on.

PR #4270, opened 2020-10-15
Migration from imp to importlib

tiagohonorato:
Migrates from imp to importlib, as the former is depreciated since version 3.4.
As discussed in #3468, for now, only safeloader.py was altered.

### DIFF
--- a/avocado/core/safeloader.py
+++ b/avocado/core/safeloader.py
@@ -19,11 +19,11 @@ Safe (AST based) test loader module utilities
 
 import ast
 import collections
-import importlib
 import json
 import os
 import re
 import sys
+from importlib.machinery import PathFinder
 
 from ..utils import data_structures
 
@@ -497,8 +497,8 @@ def _examine_class(path, class_name, match, target_module, target_class,
 
             modules_paths = [parent_path,
                              os.path.dirname(module.path)] + sys.path
-            found_ppath = importlib.machinery.PathFinder.find_spec(
-                parent_module, modules_paths).origin
+            found_ppath = PathFinder.find_spec(parent_module,
+                                               modules_paths).origin
             _info, _disabled, _match = _examine_class(found_ppath,
                                                       parent_class,
                                                       match,
@@ -630,8 +630,8 @@ def find_python_tests(module_name, class_name, determine_match, path):
 
             modules_paths = [parent_path,
                              os.path.dirname(module.path)] + sys.path
-            found_ppath = importlib.machinery.PathFinder.find_spec(
-                parent_module, modules_paths).origin
+            found_ppath = PathFinder.find_spec(parent_module,
+                                               modules_paths).origin
             if found_ppath is None:
                 continue
             _info, _dis, _python_test = _examine_class(found_ppath,
--- a/avocado/core/safeloader.py
+++ b/avocado/core/safeloader.py
@@ -19,7 +19,7 @@ Safe (AST based) test loader module utilities
 
 import ast
 import collections
-import imp
+import importlib
 import json
 import os
 import re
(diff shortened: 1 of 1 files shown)


avocado/core/safeloader.py
```diff
@@ -630,9 +630,9 @@ def find_python_tests(module_name, class_name, determine_match, path):
 
             modules_paths = [parent_path,
                              os.path.dirname(module.path)] + sys.path
-            try:
-                _, found_ppath, _ = imp.find_module(parent_module, modules_paths)
-            except ImportError:
+            found_ppath = importlib.machinery.PathFinder.find_spec(
+                parent_module, modules_paths).origin
+            if found_ppath is None:
```
richtja, line 635: Hi @tiagohonorato, I don't understand the purpose of this If statements.

avocado/core/safeloader.py
```diff
@@ -497,8 +497,8 @@ def _examine_class(path, class_name, match, target_module, target_class,
 
             modules_paths = [parent_path,
                              os.path.dirname(module.path)] + sys.path
-            _, found_ppath, _ = imp.find_module(parent_module,
-                                                modules_paths)
+            found_ppath = importlib.machinery.PathFinder.find_spec(
```
beraldoleal, line 500: Since your are using only Pathfinder, my suggestion is to change the import to:

```python
from importlib.machinery import PathFinder
```

So this line will be shorter:

```python
found_ppath = PathFinder.find_spec(...
```

avocado/core/safeloader.py
beraldoleal: Since the try.. except was removed, I guess it is used as a replacement, to continue or not.

beraldoleal (changes requested):

avocado/core/safeloader.py
richtja: Yes, of course :man_facepalming: thanks for explanation.

tiagohonorato: Hi @richtja and @beraldoleal, thanks for the review! Adding on @beraldoleal's comment, the `find_spec` method does not raise `ImportError`, instead it returns `None`. So, that's why I replaced the try... except with the if statement.

avocado/core/safeloader.py
tiagohonorato: Sure @beraldoleal, that`s a good point. I will change it!

avocado/core/safeloader.py
```diff
@@ -497,8 +497,8 @@ def _examine_class(path, class_name, match, target_module, target_class,
 
             modules_paths = [parent_path,
                              os.path.dirname(module.path)] + sys.path
-            _, found_ppath, _ = imp.find_module(parent_module,
-                                                modules_paths)
+            found_ppath = PathFinder.find_spec(parent_module,
+                                               modules_paths).origin
```
beraldoleal, line 501: IIUC, according to the documentation, this method can return None, and you trying to access the `.origin` property of a `None` value. This will raise exceptions.

You should properly check if the return is `None` or not, before using.

avocado/core/safeloader.py
```diff
@@ -630,9 +630,9 @@ def find_python_tests(module_name, class_name, determine_match, path):
 
             modules_paths = [parent_path,
                              os.path.dirname(module.path)] + sys.path
-            try:
-                _, found_ppath, _ = imp.find_module(parent_module, modules_paths)
-            except ImportError:
+            found_ppath = PathFinder.find_spec(parent_module,
+                                               modules_paths).origin
```
beraldoleal, line 634: same here.

beraldoleal (changes requested):

avocado/core/safeloader.py
tiagohonorato: Hi @beraldoleal, thanks for pointing that out :man_facepalming:, you are absolutely right. I will add an if statement here to call `_examine_class` only if `find_spec` does not return `None`.

avocado/core/safeloader.py
tiagohonorato: Sure, I will maintain the `continue`, but checking for `find_spec` output instead. Thanks!

willianrampazzo (approved):
I think we are good to go!

beraldoleal (approved):
LGTM, thanks.

richtja (approved):
LGTM, thanks

PR #4271, opened 2020-10-19
Added a library function which selinux in enforcing mode

PraveenPenguin:
added utility which check if system is running in selinux enforcing mode if not it enable it
and return True otherwise False

### DIFF
--- a/avocado/utils/linux.py
+++ b/avocado/utils/linux.py
@@ -60,3 +60,15 @@ def is_selinux_enforcing():
     if '1' in genio.read_one_line('/sys/fs/selinux/enforce'):
         return True
     return False
+
+
+def enable_selinux_enforcing():
+    """
+    Enable  SELinux Enforcing in system
+
+    :return: True if SELinux enable in enforcing mode, False if not enabled
+    """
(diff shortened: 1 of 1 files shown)


avocado/utils/linux.py
```diff
@@ -60,3 +60,17 @@ def is_selinux_enforcing():
...
+
+def enable_selinux_enforcing():
+    """
+    Enable  SELinux Enforcing in system
+
+    :return: True if SELinux enable in enforcing mode, False if not enabled
+    """
+    if is_selinux_enforcing():
+        return True
+    genio.write_one_line('/sys/fs/selinux/enforce', '1')
```
beraldoleal, line 73: Since the kernel will handle properly to write or not, I would suggest removing the lines 71-72.

beraldoleal (changes requested):

avocado/utils/linux.py
PraveenPenguin: ok sure , make sense will do

willianrampazzo (approved):
With the change requested by @beraldoleal, LGTM.

beraldoleal (approved):
LGTM, thanks.

PR #4277, opened 2020-10-26
status_server: split status_server_uri logic

### LINKED ISSUE #4269: status server uri should be different from listen
nrunner.status_server_uri should not be used to both, server and client. We should have different settings for configuring the client.

beraldoleal:
Sometimes we can have a status_server running in a different machine or
port. This patch will introduce the status_server_listen option that
will be used when listening the status, while status_server_uri will be
used to connect to a status server.

This fixes #4269

### DIFF
--- a/avocado/plugins/runner_nrunner.py
+++ b/avocado/plugins/runner_nrunner.py
@@ -50,7 +50,17 @@ class RunnerInit(Init):
                                  help_msg=help_msg,
                                  key_type=bool)
 
-        help_msg = 'URI for the status server, usually a "HOST:PORT" string'
+        help_msg = ('URI for listing the status server. Usually '
+                    'a "HOST:PORT" string')
+        settings.register_option(section=section,
+                                 key='status_server_listen',
+                                 default='127.0.0.1:8888',
+                                 metavar="HOST:PORT",
+                                 help_msg=help_msg)
+
+        help_msg = ('URI for connecting to the status server, usually '
+                    'a "HOST:PORT" string. Use this if your status server '
+                    'is in another host, or different port')
         settings.register_option(section=section,
                                  key='status_server_uri',
                                  default='127.0.0.1:8888',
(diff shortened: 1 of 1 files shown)


willianrampazzo:
I have not tested your code, just did a `grep` for `status_server_uri` on Avocado's code and found an entry that may not be covered by your code: https://github.com/avocado-framework/avocado/blob/ee7e9f8fbeb9d18a25811a095f2d52640cf660e0/avocado/core/utils.py#L72

Maybe you want to double-check that.

beraldoleal:
> I have not tested your code, just did a `grep` for `status_server_uri` on Avocado's code and found an entry that may not be covered by your code:
> 
> https://github.com/avocado-framework/avocado/blob/ee7e9f8fbeb9d18a25811a095f2d52640cf660e0/avocado/core/utils.py#L72
> 
> Maybe you want to double-check that.

I saw that and AFAICT, this is to be like this, since this information is saved along the tasks and used for connecting and not for listening.

willianrampazzo:
> > I have not tested your code, just did a `grep` for `status_server_uri` on Avocado's code and found an entry that may not be covered by your code:
> > https://github.com/avocado-framework/avocado/blob/ee7e9f8fbeb9d18a25811a095f2d52640cf660e0/avocado/core/utils.py#L72
> > 
> > Maybe you want to double-check that.
> 
> I saw that and AFAICT, this is to be like this since this information is saved along with the tasks and used for connecting to the status server and not for listening.

I see how your code does not affect the code I mentioned, but I got confused.

Imagine I'm creating some code on Avocado core/plugins and I'm using this `resolution_to_tasks` function for any reason, but my code is meant to use a custom status server with an address received as a parameter. The end result here will be the default status server being created anyway. As long as the address and port do not conflict with the custom status server address and port, that is fine, but I don't think that is necessary. Avocado will be creating a resource that won't be used. Maybe your PR is not the place to fix this situation, but we should keep that in mind to better organize things later.

beraldoleal:
Yes, in fact, Avocado is creating the server, because I didn't wanna be too agressive. One possible solution is to remove the listen in another PR. So, yes, I agree with you.

clebergnu (approved):
Thanks @beraldoleal this is definitely an improvement.

In the near future we'll probably have to deal with spawners giving hints to tasks about the uri after some type of mapping (think of a task running on a vm connected via a bridged network).

willianrampazzo:
> Yes, in fact, Avocado is creating the server, because I didn't wanna be too agressive. One possible solution is to remove the listen in another PR. So, yes, I agree with you.

Okie dokie, approving!

willianrampazzo (approved):

Issue #4246, opened 2020-09-22
Automated testing of examples in the documentation 


richtja:
Before the LTS release, we spent a lot of time testing and fixing the documentation examples to make them compatible with changes in the avocado code. Because of that, we should create automated tests for checking the compatibility between documentation and code. This issue should bring a discussion about what will be the best solution to fo this problem. Here are some possible solutions:

- Use sphinx with doctest. This does exactly what we want but it can work just with python interpreter and for the command-line examples we have to use `subprocess` module.

- For command-line examples uses some doctest extensions like `doctest-cli` from people who were dealing with the same problem. But here we cannot ensure reliability of such tools. 

- We can create interaction tests for command-line examples and then these tests import into the documentation by sphinx.

These are ideas about how we can deal with this problem, but I am not familiar with documentation testing. So any ideas are welcome.

willianrampazzo:
I was thinking when this problem appeared, before the LTS release, can't we import the tests from `examples/` dir? Then, we test only once, and the user will have the test ready to run when they want.

If this is possible, but we find that many examples are not in the directory right now, we can create something like `examples/docs` and reference that as the examples from our documentation.

Just throwing the idea here :)

richtja:
> I was thinking when this problem appeared, before the LTS release, can't we import the tests from examples/ dir? Then, we test only once, and the user will have the test ready to run when they want.

@willianrampazzo Thanks for the reply, because this is exactly what I mean by the third option; maybe I didn't express myself clearly. 

Here I can see a problem that we have to import all those examples into rst documentation, and I think that won't be easy to maintain in the future.  But maybe I am wrong.

willianrampazzo:
> > I was thinking when this problem appeared, before the LTS release, can't we import the tests from examples/ dir? Then, we test only once, and the user will have the test ready to run when they want.
> 
> @willianrampazzo Thanks for the reply, because this is exactly what I mean by the third option; maybe I didn't express myself clearly.
> 

Okay, got it!

> Here I can see a problem that we have to import all those examples into rst documentation, and I think that won't be easy to maintain in the future. But maybe I am wrong.

I think keeping the tests in one place and test them will make things easy for the users and us whether this is on the `examples` directory or into the documentation.

The idea to use the tests from the `examples` directory is to have the documentation examples ready to be executed and avoid the copy/paste from the documentation to a test file and then running on the user side.

[...]

PR #4283, opened 2020-10-30
Testing of scripts inside documentation

richtja:
This is a draft for documentation scripts testing.
It is about testing every script inside avocado documentation and keep documentation updated and also create runnable examples for users which will correspond with documentation #4246.  This is an example of how it could look like. I will be glad for any comments before I will start with creating scripts from the documentation.


willianrampazzo:
Humm, I was thinking about something much simpler, but, maybe, what I was thinking is not possible, or there is a better way of doing that.

My idea was to have the tests under `examples/docs` as avocado python tests and a script under `selftests` to run and test them. Those tests would be imported into the documentation as examples.

I know there are tools to test code into the documentation, but I don't know if we need all those features.

Anyway, let me know if what I'm proposing here makes no sense :)

beraldoleal:
Having scripts that will run the current avocado output and replacing the documentation could be nice but at same time I'm asking if we are not putting the issues "under the carpet".

I was thinking similar to @willianrampazzo . Did you tried that before? Found some issues?

Also, I would try to convert the bash scripts to python too.

richtja:
> My idea was to have the tests under `examples/docs` as avocado python tests and a script under `selftests` to run and test them. Those tests would be imported into the documentation as examples.

Thanks, @willianrampazzo, actually I can't see how to use the tests under `examples/docs` as avocado python tests, because most of the examples inside documentation are examples of Avocado command-line and I want to create examples scripts which users can run and will be same as Avocado command-line. But you are right that we can test these scripts under `selftests` by avocado and remove the files with exit statuses.

richtja:
@beraldoleal Thanks for your comments.

> Having scripts that will run the current avocado output and replacing the documentation could be nice but at same time I'm asking if we are not putting the issues "under the carpet".

think that this is not putting the issues "under the carpet" because we would test every script inside the documentation if it's not broken. 

> I was thinking similar to @willianrampazzo . Did you tried that before? Found some issues?

Do you mean use tools to test code into the documentation? I did some research and look for possibilities and didn't find anything without problems. We had a discussion about it here # 4246

> Also, I would try to convert the bash scripts to python too.

Yes I agree and I will do it.

beraldoleal (changes requested):
Hi @richtja ,

My only suggestion by now is to replace the bash script with a python script.

beraldoleal:
@richtja let me know what do you think about having the python script. If you think that is not possible to add before the next version, no worries.

richtja:
@beraldoleal sorry that I didn't answer. I was focusing on the other tasks and leave this behind. I already replace the bash script with a python script, but when I started cover the documentation I ran into a lot of problems which have to be solved for different types of examples in documentation. Because of that I didn't push the solution yet.

PR #4285, opened 2020-11-03
Add deprecation warning messages to "string boolean" args

beraldoleal:
This is part of a big effort to convert all "string boolean" args to real boolean. Please check BP001 for details.

### DIFF
--- a/avocado/plugins/jsonresult.py
+++ b/avocado/plugins/jsonresult.py
@@ -19,6 +19,7 @@ JSON output module.
 
 import json
 import os
+import warnings
 
 from avocado.core.output import LOG_UI
 from avocado.core.parser import FileOrStdoutAction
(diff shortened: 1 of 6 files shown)


pevogam:
Hi @beraldoleal, I thought all arguments are fully converted and BP0001 fully implemented by now. If not is there a list where the still unconverted arguments are tracked?

beraldoleal:
> Hi @beraldoleal, I thought all arguments are fully converted and BP0001 fully implemented by now. If not is there a list where the still unconverted arguments are tracked?

I was tracking on a personal list, and afaict, those are the left ones.

pevogam (approved):

richtja (approved):
LGTM

willianrampazzo (commented):
Isn't it better to collapse all commits and avoid failures as the unit test fix is applied in the end?

beraldoleal:
> Isn't it better to collapse all commits and avoid failures as the unit test fix is applied in the end?
The unit test fix is applied at the begin. I prefer to have commits separeted in this case to avoid a big change.

willianrampazzo:
> > Isn't it better to collapse all commits and avoid failures as the unit test fix is applied in the end?
> 
> The unit test fix is applied at the begin. I prefer to have commits separeted in this case to avoid a big change.

Okay, I misread the order.

willianrampazzo (approved):

PR #4288, opened 2020-11-04
Yaml to mux documentation fix

richtja:
Inside documentation of yaml to mux plugin is a problem with None type, which
is passed to parameters as a string due to distinguish between empty values and
default values. This commit should specify this behavior.

### DIFF
--- a/docs/source/plugins/optional/varianter_yaml_to_mux.rst
+++ b/docs/source/plugins/optional/varianter_yaml_to_mux.rst
@@ -140,8 +140,9 @@ The environment created for the nodes ``fedora`` and ``osx`` are:
 
 Note that due to different usage of key and values in environment we disabled
 the automatic value conversion for keys while keeping it enabled for values.
-This means that the value can be of any YAML supported value, eg. bool, None,
-list or custom type, while the key is always string.
+This means that the key is always a string and the value can be YAML value,
+eg. bool, list, custom type, or string. Please be aware that the None type
+is unsupported and cannot be provided in yaml as null.
 
 Variants
 --------


docs/source/plugins/optional/varianter_yaml_to_mux.rst
```diff
@@ -140,8 +140,8 @@ The environment created for the nodes ``fedora`` and ``osx`` are:
 
 Note that due to different usage of key and values in environment we disabled
 the automatic value conversion for keys while keeping it enabled for values.
-This means that the value can be of any YAML supported value, eg. bool, None,
-list or custom type, while the key is always string.
+This means that the key is always string and the value can be of any YAML supported value, eg. bool, None,
```
pevogam, line 143: "the key is always *a* string"

This also seems to not follow the line breaking from the original document.

docs/source/plugins/optional/varianter_yaml_to_mux.rst
```diff
@@ -140,8 +140,8 @@ The environment created for the nodes ``fedora`` and ``osx`` are:
...
+list or custom type, be aware that the None type will be passed to parameters as a string.
```
pevogam, line 144: This is best put as a separate sentence: "list or custom type. Please be aware that..."

docs/source/plugins/optional/varianter_yaml_to_mux.rst
harish-24: Expecting None to be removed from here as there is no way it can be provided. You could add that as a limitation explicitly in the next sentence. Something like
`Please be aware that the None type can be passed to parameters only as a string`

richtja:
@pevogam, @harish-24 Thanks for your comments, I will push the changes here.

docs/source/plugins/optional/varianter_yaml_to_mux.rst
```diff
@@ -140,8 +140,9 @@ The environment created for the nodes ``fedora`` and ``osx`` are:
 
 Note that due to different usage of key and values in environment we disabled
 the automatic value conversion for keys while keeping it enabled for values.
-This means that the value can be of any YAML supported value, eg. bool, None,
-list or custom type, while the key is always string.
+This means that the key is always a string and the value can be of any YAML
+supported value, eg. bool, None, list or custom type. Please be aware that
```
harish-24, line 144: This is contradicting, it says supported value but None is not supported. As mentioned earlier, expecting it to be removed from this line for better clarity.

pevogam: So I guess just dropping `None` from the items above should be enough.

harish-24: @pevogam Although it does convey the meaning, IMO it is not logical to support None as a "string" because it is a separate datatype. `"null"` or even `"None"` can be used now in yaml but it does not qualify to supporting a `None` type, hence asked for its removal as a supported type. I will leave it to @richtja for his inputs w.r.t this.

pevogam: Good points, fully agreed with you. I was just suggesting a minimal diff to achieve a compromise but if @richtja is willing to take an extra step which is more YAML compatible it would be great.

richtja: @harish-24 I think you are right that we shouldn't specify `None` as a supported value. Also, we shouldn't say that we support any YAML values. What about this:

`This means that the key is always a string and the value can be YAML value, eg. bool, list, custom type, or string. Please be aware that the None type can be passed to parameters only as a string.`

harish-24: @richtja The first part is good. But second does not make sense. As a matter of fact None type cannot be passed at all. "null" or "None" is still a valid string so I don't think "can be passed to parameters only as a string" makes sense. How about this?
```
Please be aware that the None type is unsupported and cannot be provided in yaml as null. 
```

On further note if we can hack around the code to handle `"null"` (string) specifically and replace it with None type then it makes sense to document with `"null"` (string) specifically.

richtja: My point was to make a hint for users who want to use None, but yes I think you are right, this will be the most clear solution.

And yes we can specify "null" or "None" string, when we hack around the code to handle "null" string.

richtja:
@harish-24 I did your proposed changes, please have a look.

harish-24 (approved):
LGTM, Thanks @richtja !

willianrampazzo:
I restarted the job for Python 3.8 as it was failing `./optional_plugins/varianter_cit/tests/test_basic.py:Run.test:  FAIL (0.98 s)` which does not seems to be related to this PR.

willianrampazzo (approved):

pevogam (approved):

Issue #3470, opened 2020-01-08
Use symbolic `os.path.pardir` instead of `..`


willianrampazzo:
To support future O.S. compatibility, places where the parent path is referenced as `..` needs to be changed to 'os.path.pardir`.

nasirhm:
I would like to work on it.

tiagohonorato:
Hi @willianrampazzo  and @nasirhm, I noticed that there are also some references to grandparent directory `...`:

```
avocado/utils/software_manager/backends/dnf.py:3:from ... import process
avocado/utils/software_manager/backends/zypper.py:4:from ... import path as utils_path
avocado/utils/software_manager/backends/zypper.py:5:from ... import process
avocado/utils/software_manager/backends/rpm.py:5:from ... import path as utils_path
avocado/utils/software_manager/backends/rpm.py:6:from ... import process
avocado/utils/software_manager/backends/yum.py:8:from ... import data_factory
avocado/utils/software_manager/backends/yum.py:9:from ... import path as utils_path
avocado/utils/software_manager/backends/yum.py:10:from ... import process
avocado/utils/software_manager/backends/dpkg.py:4:from ... import path as utils_path
avocado/utils/software_manager/backends/dpkg.py:5:from ... import process
avocado/utils/software_manager/backends/apt.py:6:from ... import path as utils_path
avocado/utils/software_manager/backends/apt.py:7:from ... import process
```
Do you think those could be an issue in the future? If so, since `os.path` does not go beyond the parent directory, changing from `...` to `....utils` would avoid future compatibility issues?

willianrampazzo:
> Hi @willianrampazzo and @nasirhm, I noticed that there are also some references to grandparent directory `...`:
> 
> ```
> avocado/utils/software_manager/backends/dnf.py:3:from ... import process
> avocado/utils/software_manager/backends/zypper.py:4:from ... import path as utils_path
> avocado/utils/software_manager/backends/zypper.py:5:from ... import process
> avocado/utils/software_manager/backends/rpm.py:5:from ... import path as utils_path
> avocado/utils/software_manager/backends/rpm.py:6:from ... import process
> avocado/utils/software_manager/backends/yum.py:8:from ... import data_factory
> avocado/utils/software_manager/backends/yum.py:9:from ... import path as utils_path
> avocado/utils/software_manager/backends/yum.py:10:from ... import process
> avocado/utils/software_manager/backends/dpkg.py:4:from ... import path as utils_path
[...]

PR #4289, opened 2020-11-04
Cross-platform reference to current and parent directories

tiagohonorato:
Changes string references to parent `'..'` and current `'.'` directories to cross-platform using `os.path.pardir` and `os.path.curdir` as discussed in  #3470

### DIFF
--- a/optional_plugins/varianter_yaml_to_mux/tests/test_unit.py
+++ b/optional_plugins/varianter_yaml_to_mux/tests/test_unit.py
@@ -11,8 +11,8 @@ from avocado.core import parameters, tree
 from avocado.utils import astring
 from avocado_varianter_yaml_to_mux import mux
 
-BASEDIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), '..')
-BASEDIR = os.path.abspath(BASEDIR)
+BASEDIR = os.path.dirname(os.path.abspath(__file__))
+BASEDIR = os.path.abspath(os.path.join(BASEDIR, os.path.pardir))
 
 
 def combine(leaves_pools):
--- a/selftests/__init__.py
+++ b/selftests/__init__.py
@@ -7,7 +7,8 @@ import unittest
 import pkg_resources
 
 #: The base directory for the avocado source tree
-BASEDIR = os.path.abspath(os.path.join(os.path.dirname(os.path.abspath(__file__)), '..'))
+BASEDIR = os.path.dirname(os.path.abspath(__file__))
+BASEDIR = os.path.abspath(os.path.join(BASEDIR, os.path.pardir))
 
 #: The name of the avocado test runner entry point
 AVOCADO = os.environ.get("UNITTEST_AVOCADO_CMD",
--- a/optional_plugins/golang/avocado_golang/__init__.py
+++ b/optional_plugins/golang/avocado_golang/__init__.py
@@ -57,7 +57,7 @@ def find_files(path, recursive=True):
                 matches.append(os.path.join(root, filename))
         return matches
 
-    if path != '.':
+    if path != os.path.curdir:
         pattern = os.path.join(path, pattern)
     return glob.iglob(pattern)
 


optional_plugins/varianter_yaml_to_mux/tests/test_unit.py
```diff
@@ -11,8 +11,8 @@
 from avocado.utils import astring
 from avocado_varianter_yaml_to_mux import mux
 
-BASEDIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), '..')
-BASEDIR = os.path.abspath(BASEDIR)
+BASEDIR = os.path.dirname(os.path.abspath(__file__))
+BASEDIR = os.path.abspath(os.path.join(BASEDIR, os.path.pardir))
```
beraldoleal, line 15: Unless I'm missing something, in line 14 you are using `os.path.abspath` that will return already an absolute path. os.path.dirname will keep the absolute path. IMO, the second `abspath` (line 15) is unecessary.

What do you think?

selftests/__init__.py
```diff
@@ -7,7 +7,8 @@
 import pkg_resources
 
 #: The base directory for the avocado source tree
-BASEDIR = os.path.abspath(os.path.join(os.path.dirname(os.path.abspath(__file__)), '..'))
+BASEDIR = os.path.dirname(os.path.abspath(__file__))
+BASEDIR = os.path.abspath(os.path.join(BASEDIR, os.path.pardir))
```
beraldoleal, line 11: same here.

beraldoleal (changes requested):

optional_plugins/varianter_yaml_to_mux/tests/test_unit.py
tiagohonorato: Hi @beraldoleal, that's a good question. From what I understand, the second `abspath` is to evaluate the absolute path after joining with `'..'` (or other O.S. equivalent). For instance, `os.path.join` would return something like `/home/user/avocado/optional_plugins/varianter_yaml_to_mux/tests/..` then, after applying `os.path.abspath` it would return `/home/user/avocado/optional_plugins/varianter_yaml_to_mux`.

Let me know if that answers your question.

beraldoleal: I got your point/example and now I can see why is there. Thanks. But yes, I think that maybe could improve readability a little bit:
```python
from os.path import abspath, dirname

BASEDIR = dirname(dirname(abspath(__file__))))
```

Not sure if will works, didn't tested. What do you think ?

willianrampazzo: Considering we are using `os.path.` everywhere in the project, I don't see a problem using it here. I think importing and using the would confuse more based on what we already do in the project.

This does not mean we can not open an issue to discuss if we should move all the `os.path.` entries to specific functions.

willianrampazzo (approved):
Looks good to me, thanks @tiagohonorato!

optional_plugins/varianter_yaml_to_mux/tests/test_unit.py
beraldoleal: You have a point, I'm not going to block the PR because of this, but to be honest I don't like much the way it is. We need to get "the parent of the parent" and we using 2 lines with 4 "methods".

willianrampazzo: Yep, I agree if you!

beraldoleal (approved):

optional_plugins/varianter_yaml_to_mux/tests/test_unit.py
tiagohonorato: > I got your point/example and now I can see why is there. Thanks. But yes, I think that maybe could improve readability a little bit:
> 
> ```python
> from os.path import abspath, dirname
> 
> BASEDIR = dirname(dirname(abspath(__file__))))
> ```
> 
> Not sure if will works, didn't tested. What do you think ?

This seems to return the same result from the initial testing I've done. I would say this would be more readable, even without importing `abspath` and `dirname` separately. 

> Considering we are using `os.path.` everywhere in the project, I don't see a problem using it here. I think importing and using the single funciton would confuse more based on what we already do in the project.
> 
> This does not mean we can not open an issue to discuss if we should move all the `os.path.` entries to specific functions.

That's a good point indeed. Another alternative would be to use `pathlib` alongside `os.path`, but this might be a bit confusing as well, what do you think @beraldoleal and @willianrampazzo? So far, the following seems to have a similar result:

```python
import pathlib

BASEDIR = pathlib.Path(__file__).resolve().parents[1]
```

> You have a point, I'm not going to block the PR because of this, but to be honest I don't like much the way it is. We need to get "the parent of the parent" and we are using 2 lines with 4 "methods".

I feel the same. I can open an issue if you all agree...

willianrampazzo: > > Considering we are using `os.path.` everywhere in the project, I don't see a problem using it here. I think importing and using the single funciton would confuse more based on what we already do in the project.
> > This does not mean we can not open an issue to discuss if we should move all the `os.path.` entries to specific functions.
> 
> That's a good point indeed. Another alternative would be to use `pathlib` alongside `os.path`, but this might be a bit confusing as well, what do you think @beraldoleal and @willianrampazzo? So far, the following seems to have a similar result:
> 
> ```python
> import pathlib
> 
> BASEDIR = pathlib.Path(__file__).resolve().parents[1]
> ```
> 

We already use `os.path` in the project. Having another package to solve the same already solved by `os.path` seems to me just transferring the action from one package to another.

> > You have a point, I'm not going to block the PR because of this, but to be honest I don't like much the way it is. We need to get "the parent of the parent" and we are using 2 lines with 4 "methods".
> 
> I feel the same. I can open an issue if you all agree...

Discussion is always good to document the pros and cons of some approaches. If we decide what we have now is good enough, we can document that and close the issue.

PR #4295, opened 2020-11-06
Test suite specific loader

richtja:
We have multiple test suites inside one Job and each of them can have different
configurations from the user. Unfortunately, we have just one instance of the
test loader which is not able to deal with different configurations of test
suites. This commit creates for each test suite instance its own test loader
which will work with the right configuration.

### DIFF
--- a/avocado/core/loader.py
+++ b/avocado/core/loader.py
@@ -75,23 +75,42 @@ class LoaderUnhandledReferenceError(LoaderError):
 
 class TestLoaderProxy:
 
+    class PluginsRegister:
+
+        def __init__(self):
+            self.registered_plugins = []
+
+        def register_plugin(self, plugin):
+            try:
+                if issubclass(plugin, TestLoader):
+                    if plugin not in self.registered_plugins:
+                        self.registered_plugins.append(plugin)
+                else:
+                    raise ValueError
+            except ValueError:
+                raise InvalidLoaderPlugin("Object %s is not an instance of "
+                                          "TestLoader" % plugin)
+
+        def clear_plugins(self):
+            self.registered_plugins = []
+
     def __init__(self):
         self._initialized_plugins = []
-        self.registered_plugins = []
         self.reference_plugin_mapping = {}
         self._label_mapping = None
         self._decorator_mapping = None
 
-    def register_plugin(self, plugin):
-        try:
-            if issubclass(plugin, TestLoader):
-                if plugin not in self.registered_plugins:
-                    self.registered_plugins.append(plugin)
-            else:
-                raise ValueError
-        except ValueError:
-            raise InvalidLoaderPlugin("Object %s is not an instance of "
-                                      "TestLoader" % plugin)
+    @property
+    def registered_plugins(self):
+        return register.registered_plugins
+
+    @staticmethod
+    def register_plugin(plugin):
+        register.register_plugin(plugin)
+
+    @staticmethod
+    def clear_plugins():
+        register.clear_plugins()
 
     def load_plugins(self, config):
         def _good_test_types(plugin):
@@ -261,7 +280,8 @@ class TestLoaderProxy:
         self._update_mappings()
         return tests
 
-    def load_test(self, test_factory):
+    @staticmethod
+    def load_test(test_factory):
         """
         Load test from the test factory.
 
@@ -305,9 +325,6 @@ class TestLoaderProxy:
         test_instance = test_class(**test_parameters)
 
         return test_instance
-
-    def clear_plugins(self):
-        self.registered_plugins = []
 
 
 class TestLoader:
@@ -912,4 +929,4 @@ class TapLoader(SimpleFileLoader):
                                executable=test_path)
 
 
-loader = TestLoaderProxy()
+register = TestLoaderProxy.PluginsRegister()
--- a/avocado/core/nrunner_avocado_instrumented.py
+++ b/avocado/core/nrunner_avocado_instrumented.py
@@ -2,7 +2,8 @@ import multiprocessing
 import tempfile
 import time
 
-from . import job, loader, nrunner
+from . import job, nrunner
+from .loader import TestLoaderProxy
 from .test import TestID
 from .tree import TreeNode
 
@@ -44,7 +45,7 @@ class AvocadoInstrumentedTestRunner(nrunner.BaseRunner):
                          'run.results_dir': tempfile.mkdtemp(),
                          }]
 
-        instance = loader.loader.load_test(test_factory)
+        instance = TestLoaderProxy.load_test(test_factory)
         instance.run_avocado()
         state = instance.get_state()
         # This should probably be done in a translator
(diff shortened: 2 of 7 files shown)


pevogam (commented):
This is an awesome idea @richtja! I guess this will need some elaborate testing but it really gives some more sense to the multiple test suite architecture.

beraldoleal:
Hi @richtja this is nice, super cool.

But, please help me with something: I'm might be wrong, but I had understood from our conversations that after LTS, all the legacy runner and loader stuff would be removed. Probably I'm missing something.

pevogam:
Hi @beraldoleal,

> But, please help me with something: I'm might be wrong, but I had understood from our conversations that after LTS, all the legacy runner and loader stuff would be removed. Probably I'm missing something.

I still it is still nice if this could be added and at least backported for the serial runner as it is still used by many people that could benefit from multiple test suites with nonhomogeneous runners. Also, I am still not sure why the presence of a parallel runner should necessarily imply the serial runner is obsolete but I know I am digging into something that was already decided.

beraldoleal:
> Hi @beraldoleal,
> 
> > But, please help me with something: I'm might be wrong, but I had understood from our conversations that after LTS, all the legacy runner and loader stuff would be removed. Probably I'm missing something.
> 
> I still it is still nice if this could be added and at least backported for the serial runner as it is still used by many people that could benefit from multiple test suites with nonhomogeneous runners. Also, I am still not sure why the presence of a parallel runner should necessarily imply the serial runner is obsolete but I know I am digging into something that was already decided.

My only concern is that this could have some impact on the decision to migrate to the new runner. Users will see a new "feature" on the legacy and will have the tendency to postpone the migration. But again, I'm just pointing out, and hopefully I'm not create to much noise here.

If it is really important and it is a bug, and IIUC, then yes, I agree with your @pevogam, probably we will need to backport this.

pevogam:
> > I still it is still nice if this could be added and at least backported for the serial runner as it is still used by many people that could benefit from multiple test suites with nonhomogeneous runners. Also, I am still not sure why the presence of a parallel runner should necessarily imply the serial runner is obsolete but I know I am digging into something that was already decided.
> 
> My only concern is that this could have some impact on the decision to migrate to the new runner. Users will see a new "feature" on the legacy and will have the tendency to postpone the migration. But again, I'm just pointing out, and hopefully I'm not create to much noise here.
> 
> If it is really important and it is a bug, and IIUC, then yes, I agree with your @pevogam, probably we will need to backport this.

Actually yes, you have more right here since the main point of backporting to the LTS is for stability reasons and fixes only, no new features.

So I guess it boils down to the question whether parallelization should be enforced for everyone and thus they should be "stimulated" to move away from older versions or not. Forgive me if I forgot something critical but was there a way to use the parallel runner for serial test running tasks?

Of course, we could move this discussion to another place since I don't want to kidnap the PR's main topic here and my last question got slightly off-topic.

willianrampazzo:
>Forgive me if I forgot something critical but was there a way to use the parallel runner for serial test running tasks?
> 

Hey @pevogam, not joining the discussion, but just giving a pointer here. Check the following option:

```
435:nrunner.max_parallel_tasks
436-~~~~~~~~~~~~~~~~~~~~~~~~~~
437:Number of maximum number tasks running in parallel. You can disable
438:parallel execution by setting this to 1. Defaults to the amount of
439-CPUs on this machine.
440-
441-* Default: 8
442-* Type: <class 'int'>
```

Command-line parameter:

```
  --nrunner-max-parallel-tasks NRUNNER.MAX_PARALLEL_TASKS
                        Number of maximum number tasks running in parallel. You can disable parallel execution by
                        setting this to 1. Defaults to the amount of CPUs on this machine.
```

richtja:
> > Hi @beraldoleal,
> > > But, please help me with something: I'm might be wrong, but I had understood from our conversations that after LTS, all the legacy runner and loader stuff would be removed. Probably I'm missing something.
> > 
> > 
> > I still it is still nice if this could be added and at least backported for the serial runner as it is still used by many people that could benefit from multiple test suites with nonhomogeneous runners. Also, I am still not sure why the presence of a parallel runner should necessarily imply the serial runner is obsolete but I know I am digging into something that was already decided.
> 
> My only concern is that this could have some impact on the decision to migrate to the new runner. Users will see a new "feature" on the legacy and will have the tendency to postpone the migration. But again, I'm just pointing out, and hopefully I'm not create to much noise here.

@beraldoleal Thanks for your comments.

I created this because I wasn't able to run multiple external runners on the legacy runner in one job and I am still using legacy runner because AFAIK the nrunner don't support avocado-vt tests which I also want to run in the same job. I don't see this as a new feature, I see it as fixing the expected behavior. But you are right that I didn't consider that this can postpone the migration.

Actually we still didn't decide how the compatibility of avocado and avocado-vt will look like after we remove the legacy runner.

beraldoleal:
Maybe we could focus on make avocado-vt running on the next runner. I'm not sure. @clebergnu any comments here?

pevogam:
Hi @willianrampazzo,

> Command-line parameter:
> 
> ```
>   --nrunner-max-parallel-tasks NRUNNER.MAX_PARALLEL_TASKS
>                         Number of maximum number tasks running in parallel. You can disable parallel execution by
>                         setting this to 1. Defaults to the amount of CPUs on this machine.
> ```

Pointer more than welcome. However, then I guess my perception here was wrong since I was implicitly associating tasks with tests according to the implementations I saw previously. I guess this simply reuses the name "task" for a different purpose? Or perhaps we have a handler that could split N tests into M tasks where N!=M?

willianrampazzo:
> Hi @willianrampazzo,
> 
> > Command-line parameter:
> > ```
> >   --nrunner-max-parallel-tasks NRUNNER.MAX_PARALLEL_TASKS
> >                         Number of maximum number tasks running in parallel. You can disable parallel execution by
> >                         setting this to 1. Defaults to the amount of CPUs on this machine.
> > ```
> 
> Pointer more than welcome. However, then I guess my perception here was wrong since I was implicitly associating tasks with tests according to the implementations I saw previously. I guess this simply reuses the name "task" for a different purpose? Or perhaps we have a handler that could split N tests into M tasks where N!=M?

If you look at BP003 (https://avocado-framework.readthedocs.io/en/82.0/blueprints/BP003.html) about the Task state machine, you will notice that it uses as an example one reference resolution.

Looking at the current implementation of a Task (https://github.com/avocado-framework/avocado/blob/master/avocado/core/nrunner.py#L617), it "wraps the execution of a runnable". The only use we have right now for a Task is to run a runnable from a reference resolution, but this is subject to change soon when we start to create runnable for the requirements and wrap them into Tasks as we discussed in the Requirements Resolver PR.

In a nutshell, a Task is a much broader entity than a simple test representation.

richtja:
I am closing this, because our priority is to focus on nrunner where this problem doesn't exist. There is still problem of compatibility of nrunner and avocado-vt and @clebergnu started working on that at [avocado-vt  side](https://github.com/avocado-framework/avocado-vt/pull/2904)

Issue #4290, opened 2020-11-05
controlpath file is not removing if reboot is happen in middle of the test 


smruti77:
we already added in the test to exit the session after completing the test. if test is running fine we are not getting ERROR: Failed connecting issue. but sometime in between test any interruption happen or in middle of the test any reboot is happening than the controlpath file is not removing ,we are again facing ERROR: Failed connecting issue.

beraldoleal:
Can you please provide more information ? What controlpath are you talking about? Can you describe a "how to reproduce" here?

smruti77:
i am talking about ssh controlpath what is present in this directory, /root/.ssh 
ex, avocado-master-root@121.1.1.88:22

abdhaleegit:
Hey @beraldoleal this is a similar one as discussed https://github.com/avocado-framework/avocado/issues/3810#issuecomment-645370367 

We have a CI automation framework where in the same tests were run on multiple different kexec kernels and things are fine till here as each test is calling host.remote_session.quit() to cleanup master control path job and files ( a good path test)

At times we have seen a kernel crash happens when testing were running due to real linux issues and kernel panics and system boots and the test which was running did not call host.remote_session.quit() to clean up master control files so  next time the test again try to use the old control path file and fails to connect to peer

let me elaborate here with actual recreation steps

I have my lpar booted to base kernel 4.18.0-147.el8.ppc64le and there was no master control started 

```
# ls .ssh/
id_rsa  id_rsa.pub  known_hosts
#  ssh -o 'StrictHostKeyChecking=no' -o 'UpdateHostKeys=no' -o 'ControlPath=~/.ssh/avocado-master-%r@%h:%p' -l root -p 22 -O check 121.1.1.88
Control socket connect(/root/.ssh/avocado-master-root@121.1.1.88:22): No such file or directory
```
Now start the ssh connection and create a control path file 

```
 # ssh -o 'ControlMaster=yes' -o 'ControlPersist=yes' -o 'StrictHostKeyChecking=no' -o 'UpdateHostKeys=no' -o 'ControlPath=~/.ssh/avocado-master-%r@%h:%p' -l root -p 22 121.1.1.88
root@121.1.1.88's password: 
Activate the web console with: systemctl enable --now cockpit.socket

This system is not registered to Red Hat Insights. See https://cloud.redhat.com/
To register this system, run: insights-client --register

Last login: Mon Nov  9 05:28:23 2020 from 121.1.1.77
 # exit
logout
[...]

PR #4326, opened 2020-11-25
added check_connection() to check ssh connection if failed reconnect it.

smruti77:
check_connection() will check for ssh connection if failed it will exit the
previous connection again also its failed it will check for control path file
and remove it,than do the ssh connection again.

purpose of the function: sometimes during test if kernel crash happen,control path is
not removing if we will use quit() also,and getting "Connection refused".for that we
need to remove the old control path file from .ssh directory.

### DIFF
--- a/avocado/utils/ssh.py
+++ b/avocado/utils/ssh.py
@@ -11,6 +11,7 @@ Example of use:
         if result.exit_status == 0:
             print(result.stdout_text)
 """
+import glob
 import os
 import shlex
 import stat
@@ -248,3 +249,23 @@ class Session:
             return result.exit_status == 0
         except process.CmdError as exc:
             raise NWException("failed to copy file {}".format(exc))
+
+    def remove_control_path_reconnect(self):
+        """
+        Establishes connection to the remote host if failed try to
+        close the session.if it failed remove control path file and
+        reconnect it.
+
+        :returns: True if success or else raise exception
+        :rtype: bool
+        """
+        if not self.connect():
+            if not self.quit():
+                full_path = os.path.expanduser('~/.ssh')
+                file_list = glob.glob("{}/avocado-master-root*".format(full_path))
+                for path in file_list:
+                    if self.host in path:
+                        os.remove(path)
+            if not self.connect():
+                raise NWException("failed to connect to {}".format(self.host))
+        return True


smruti77:
issue: #4290

avocado/utils/ssh.py
```diff
@@ -248,3 +248,28 @@ def copy_files(self, source, destination, recursive=False):
             return result.exit_status == 0
         except process.CmdError as exc:
             raise NWException("failed to copy file {}".format(exc))
+
+    def check_connection(self):
```
beraldoleal, line 252: the method says "check_connection" but you are doing more than that. The method name should be renamed to reflect the code. Something like "remove_control_path()"

avocado/utils/ssh.py
```diff
@@ -248,3 +248,28 @@ def copy_files(self, source, destination, recursive=False):
...
+        """
+        Check for ssh connection if failed remove control path file
+        and reconnect it.
+
+        :returns: True if success or else False
+        :rtype: bool
+        """
+        if not self.connect():
+            if not self.quit():
+                for root, dirct, files in os.walk("/root/.ssh"):
```
beraldoleal, line 262: We should make this not depending on root user. ~/.ssh for instance.

avocado/utils/ssh.py
```diff
@@ -248,3 +248,28 @@ def copy_files(self, source, destination, recursive=False):
...
+                    for file in files:
+                        if file.startswith("avocado-master-root"):
+                            path = os.path.join(root, file)
+                            os.remove(path)
```
beraldoleal, line 266: Another way to remove this is:

```python
file_list = glob.glob('~/.ssh/*avocado-master-root*')
for path in file_list:
   os.remove(path)
```
didn't test the code, just for have an idea.

avocado/utils/ssh.py
```diff
@@ -248,3 +248,28 @@ def copy_files(self, source, destination, recursive=False):
...
+                if self.connect():
+                    return True
+                return False
+            else:
+                if self.connect():
+                    return True
+                return False
+        else:
+            return True
```
beraldoleal, line 275: This method is also returning a lot of true and false, with different situations. We could raise exceptions with explanations instead of true|false.

beraldoleal (changes requested):
Hi @smruti77 thanks for this. I'm ok with this solution since is not aggressive and we will still let users decide to remove it or not.

I just left a few comments for you.

avocado/utils/ssh.py
narasimhan-v, line 275:
```
if self.connect():
    return True
rest of the code
```

avocado/utils/ssh.py
smruti77: glob.glob('~/.ssh/*avocado-master-root*') with ~/.ssh this not getting any output, so used /root/.ssh.

avocado/utils/ssh.py
smruti77: try to use ~/.ssh this but won't get file path.

avocado/utils/ssh.py
```diff
@@ -248,3 +249,20 @@ def copy_files(self, source, destination, recursive=False):
...
+        """
+        Check for ssh connection if failed remove control path file
+        and reconnect it.
+
+        :returns: True if success or else raise exception
+        :rtype: bool
+        """
+        if not self.connect():
+            if not self.quit():
+                file_list = glob.glob("/root/.ssh/avocado-master-root*")
```
beraldoleal, line 263: You are assuming that the user is root. This is not going to be accepted. We need to make this dynamic. You could use `$HOME` env variable or even better: `os.path.expanduser()` method.

avocado/utils/ssh.py
```diff
@@ -248,3 +249,20 @@ def copy_files(self, source, destination, recursive=False):
...
+
+    def remove_control_path(self):
+        """
+        Check for ssh connection if failed remove control path file
+        and reconnect it.
+
+        :returns: True if success or else raise exception
+        :rtype: bool
+        """
+        if not self.connect():
```
beraldoleal, line 261: I'm trying to understand why you need to establish a connection to the remote host, in order to remove a local file.

self.connect() is not only checking for connection but is trying to connect to the  host. And this is not what your method or docstring is saying.

beraldoleal (changes requested):

avocado/utils/ssh.py
smruti77: actually this function is for, if check connection failed than we will got for quit() to close session if that also not work than we will remove the control path file.if you want me to do this method for only removing control path file thats also fine.

avocado/utils/ssh.py
smruti77: ok,i will change it.

avocado/utils/ssh.py
beraldoleal: @smruti77 I'm just trying to understand what you are trying to do here. When you say on docstring " Check for ssh connection if failed ..." I was reading "Check for ssh connection" (just a check), but you are actually trying to connect. 

Could you please try to improve the docstring (can be here on a comment) with a very detailed docstring of what are you trying to achieve here? Maybe the sentence is not correct and I'm getting confused.

smruti77: ok

avocado/utils/ssh.py
```diff
@@ -248,3 +249,22 @@ def copy_files(self, source, destination, recursive=False):
             return result.exit_status == 0
         except process.CmdError as exc:
             raise NWException("failed to copy file {}".format(exc))
+
+    def remove_control_path_reconnect(self):
+        """
+        Establishes connection to the remote host if failed try to
+        close the session.if it failed remove control path file and
+        reconnect it.
```
beraldoleal, line 257: Now that I'm reading this docstring, this function sounds really strange to me. Let me explain:

Users of this function will try to connect to a host and if not possible you are removing ***all*** control path, affecting other connections, for instance. 

If you would like to remove the ***specific*** control path for that connection that failed, IMO makes sense to have this inside the "connect()" function.

beraldoleal (changes requested):

avocado/utils/ssh.py
```diff
@@ -248,3 +249,23 @@ def copy_files(self, source, destination, recursive=False):
...
+        :returns: True if success or else raise exception
+        :rtype: bool
+        """
+        if not self.connect():
+            if not self.quit():
+                full_path = os.path.expanduser('~/.ssh')
+                file_list = glob.glob("{}/avocado-master-root*".format(full_path))
+                for path in file_list:
+                    if self.host in path:
+                        os.remove(path)
```
beraldoleal, line 268: Same comment as before: you are trying to remove others' control files.

Imagine that we have multiple tests, with multiple host connections, and because of one connection you are removing all control files.

beraldoleal (changes requested):

avocado/utils/ssh.py
smruti77: @beraldoleal now i am removing one control path file,based on ip address.

beraldoleal: Ohh I see now, sorry for not noticing before.

But still, this is a little strange to me. Let me take a proper look here, and see what we can do.

beraldoleal: Ok, @smruti77, I did a closer look here to see what we can do. 

Your PR is assuming here that all connections are made to remote users 'root'. When, in fact, a different user can be used. Also, you are removing all files that have the same host, regardless of the port. IMO, this is a little over needed.

The reconnection portion is a specific use case, and different users can have different needs.

I have created a new PR adding one new method and one new property.  Users could run the cleanup of control files if needed. Please take a look at https://github.com/avocado-framework/avocado/pull/4350 and let me know if this will work for you.

smruti77:
this one handle in #4350

PR #4331, opened 2020-11-26
Fix broken -f option on selftests/check.py

willianrampazzo:
This fixes the following:

```
$ ./check.py -f
  File "/home/wrampazz/src/avocado/avocado.quick-dev/selftests/./check.py", line 518, in <module>
    sys.exit(main())
  File "/home/wrampazz/src/avocado/avocado.quick-dev/selftests/./check.py", line 500, in main
    for variants in suite.config['run.dict_variants']:
KeyError: 'run.dict_variants'
```

### DIFF
--- a/selftests/check.py
+++ b/selftests/check.py
@@ -496,8 +496,9 @@ def main():
     if args.features:
         features = []
         for suite in suites:
-            for variants in suite.config['run.dict_variants']:
-                features.append(variants['namespace'])
+            if suite.config.get('run.dict_variants') is not None:
+                for variants in suite.config['run.dict_variants']:
+                    features.append(variants['namespace'])
 
         unique_features = sorted(set(features))
         print('Features covered (%i):' %len(unique_features))


selftests/check.py
```diff
@@ -496,8 +496,9 @@ def main():
     if args.features:
         features = []
         for suite in suites:
-            for variants in suite.config['run.dict_variants']:
-                features.append(variants['namespace'])
+            if suite.config.get('run.dict_variants') is not None:
+                for variants in suite.config['run.dict_variants']:
+                    features.append(variants['namespace'])
```
beraldoleal, line 501: We are assuming that all configs are there, because there was registered with a default value. So, if users are not setting a new value, a default value should be there.

I tried to avoid having checks like this in the code. Maybe I'm missing something, but this brings me a "yellow flag" that we should investigate why 'run.dict_variants' is not being populated with the default value.

Not sure if makes sense.

willianrampazzo: That absolutely makes sense. Thanks for the review, @beraldoleal.

willianrampazzo:
Closing in favor of #4332.

PR #4332, opened 2020-11-26
Register run.dict_variants missing option from dict_variants plugin.

willianrampazzo:
This fixes the following:

```
$ ./check.py -f
  File "/home/wrampazz/src/avocado/avocado.quick-dev/selftests/./check.py", line 518, in <module>
    sys.exit(main())
  File "/home/wrampazz/src/avocado/avocado.quick-dev/selftests/./check.py", line 500, in main
    for variants in suite.config['run.dict_variants']:
KeyError: 'run.dict_variants'
```

### DIFF
--- a/avocado/plugins/dict_variants.py
+++ b/avocado/plugins/dict_variants.py
@@ -13,9 +13,23 @@
 # Authors: Cleber Rosa <crosa@redhat.com>
 
 from avocado.core import varianter
-from avocado.core.plugin_interfaces import Varianter
+from avocado.core.plugin_interfaces import Init, Varianter
+from avocado.core.settings import settings
 from avocado.core.tree import TreeNode
 
+
+class DictVariantsInit(Init):
+
+    name = 'dict_variants'
+    description = "Python Dictionary based varianter"
+
+    def initialize(self):
+        help_msg = 'Load the Variants from Python dictionaries'
+        settings.register_option(section='run',
+                                 key='dict_variants',
+                                 default=[],
+                                 key_type=list,
+                                 help_msg=help_msg)
 
 class DictVariants(Varianter):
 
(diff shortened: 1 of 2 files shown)


avocado/plugins/dict_variants.py
```diff
@@ -13,10 +13,23 @@
...
+class DictVariantsInit(Init):
+
+    name = 'dict_variants'
+    description = "Python Dictionary based varianter"
+
+    def initialize(self):
+        help_msg = 'Load the Variants from Python dictionaries'
+        settings.register_option(section='run',
+                                 key='dict_variants',
+                                 default=[],
```
beraldoleal, line 30: I don't remember to have implemented the "key_type" getting from the default value. Not sure. Just double-checking here. Maybe we need to add "key_type" too.

willianrampazzo: OKay, this code was mostly copy/paste from https://github.com/avocado-framework/avocado/blob/master/avocado/plugins/json_variants.py#L33. It fixed the problem I was having, but if `key_type` is needed, let me know and I can push a fix for both files.

beraldoleal: Yes, I took a look at the code and default is 'str' and there is no magic, as far I could notice. So yes, we need to pass `key_type=list`.

beraldoleal (changes requested):
Just one small comment about `key_type`.

avocado/plugins/dict_variants.py
willianrampazzo: Thanks for the update. I pushed a new commit.

beraldoleal (approved):
Thanks @willianrampazzo , this LGTM.

richtja (approved):

PR #4333, opened 2020-11-27
bp000: option-arguments should not be optional

### LINKED ISSUE #3441: BP001: Make the option-arguments not optional
Related #3427

beraldoleal:
This is a cosmetic change but help us having a better command-line
syntax as discussed on BP000. Basically this will make argparse raise
errors when option-arguments have no arguments and also fixes the syntax
displayed on help. Fixes #3441.

AFAICT, the only remaining optional option-arguments is 'refeferences' for both 'run'  and 'list' commands. Those are kept there because makes sense.

### DIFF
--- a/avocado/core/loader.py
+++ b/avocado/core/loader.py
@@ -407,7 +407,7 @@ def add_loader_options(parser, section='run'):
                 "settings->plugins.loaders.")
     settings.register_option(section=section,
                              key='loaders',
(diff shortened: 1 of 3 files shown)


willianrampazzo (approved):

clebergnu (approved):
LGTM, thanks!

PR #4334, opened 2020-11-30
Fix: Remove 16G hugepage only for Hash MMU

harish-24:
Currently a PowerVM machine has the capability to run on both Hash and Radix MMU, hence remove 16G hugepages only in case of Radix.

### DIFF
--- a/avocado/utils/memory.py
+++ b/avocado/utils/memory.py
@@ -258,7 +258,8 @@ def get_supported_huge_pages_size():
     # the <size> will always start from index 10.
     output = [int(each[10:].rstrip('kB')) for each in output]
     if os.uname()[4] in ['ppc64', 'ppc64le'] and b'PowerVM'\
-            in process.system_output("pseries_platform", ignore_status=True):
+            in process.system_output("pseries_platform", ignore_status=True)\
+            and 'MMU\t\t: Hash' in genio.read_file('/proc/cpuinfo').rstrip('\t\r\n\0'):
         output.remove(16777216)
     return output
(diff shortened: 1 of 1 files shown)


avocado/utils/memory.py
```diff
@@ -258,7 +258,8 @@ def get_supported_huge_pages_size():
     # the <size> will always start from index 10.
     output = [int(each[10:].rstrip('kB')) for each in output]
     if os.uname()[4] in ['ppc64', 'ppc64le'] and b'PowerVM'\
-            in process.system_output("pseries_platform", ignore_status=True):
+            in process.system_output("pseries_platform", ignore_status=True)\
+            and 'Hash' in genio.read_file('/proc/cpuinfo').rstrip('\t\r\n\0'):
```
willianrampazzo, line 262: I'm wondering if this statement isn't too generic. It is looking for `Hash` on the whole file. What happens if we have a new processor with `Hash` in the name, and it also supports huge pages?

harish-24: @willianrampazzo makes sense, added a more reliable check here. Thanks.

willianrampazzo (approved):
LGTM, thanks!

harish-24:
@willianrampazzo Please help in merging this, Thanks.

willianrampazzo:
Hi @harish-24, I know it is not mandatory for the Avocado Project, but we have been trying to wait for at least two approvals before merging a PR.

PraveenPenguin (approved):

PR #4350, opened 2020-12-08
utils.ssh: adding a cleanup master method and a property

beraldoleal:
The first will remove the control master file, if exists. The second
will return the same file, also only in cases that the file exists.

### DIFF
--- a/avocado/utils/ssh.py
+++ b/avocado/utils/ssh.py
@@ -127,6 +127,11 @@ class Session:
     def _check(self):
         return self._master_command('check')
 
+    def cleanup_master(self):
+        """Removes master file if exists."""
+        if self.control_master:
+            os.unlink(self.control_master)
+
     def connect(self):
         """
         Establishes the connection to the remote endpoint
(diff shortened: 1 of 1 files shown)


PraveenPenguin (approved):
LGTM

avocado/utils/ssh.py
```diff
@@ -168,6 +173,14 @@ def connect(self):
             self._connection = master
         return self._check()
 
+    @property
+    def control_master(self):
+        control = "~/.ssh/avocado-master-{}@{}:{}".format(self.user,
+                                                          self.host,
+                                                          self.port)
+        if os.path.exists(control):
```
smruti77, line 181: @beraldoleal this approach is fine, but code is not working properly. as control is "~/.ssh/avocado-master-root@121.1.1.88:22"  so  both os.path.exists() and os.unlink() will fail. so its can be like this,
```
    def control_master(self):
        control = "~/.ssh/avocado-master-{}@{}:{}".format(self.user,
                                                          self.host,
                                                          self.port)
        control_path = os.path.expanduser(control)
        if os.path.exists(control_path):
            return control_path
```

smruti77 (changes requested):

avocado/utils/ssh.py
beraldoleal: You are right @smitterl , thanks for this catch. I will fix on a force push.

smruti77: now its fine.

smruti77 (approved):

willianrampazzo (approved):

PR #4358, opened 2020-12-10
Kvm unit tests: add skip exitcode

zhijianli88:


willianrampazzo (approved):

contrib/testsuites/run-kvm-unit-test.sh
```diff
@@ -55,6 +55,23 @@ while [ true ]; do
...
+setup_skip_exitcode()
+{
+	mkdir -p $(dirname $CONF_FILE)
+	[ -f $CONF_FILE ] && cp "$CONF_FILE" "$CONF_FILE".kvm-unit-tests
+	trap restore_config EXIT
+
+	echo "[runner.exectest.exitcodes]" >>$CONF_FILE
+	echo "skip = [2, 77]" >>$CONF_FILE
+}
+
```
beraldoleal, line 74: Hi, @zhijianli88 , thanks for this. 

Just one question: Any reason for not using the JobAPI (python) here?

zhijianli88: It sounds good, i have no much idea with JobAPI now. is below is a JobAPI example ?
```python
 14 class NRunnerFeatures(unittest.TestCase):                                       
 15     @skipUnlessPathExists('/bin/false')                                         
 16     def test_custom_exit_codes(self):                                           
 17         status_server = "127.0.0.1:%u" % find_free_port()                       
 18         config = {'run.references': ['/bin/false'],                             
 19                   'run.test_runner': 'nrunner',                                 
 20                   'runner.exectest.exitcodes.skip': [1],                        
 21                   'nrunner.status_server_listen': status_server,                
 22                   'nrunner.status_server_uri': status_server,                   
 23                   'run.keep_tmp': True}                                         
 24         with Job.from_config(job_config=config) as job:                         
 25             self.assertEqual(job.run(), 0)   
```
do you also mean rewrite the whole script in python ?

beraldoleal: Yes, this is a jobapi example. I don't mind having a bash script. So I'm ok with this PR. I'm just pointing another way, that feels more natural to me.

But I understand if that is not the time and we can migrate this in a near future.

Issue #4359, opened 2020-12-11
Avocado Status Server not compatible with python 3.6


richtja:
The [server](https://docs.python.org/3.6/library/asyncio-eventloop.html#asyncio.Server) python 3.6 has no function `serve_forever()` 

Because of this the avocado status server not working on python 3.6

problem in [server.py](https://github.com/avocado-framework/avocado/blob/master/avocado/core/status/server.py#L41)

beraldoleal:
@richtja I had the error message today. I'm still investigating but seems to me that the problem is with my code. I tried with master and it works.

Can you please, double-check if on master your case is working?

richtja:
@beraldoleal Thanks that you are looking at this. I still have a problem on master. I did some investigation and I think that the problem is that `asyncio.start_server()` creates [asyncio.Server](https://docs.python.org/3.6/library/asyncio-eventloop.html#asyncio.Server) which don't have method `serve_forever()` on python3.6

I am getting this error:

> Task exception was never retrieved
future: <Task finished coro=<StatusServer.serve_forever() done, defined at /home/jarichte/Avocado/avocado/avocado/core/status/server.py:38> exception=AttributeError("'Server' object has no attribute 'serve_forever'",)>
Traceback (most recent call last):
  File "/home/jarichte/Avocado/avocado/avocado/core/status/server.py", line 41, in serve_forever
    await self._server_task.serve_forever()
AttributeError: 'Server' object has no attribute 'serve_forever'
2020-12-14 15:37:24,304 base_events      L1285 ERROR| Task exception was never retrieved
future: <Task finished coro=<Runner._update_status() done, defined at /home/jarichte/Avocado/avocado/avocado/plugins/runner_nrunner.py:200> exception=AttributeError("'NoneType' object has no attribute 'upper'",)>
Traceback (most recent call last):
  File "/home/jarichte/Avocado/avocado/avocado/plugins/runner_nrunner.py", line 223, in _update_status
    test_state = {'status': last_task_status.get('result').upper()}
AttributeError: 'NoneType' object has no attribute 'upper'

beraldoleal:
Looks like is not an error on status server itself.

beraldoleal:
```
 AttributeError: 'NoneType' object has no attribute 'upper'
```

And because of that, for some reason, we have a problem with status_server.

richtja:
Actually I think is opposite, we have this error because of some problem in status_server. This happens when nrunner reads data from status server and it has none.

[...]

PR #4360, opened 2020-12-14
Status Server compatibility with python 3.6

richtja:
`asyncio` package is not compatible between python 3.6 and 3.7 this checks the
python version and solves problems with status server on python 3.6

Reference:#4359

### DIFF
--- a/avocado/core/status/server.py
+++ b/avocado/core/status/server.py
@@ -1,4 +1,5 @@
 import asyncio
+import sys
 
 from ..settings import settings
 
@@ -38,7 +39,10 @@ class StatusServer:
     async def serve_forever(self):
         if self._server_task is None:
             await self.create_server()
-        await self._server_task.serve_forever()
+
+        # TODO: Delete this after deprecation of python 3.6
+        if sys.version_info >= (3, 7):
+            await self._server_task.serve_forever()
 
     def close(self):
         self._server_task.close()


avocado/core/status/server.py
```diff
@@ -38,7 +39,10 @@ async def create_server(self):
     async def serve_forever(self):
         if self._server_task is None:
             await self.create_server()
-        await self._server_task.serve_forever()
+
+        # TODO: Delete this after deprecation of python 3.6
+        if sys.version_info >= (3, 7):
+            await self._server_task.serve_forever()
```
beraldoleal, line 45: This is strange... because we have python 3.6 on travis making our tests and also python3.6 docs shows a `serve_forever()` method:

https://docs.python.org/3.6/library/socketserver.html#socketserver.BaseServer.serve_forever

richtja: Unfortunately I didn't find in the documentation that the [asyncio.Server](https://docs.python.org/3.6/library/asyncio-eventloop.html#asyncio.Server) inherits from [socketserver.BaseServer](https://docs.python.org/3.6/library/socketserver.html#socketserver.BaseServer.serve_forever) but maybe I am missing something. I didn't find why Travis didn't catch that but maybe we have some problem in our testing.

richtja: We can make a draft from this and investigate the problem further.

willianrampazzo (approved):

beraldoleal (approved):

PR #4369, opened 2020-12-30
[Requirements resolver] Initial integration with the state machine and suite

willianrampazzo:
This is a minimum set of changes to introduce and integrate the Requirements Resolver with the state machine and the test suite. Although there are some caveats, the changes do not impact the current behavior if the requirements docstrings are not used.

The following test enables the code of this PR:

```python
from avocado import Test


class PassTest(Test):

    """
    Example test that passes.

    :avocado: tags=fast
    """

    def test(self):
        """
        """
```

Caveats:
- Although tasks with dependencies wait for them to finish, if some dependency fails, the test will still run due to lack of status after a task finishes. To solve this problem, after a task finishes, it needs to have the real finish state (success, fail), and the state machine needs to have access to it.
- Unsupported `type` will crash Avocado due to lack of exception handling on the state machine.

### DIFF
--- a/avocado/core/nrunner.py
+++ b/avocado/core/nrunner.py
@@ -15,6 +15,7 @@ import sys
 import tempfile
 import time
 import unittest
+from uuid import uuid4
 
 try:
     import pkg_resources
@@ -664,6 +665,7 @@ class Task:
     def __init__(self, identifier, runnable, status_uris=None,
                  known_runners=None):
         self.identifier = identifier
+        self.uid = str(uuid4())
         self.runnable = runnable
         self.status_services = []
         if status_uris is not None:
@@ -672,12 +674,15 @@ class Task:
         if known_runners is None:
             known_runners = {}
         self.known_runners = known_runners
+        self.prereqs = []
         self.spawn_handle = None
         self.output_dir = None
 
     def __repr__(self):
-        fmt = '<Task identifier="{}" runnable="{}" status_services="{}"'
-        return fmt.format(self.identifier, self.runnable, self.status_services)
+        fmt = ('<Task identifier="{}" runnable="{}" prereqs="{}"'
+               ' status_services="{}"')
+        return fmt.format(self.identifier, self.runnable, self.prereqs,
+                          self.status_services)
 
     def are_requirements_available(self, runners_registry=None):
         """Verifies if requirements needed to run this task are available.
--- a/avocado/core/nrunner.py
+++ b/avocado/core/nrunner.py
@@ -448,6 +448,25 @@ class ExecTestRunner(ExecRunner):
 RUNNERS_REGISTRY_PYTHON_CLASS['exec-test'] = ExecTestRunner
 
 
+class RequirementRunner(ExecRunner):
+    """
+    Runner for standalone executables treated as requirements
+
+    Runnable attributes usage is identical to :class:`ExecRunner`
+    """
+    def run(self):
+        for most_current_execution_state in super().run():
+            returncode = most_current_execution_state.get('returncode')
+            if returncode == 0:
+                most_current_execution_state['result'] = 'pass'
+            else:
+                most_current_execution_state['result'] = 'fail'
+            yield most_current_execution_state
+
+
+RUNNERS_REGISTRY_PYTHON_CLASS['requirement'] = RequirementRunner
+
+
 class PythonUnittestRunner(BaseRunner):
     """
     Runner for Python unittests
--- a/avocado/core/requirements/resolver.py
+++ b/avocado/core/requirements/resolver.py
@@ -1,0 +1,106 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: Red Hat Inc. 2020
+# Authors: Willian Rampazzo <willianr@redhat.com>
+
+"""
+Requirements Resolver module.
+"""
+
+from ..nrunner import RUNNERS_REGISTRY_PYTHON_CLASS, Runnable, Task
+
+SUPPORTED_REQUIREMENTS = {}
+
+
+def create_requirements_tasks(tasks):
+    """Entry point for the Requirements Resolver.
+
+    This function isolates the Requirements Resolver object into this module.
+
+    :param tasks: List of tasks
+    :type tasks: list
+    """
+    resolver = RequirementsResolver(tasks)
+    return resolver.create_requirements_tasks()
+
+
+class UnsupportedRequirementType(Exception):
+    pass
+
+
+class RequirementsResolver:
+    """Requirements Resolver class."""
+
+    def __init__(self, tasks):
+        """
+        :param tasks: List of tasks to be resolved
+        :type tasks: list
+        """
+        self._tasks = tasks
+
+    def create_requirements_tasks(self):
+        """Create the list of tasks based on the test requirements."""
+        requirements_tasks = []
+        for task in self._tasks:
+            if not task.runnable.requirements:
+                continue
+
+            uris = [status_service.uri
+                    for status_service in task.status_services]
+
+            for requirement in task.runnable.requirements:
+                try:
+                    type_klass = SUPPORTED_REQUIREMENTS[requirement['type']]
+                except KeyError:
+                    raise UnsupportedRequirementType
+
+                type_resolver = type_klass(requirement, uris)
+                # create the new task based on requirement information
+                requirement_task = type_resolver.task_from_requirement()
+                # add the new task to the requirements tasks list
+                requirements_tasks.append(requirement_task)
+                # update the dependency list of the father task
+                task.prereqs.append(requirement_task.uid)
+        # update the tasks list with the new requirements tasks
+        self._tasks.extend(requirements_tasks)
+        return self._tasks
+
+
+class BaseRequirement:
+    """Base interface for a requirement type."""
+
+    def __init__(self, requirement, uris):
+        self._requirement = requirement
+        self._uris = uris
+
+    def task_from_requirement(self):
+        raise NotImplementedError
+
+
+class PackageRequirement(BaseRequirement):
+    """Handle requirements of type `package`."""
+
+    def task_from_requirement(self):
+        """Create a new software-manager task from the requirement info."""
+        # create a software-manager runnable
+        runnable = Runnable('requirement', 'avocado-software-manager',
+                            'install',
+                            self._requirement['name'])
+        task_id = 'requirement-%s-%s' % (self._requirement['type'],
+                                         self._requirement['name'])
+        # create the new software-manager task for the requirement
+        task = Task(task_id, runnable,
+                    status_uris=self._uris,
+                    known_runners=RUNNERS_REGISTRY_PYTHON_CLASS)
+        return task
+
+SUPPORTED_REQUIREMENTS['package'] = PackageRequirement
--- a/avocado/core/suite.py
+++ b/avocado/core/suite.py
@@ -22,6 +22,7 @@ from .exceptions import (JobTestSuiteReferenceResolutionError,
 from .loader import (DiscoverMode, LoaderError, LoaderUnhandledReferenceError,
                      loader)
 from .parser import HintParser
+from .requirements.resolver import create_requirements_tasks
 from .resolver import resolve
 from .settings import settings
 from .tags import filter_test_tags
@@ -141,6 +142,7 @@ class TestSuite:
             raise TestSuiteError(details)
 
         tasks = resolutions_to_tasks(resolutions, config)
+        tasks = create_requirements_tasks(tasks)
 
         if name is None:
             name = str(uuid4())
--- a/avocado/core/task/statemachine.py
+++ b/avocado/core/task/statemachine.py
@@ -79,14 +79,23 @@ class Worker:
         except IndexError:
             return
 
-        requirements_ok = await self._spawner.check_task_requirements(runtime_task)
-        if requirements_ok:
-            async with self._state_machine.lock:
-                self._state_machine.ready.append(runtime_task)
-        else:
-            async with self._state_machine.lock:
-                self._state_machine.finished.append(runtime_task)
-                runtime_task.status = 'FAILED ON TRIAGE'
+        # right now, support triage based on requirements only
+        if runtime_task.task.prereqs:
+            finished_tasks_uid = [runtime_task.task.uid for runtime_task
+                                  in self._state_machine.finished]
+            for prereq in runtime_task.task.prereqs:
+                if prereq in finished_tasks_uid:
+                    runtime_task.task.prereqs.remove(prereq)
+            # check if the last task was removed
+            if runtime_task.task.prereqs:
+                async with self._state_machine.lock:
+                    self._state_machine.triaging.append(runtime_task)
+                    runtime_task.status = 'WAITING PRE-REQS'
+                await asyncio.sleep(0.1)
+                return
+
+        async with self._state_machine.lock:
+            self._state_machine.ready.append(runtime_task)
 
     async def start(self):
         """Reads from ready, moves into either: started or finished."""
--- a/python-avocado.spec
+++ b/python-avocado.spec
@@ -214,6 +214,7 @@ PATH=$HOME/.local/bin:$PATH LANG=en_US.UTF-8 AVOCADO_CHECK_LEVEL=0 %{__python3} 
 %{_bindir}/avocado-runner-noop
 %{_bindir}/avocado-runner-exec
 %{_bindir}/avocado-runner-exec-test
+%{_bindir}/avocado-runner-requirement
 %{_bindir}/avocado-runner-python-unittest
 %{_bindir}/avocado-runner-avocado-instrumented
 %{_bindir}/avocado-runner-tap
--- a/setup.py
+++ b/setup.py
@@ -67,6 +67,7 @@ if __name__ == '__main__':
                   'avocado-runner-noop = avocado.core.nrunner:main',
                   'avocado-runner-exec = avocado.core.nrunner:main',
                   'avocado-runner-exec-test = avocado.core.nrunner:main',
+                  'avocado-runner-requirement = avocado.core.nrunner:main',
                   'avocado-runner-python-unittest = avocado.core.nrunner:main',
                   'avocado-runner-avocado-instrumented = avocado.core.nrunner_avocado_instrumented:main',
                   'avocado-runner-tap = avocado.core.nrunner_tap:main',


willianrampazzo:
isort is failing but the failure is not related to this PR: (check-855/955) selftests/isort.sh:  FAIL (3.47 s)

willianrampazzo:
`isort` issues should be fixed on https://github.com/avocado-framework/avocado/pull/4370

avocado/core/requirements/resolver.py
```diff
@@ -0,0 +1,102 @@
...
+
+    :param tasks: List of tasks
+    :type tasks: list
+    """
+    resolver = RequirementsResolver(tasks)
+    return resolver.create_requirements_tasks()
+
+class UnsupportedRequirementType(Exception):
+    pass
+
```
richtja, line 37: Style hint, this file doesn't have two blank lines between classes.  [PEP 8](https://www.python.org/dev/peps/pep-0008/#blank-lines)

willianrampazzo: Opss, indeed, good catch!

willianrampazzo: I'm wondering why `inspekt-style.sh` did not get it.

richtja: That's true I will investigate that

avocado/core/task/statemachine.py
```diff
@@ -79,14 +79,23 @@ async def triage(self):
...
-            async with self._state_machine.lock:
-                self._state_machine.finished.append(runtime_task)
-                runtime_task.status = 'FAILED ON TRIAGE'
+        # right now, support triage based on requirements only
+        if runtime_task.task.prereqs:
+            finished_tasks_uid = [runtime_task.task.uid for runtime_task
+                                  in self._state_machine.finished]
+            for prereq in runtime_task.task.prereqs:
+                if prereq in finished_tasks_uid:
+                    runtime_task.task.prereqs.remove(prereq)
```
pevogam, line 88: This logic is mixing having a prerequisite with a fulfilled prerequisite by considering `prereqs` for both. Can we have two attributes for fulfilled and unfulfilled prerequisites so that we can track both? Alternatively can we have `prereqs` renamed to something less misleading like `blocked_by` or `blockages`?

avocado/core/task/statemachine.py
```diff
@@ -79,14 +79,23 @@ async def triage(self):
...
+            # check if the last task was removed
+            if runtime_task.task.prereqs:
+                async with self._state_machine.lock:
+                    self._state_machine.triaging.append(runtime_task)
+                    runtime_task.status = 'WAITING PRE-REQS'
+                await asyncio.sleep(0.1)
```
pevogam, line 94: Is there a way to add some timeout handling to all of this? Not just for if a dependency fails but in general this will block any tasks which could have a maximum allotted time.

avocado/core/requirements/resolver.py
```diff
@@ -0,0 +1,106 @@
...
+        for task in self._tasks:
+            if not task.runnable.requirements:
+                continue
+
+            uris = [status_service.uri
+                    for status_service in task.status_services]
+
+            for requirement in task.runnable.requirements:
+                try:
+                    type_klass = SUPPORTED_REQUIREMENTS[requirement['type']]
```
pevogam, line 62: No way this could just be `type_class` which is also not a reserved word and easier to read?

avocado/core/requirements/resolver.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def create_requirements_tasks(self):
+        """Create the list of tasks based on the test requirements."""
+        requirements_tasks = []
+        for task in self._tasks:
+            if not task.runnable.requirements:
+                continue
+
+            uris = [status_service.uri
+                    for status_service in task.status_services]
```
pevogam, line 58: A general question not related to the code: what are the status URI-s used for? For the status server?

avocado/core/requirements/resolver.py
```diff
@@ -0,0 +1,106 @@
...
+        :type tasks: list
+        """
+        self._tasks = tasks
+
+    def create_requirements_tasks(self):
+        """Create the list of tasks based on the test requirements."""
+        requirements_tasks = []
+        for task in self._tasks:
+            if not task.runnable.requirements:
+                continue
```
pevogam, line 55: Is there an easy way to unify the `task.runnable.requirements` and `task.prereqs` data structures? Of course they are different types and contain different information but the relate to the same thing more or less which could simplify some of the parsing and redundancy of attributes.

avocado/core/requirements/resolver.py
pevogam: Maybe this could also be
```python
uris = [s.uri for s in task.status_services]
```
to keep everything on a single line and thus with slightly smaller cognitive complexity. The `s` or alternatively `service` is still fully readable since it is fully localized.

avocado/core/requirements/resolver.py
```diff
@@ -0,0 +1,106 @@
...
+def create_requirements_tasks(tasks):
+    """Entry point for the Requirements Resolver.
+
+    This function isolates the Requirements Resolver object into this module.
+
+    :param tasks: List of tasks
+    :type tasks: list
+    """
+    resolver = RequirementsResolver(tasks)
+    return resolver.create_requirements_tasks()
```
pevogam, line 33: I am a bit confused about this approach. Why a separate entry point into a class with the same method name? Is there a reason to keep the class within the module? It doesn't seem like a private class from the notation. I see two simpler ways to do this, either import the class in the other modules like
```python
from resolver import RequirementsResolver
```
and then use it normally or importing a single `create_requirements_tasks` module-level function which could be this one here. From what I read about the class `RequirementsResolver` below it seems to be nothing but a class with one method returning an extended list of tasks which might as well simply be this method here. Is there some nonobvious reason for the entire `RequirementsResolver` class and the current increased complexity?

avocado/core/nrunner.py
```diff
@@ -653,12 +674,15 @@ def __init__(self, identifier, runnable, status_uris=None,
...
         self.spawn_handle = None
         self.output_dir = None
 
     def __repr__(self):
-        fmt = '<Task identifier="{}" runnable="{}" status_services="{}"'
-        return fmt.format(self.identifier, self.runnable, self.status_services)
+        fmt = ('<Task identifier="{}" runnable="{}" prereqs="{}"'
+               ' status_services="{}"')
+        return fmt.format(self.identifier, self.runnable, self.prereqs,
+                          self.status_services)
```
pevogam, line 685: Could this representation be an f-string? It seems more readable if we have dropped support for Python 3.5.

pevogam (commented):
Hi @willianrampazzo, I am back with some comments as I find what you are proposing here of great importance to more elaborate test environments. I hope you find them useful.

avocado/core/nrunner.py
```diff
@@ -448,6 +448,25 @@ def run(self):
 RUNNERS_REGISTRY_PYTHON_CLASS['exec-test'] = ExecTestRunner
 
 
+class RequirementRunner(ExecRunner):
+    """
+    Runner for standalone executables treated as requirements
```
beraldoleal, line 454: I'm trying to understand why we need a special runner just for this. Maybe you could explain better on this docstring?

avocado/core/nrunner.py
```diff
@@ -664,6 +665,7 @@ class Task:
     def __init__(self, identifier, runnable, status_uris=None,
                  known_runners=None):
         self.identifier = identifier
+        self.uid = str(uuid4())
```
beraldoleal, line 668: A unique `uuid4` is a good idea here. But this might be a little confusing since we have already an `identifier`.

I would recommend change the API here and rename the identifier with `name` or something.

avocado/core/nrunner.py
```diff
@@ -672,12 +674,15 @@ def __init__(self, identifier, runnable, status_uris=None,
         if known_runners is None:
             known_runners = {}
         self.known_runners = known_runners
+        self.prereqs = []
```
beraldoleal, line 677: One suggestion is to rename this to `requirements` to be more explicity.

avocado/core/requirements/resolver.py
beraldoleal, line 33: My suggestion is to move this to the class to something like this:

```python
requirements = RequirementsResolver.requirements_from_tasks(tasks)
```

avocado/core/requirements/resolver.py
```diff
@@ -0,0 +1,106 @@
...
+
+class RequirementsResolver:
+    """Requirements Resolver class."""
+
+    def __init__(self, tasks):
+        """
+        :param tasks: List of tasks to be resolved
+        :type tasks: list
+        """
+        self._tasks = tasks
```
beraldoleal, line 48: Any reason for making this a private attribute?

avocado/core/requirements/resolver.py
```diff
@@ -0,0 +1,106 @@
...
+                # add the new task to the requirements tasks list
+                requirements_tasks.append(requirement_task)
+                # update the dependency list of the father task
+                task.prereqs.append(requirement_task.uid)
+        # update the tasks list with the new requirements tasks
+        self._tasks.extend(requirements_tasks)
+        return self._tasks
+
+
+class BaseRequirement:
```
beraldoleal, line 78: Nitpick: Could you please reorder the classes on this file to alphabetical order?

avocado/core/requirements/resolver.py
```diff
@@ -0,0 +1,106 @@
...
+    """Base interface for a requirement type."""
+
+    def __init__(self, requirement, uris):
+        self._requirement = requirement
+        self._uris = uris
```
beraldoleal, line 83: again, any reason for making them private?

avocado/core/task/statemachine.py
```diff
@@ -79,14 +79,23 @@ async def triage(self):
...
-        requirements_ok = await self._spawner.check_task_requirements(runtime_task)
-        if requirements_ok:
-            async with self._state_machine.lock:
-                self._state_machine.ready.append(runtime_task)
-        else:
-            async with self._state_machine.lock:
-                self._state_machine.finished.append(runtime_task)
-                runtime_task.status = 'FAILED ON TRIAGE'
+        # right now, support triage based on requirements only
+        if runtime_task.task.prereqs:
```
beraldoleal, line 83: If `prereqs` could return always a list, this If would not be necessary.

avocado/core/suite.py
```diff
@@ -22,6 +22,7 @@
 from .loader import (DiscoverMode, LoaderError, LoaderUnhandledReferenceError,
                      loader)
 from .parser import HintParser
+from .requirements.resolver import create_requirements_tasks
 from .resolver import resolve
```
beraldoleal, line 26: We have a module named `resolver` and now another one named `requirements.resolver`. Hum.. just thinking out of loud. What about having a package `resolvers.` with all the resolvers there? 

Yes, maybe too much work for now. Just an idea.

avocado/core/suite.py
```diff
@@ -141,6 +142,7 @@ def _from_config_with_resolver(cls, config, name=None):
             raise TestSuiteError(details)
 
         tasks = resolutions_to_tasks(resolutions, config)
+        tasks = create_requirements_tasks(tasks)
```
beraldoleal, line 145: Are you going to remove the line above? Are you going to introduce the config to this new method?

Because AFAICT, we use config to filter by tags, for instance. So I'm not sure how this is preserved.

beraldoleal (commented):
Hi @willianrampazzo nice work here.

The code looks fine to me. Just left a few small comments.

avocado/core/requirements/resolver.py
```diff
@@ -0,0 +1,106 @@
...
+            uris = [status_service.uri
+                    for status_service in task.status_services]
+
+            for requirement in task.runnable.requirements:
+                try:
+                    type_klass = SUPPORTED_REQUIREMENTS[requirement['type']]
+                except KeyError:
+                    raise UnsupportedRequirementType
+
+                type_resolver = type_klass(requirement, uris)
```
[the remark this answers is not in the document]
willianrampazzo: Okay, I like `type_requirements`.

avocado/core/requirements/resolver.py
```diff
@@ -0,0 +1,106 @@
...
+                continue
+
+            uris = [status_service.uri
+                    for status_service in task.status_services]
+
+            for requirement in task.runnable.requirements:
+                try:
+                    type_klass = SUPPORTED_REQUIREMENTS[requirement['type']]
+                except KeyError:
+                    raise UnsupportedRequirementType
```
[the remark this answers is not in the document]
willianrampazzo: Agreed. The logging part still needs a good amount of work.

avocado/core/nrunner.py
willianrampazzo: There is a couple of reason that I don't think makes sense to add to the docstring.

The `ExecRunner` do not have the `most_current_execution_state` implementation. I thought about using it for future extension of the logging capabilities.

I started with an implementation that used the `ExecTestRunner` but reached some problems because of the `skip` status, which is not desired on a requirements resolution.

To avoid the problems described above, I thought it would be worth creating a new runner and extend it in the future based on the support for new requirements types.

avocado/core/nrunner.py
willianrampazzo: I agree with you, and my first intention was to change the API.

When I started evaluating the change, I noticed it would require a massive change in the code. This would impact the current command line syntax we have.

It is worth changing it, but on a separate PR, not related to the requirements resolver.

avocado/core/task/statemachine.py
pevogam: If the list is empty this will execute as if it is `None` so this is probably correct implementation.

beraldoleal: You are right, @pevogam, I forgot about line 84 here.

I would probably change `_state_machine.finished` to a dict (using the `task.uid` as keys), instead of a list, to avoid this dict comprehension and to do everything inside the loop.

But, probably this would need a much bigger refactoring. So yes, I'm ok with that.

avocado/core/nrunner.py
willianrampazzo: This is a list of pre-requirements that needs to be fulfilled before this task starts. Do you think calling it `requirements` will not confuse with the test requirements? Whatever is clearer is fine for me

avocado/core/nrunner.py
willianrampazzo: We've been using `format` all over the code; that is the only reason I'm still using it when I have to.

avocado/core/requirements/resolver.py
willianrampazzo: If it won't be used outside of this class, a best practice is to make it private. Anyway, I don't have a strong opinion here.

avocado/core/requirements/resolver.py
willianrampazzo: Absolutely!

avocado/core/requirements/resolver.py
willianrampazzo: Same as the above answer.

avocado/core/suite.py
willianrampazzo: I agree! I initially thought these 2 would be the only resolvers, but who knows :)

avocado/core/task/statemachine.py
willianrampazzo: That makes sense, and I think `prereqs` is already always a list. I'll check that.

avocado/core/nrunner.py
pevogam: I see, it is never too late to start a possible conversion but if somebody is actually planning for a one-time run over all code one day consistency here is fine.

avocado/core/nrunner.py
pevogam: Oh, so I got this even more confused here, I thought with `prereqs` you mean `prerequisites` which is pretty much the same as requirements. I would have never expected it means "pre-requirements" and I am not even sure that this means just from the wording. I suggested some other possible names in other comments here but it seems you run into recursion problem where task requirements have their own requirements where multiple names will only exacerbate the situation.

avocado/core/suite.py
pevogam: Are they doing the same things though? I am not as accustomed to the code as the two of you but AFAIR `resolver` was used just as a new word for test `loader` to differentiate from the old runner and the current resolver is a specialized class for fulfilling test requirements.

avocado/core/nrunner.py
willianrampazzo: Yes, naming has become harder and harder when `requirements` is used in multiple places to mean different things. I'll try my best to make it clearer in the next version.

avocado/core/nrunner.py
willianrampazzo: Indeed, I think a refactoring PR on f-string would be better.

avocado/core/requirements/resolver.py
willianrampazzo: > I am a bit confused about this approach. Why a separate entry point into a class with the same method name? Is there a reason to keep the class within the module? It doesn't seem like a private class from the notation. I see two simpler ways to do this, either import the class in the other modules like
> 
> ```python
> from resolver import RequirementsResolver
> ```
> 
> and then use it normally or importing a single `create_requirements_tasks` module-level function which could be this one here. 

I find this approach counter-intuitive and, depending on how it is implemented, may become more complex to maintain. Anyway, as you and @beraldoleal opposed to it, I'll change that.

> From what I read about the class `RequirementsResolver` below it seems to be nothing but a class with one method returning an extended list of tasks which might as well simply be this method here. Is there some nonobvious reason for the entire `RequirementsResolver` class and the current increased complexity?

Indeed, this is a simple dispatcher class, just like described in the BP002. This is meant to reduce the complexity in the future addition of new requirements types. So, my implementation here is thinking about extensibility in the future. Without a dispatcher and with multiple requirements types, I don't see how it could be simpler.

willianrampazzo: Losing vote. I'll change that.

avocado/core/requirements/resolver.py
willianrampazzo: Yes, I thought about it, but it will require a mass refactor. I'm planning to do that outside of the requirements resolver work, just like the `identifier` renaming I commented on above.

avocado/core/requirements/resolver.py
willianrampazzo: > A general question not related to the code: what are the status URI-s used for? For the status server?

Correct, In the future, it will be used as a single point for reporting status.

willianrampazzo: > Maybe this could also be
> 
> ```python
> uris = [s.uri for s in task.status_services]
> ```
> 
> to keep everything on a single line and thus with slightly smaller cognitive complexity. The `s` or alternatively `service` is still fully readable since it is fully localized.

Okay, I like that!

avocado/core/requirements/resolver.py
willianrampazzo: No problem on my side. I just noticed it is a common practice in many projects to use this kind of naming, and as we are already using it on our code, I thought it would not hurt.

avocado/core/requirements/resolver.py
```diff
@@ -0,0 +1,106 @@
...
+        # update the tasks list with the new requirements tasks
+        self._tasks.extend(requirements_tasks)
+        return self._tasks
+
+
+class BaseRequirement:
+    """Base interface for a requirement type."""
+
+    def __init__(self, requirement, uris):
+        self._requirement = requirement
```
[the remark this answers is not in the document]
willianrampazzo: I don't see a reason to split a known object in its `key:values` if we know how to access the attributes. I mean, we have the consistency of a `requirement` object all over the code. Remember that different types of requirements will have different attributes and will be handled in their own requirement class.

avocado/core/suite.py
willianrampazzo: Well, if we say that a resolver is "resolving" something from one representation to another, I would agree that our resolvers are doing the same.

avocado/core/suite.py
willianrampazzo: Maybe I should rename the `tasks` variables to something more meaningful. The thing here is, `create_requirements_tasks` needs the test tasks created to update the `prereqs` of each task and create the new requirements tasks based on the list of requirements of each test task.

While I was implementing, I thought the code would become more complex if I add the requirements resolution inside the `resolutions_to_tasks` function.

So, the code is basically extending the `tasks` list and updating the test tasks that already exist.

Would it make the code clear if I do:

```python
test_tasks = resolutions_to_tasks(resolutions, config)
all_tasks = create_requirements_tasks(test_tasks)
```

Or something like that?

avocado/core/task/statemachine.py
willianrampazzo: I don't see a reason for not having a list of fulfilled and pending requirements. I'll add that.

avocado/core/task/statemachine.py
willianrampazzo: I think we have a `timeout` by tests and by test suites. If a requirement is not fulfilled on time, it will reach the test timeout. I don't think we need another timeout specifically for the requirements right now. If we conclude it is needed in the future, we can add it.

avocado/core/nrunner.py
pevogam: In this case it might be worth noting it down for everyone in an issue as a TODO card that could be tracked globally.

avocado/core/nrunner.py
pevogam: Might also be worth noting it down as a TODO for the future but it's your call, just advising on keeping track of ideas in one place for everybody.

avocado/core/requirements/resolver.py
pevogam: Can you provide more details on how it will reduce the future complexity of new requirement types? Let's say we have `PackageRequirement` and `ImageRequirement` for instance, how would helping extra dispatcher class help here?

avocado/core/requirements/resolver.py
pevogam: @willianrampazzo Do you anticipate the tasks being accessed from the resolver? I guess it depends on what you imagine in terms of how the requirement resolver will be handled in the future but this seems like something internal as we only use the `resolver.create_requirements_tasks` once and don't really use any real class advantages of the resolver, a dispatcher class as you said above.

avocado/core/requirements/resolver.py
pevogam: Great, so I guess it makes sense as another TODO card then.

avocado/core/requirements/resolver.py
pevogam: Yes, I understand and I have seen this before in the current code base too so there is nothing wrong with your line whatsoever. However, nothing stops us from making a step against it now and completing it in the future and avoid creating raising extra questions to newcomer readers of the code.

pevogam: But as you wish, this is a minor thing that just sticks out a bit.

avocado/core/requirements/resolver.py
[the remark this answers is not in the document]
pevogam: Isn't this a single instance of a `Requirement` subclass?

avocado/core/requirements/resolver.py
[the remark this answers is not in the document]
pevogam: Yes, good point. Alright then

avocado/core/suite.py
pevogam: Right but I prefer not to attach to the word itself especially considering how ubiquitous its use has become thus making it lose a lot of its technical meaning to begin with. A far better approach IMO would be to later rename the "resolver" in the other meaning back to "loader" and restore the original loader/runner pairing. In this way both modules meanings are differentiated as they should be instead of going the opposite way and even putting them in the same subpackage as if they are variations or components of the same functional goal.

pevogam: But I might misunderstand, both way would work and worse case one would create the need of more refactoring much later on once their roles are better established.

avocado/core/suite.py
pevogam: I think your new variable names are a good idea, they make your intent a lot clearer. One could realize it still if they spend more time on the code but this is a cognitive speedup.

I am not sure about all the information the config contains but specifically to the filters the current implementation should be fine because the config will be fully used to filter our the `test_tasks` and then the requirements only for those will be parsed and added to the overall list.

avocado/core/task/statemachine.py
pevogam: :+1:

avocado/core/nrunner.py
willianrampazzo: Sure, as soon as this work is merged, I'll open the issue.

avocado/core/requirements/resolver.py
willianrampazzo: I don't see a reason to access the tasks from a resolver.

avocado/core/nrunner.py
willianrampazzo: Yep, will do it after this work is merged.

avocado/core/requirements/resolver.py
willianrampazzo: In BP002 (https://avocado-framework.readthedocs.io/en/latest/blueprints/BP002.html), we specified 6 types of requirements that we are planning to support: file, image, package, git repository, source code, operating system parameters. We already discussed the requirements resolver could evolve to support environment requirements, like a specific VM, a specific host type with a special configuration in a cloud farm.

I see two options for removing the dispatcher class:
- Move the code to the already existing entry point function `create_requirements_tasks`. It would work for me without problems, but you and @beraldoleal are not happy with that function.
- Move the code inside the `suite.py` file and remove the call to the `create_requirements_tasks`. In my opinion, this will mix the requirements resolver code with the suite code and make things harder to maintain. Imagine I need to change something in the way the requirements classes are called. I would expect to change a code inside the requirements resolver file, but actually, I would need to dig the call on the `suite.py` file.

So, besides these two options, what solution would you suggest to handle 6+ different types of requirements?

avocado/core/requirements/resolver.py
willianrampazzo: Sure, as soon as this work is merged, I'll open the issue.

avocado/core/requirements/resolver.py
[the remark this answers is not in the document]
willianrampazzo: ops, I mean `type_requirement`.

avocado/core/requirements/resolver.py
pevogam: I am still not sure how the current dispatcher class helps this extensibility but what I meant before was that I am fine with either keeping this in either directly calling the `create_requirements_tasks` methods from the resolver class (rather than having a duplicate module level function) or directly calling a single module function like this one without having an entire dispatcher class just to pass around the list. So I am fine with your point 1 above, what I was pointing out earlier is the middle solution where you have both the module and the class `create_requirements_tasks` while one of them seems to be enough.

willianrampazzo: Okay, got it. Thanks for clarifying!

willianrampazzo:
Closing in favor of #4423.

PR #4370, opened 2020-12-31
Fix isort 5 issues.

willianrampazzo:
Travis has moved to isort 5. Our isort check needs some changes to work correctly.

Added all the new errors reported to the --skip section to be evaluated later.

### DIFF
--- a/selftests/isort.sh
+++ b/selftests/isort.sh
@@ -1,7 +1,25 @@
 #!/bin/sh -e
 
-isort --recursive --check-only \
-      --skip selftests/.data/loader_instrumented/dont_crash.py \
-      --skip selftests/.data/loader_instrumented/double_import.py \
-      --skip selftests/.data/loader_instrumented/dont_detect_non_avocado.py \
-      --skip avocado/__init__.py
+PARENT=$(cd ..; pwd)
+
+isort $PARENT --check-only \
+              --skip $PARENT/selftests/run \
+              --skip $PARENT/selftests/unit/test_datadir.py \
+              --skip $PARENT/selftests/.data/loader_instrumented/double_import.py \
+              --skip $PARENT/selftests/.data/loader_instrumented/dont_detect_non_avocado.py \
+              --skip $PARENT/selftests/.data/loader_instrumented/imports.py \
+              --skip $PARENT/selftests/functional/test_basic.py \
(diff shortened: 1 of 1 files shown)


PraveenPenguin (approved):
LGTM

richtja (approved):
@willianrampazzo LGTM, but I think we should create an issue that will track the fix of the skipped tests.

selftests/isort.sh
```diff
@@ -1,7 +1,25 @@
...
 
-isort --recursive --check-only \
-      --skip selftests/.data/loader_instrumented/dont_crash.py \
-      --skip selftests/.data/loader_instrumented/double_import.py \
-      --skip selftests/.data/loader_instrumented/dont_detect_non_avocado.py \
-      --skip avocado/__init__.py
+PARENT=$(cd ..; pwd)
+
+isort $PARENT --check-only \
+              --skip $PARENT/selftests/run \
```
beraldoleal, line 6: Any reason for skipping those `selftests/run` and `unit/test_datadir.py`?

willianrampazzo: They are failing on `isort`, ut as I mentioned, I added all the fails for now without checking them:

```
ERROR: /home/wrampazz/src/avocado/avocado.quick-dev/selftests/run Imports are incorrectly sorted and/or formatted.
ERROR: /home/wrampazz/src/avocado/avocado.quick-dev/selftests/unit/test_datadir.py Imports are incorrectly sorted and/or formatted.
```

willianrampazzo:
> @willianrampazzo LGTM, but I think we should create an issue that will track the fix of the skipped tests.

I agree! If the PR gets merged, I'll open an issue.

selftests/isort.sh
beraldoleal: hum.. my only concern is that we forgot about them.

willianrampazzo: What do you mean? Forgot where?

beraldoleal: Since we are skipping those files we are moving the errors under the carpet. If those errors are a quick fix, then I prefer to fix then before this PR.

willianrampazzo: Okay, I'll check them and send the fixes here. I was in a rush last week and submitted it the way it is.

willianrampazzo:
Will send a v2 with the fixes for the errors.

PR #4371, opened 2021-01-04
Fix isort 5 issues [v2]

willianrampazzo:
Travis has moved to `isort` 5. Our `isort` check needs a small change to work correctly.

Also, fixed new failures.

Changes from v1:
- Fix new and old failures,

### DIFF
--- a/avocado/__init__.py
+++ b/avocado/__init__.py
@@ -25,7 +25,7 @@ __all__ = ['Test',
            'TestCancel']
 
 
-from avocado.core import register_core_options, initialize_plugins
+from avocado.core import initialize_plugins, register_core_options
 from avocado.core.settings import settings
 
 register_core_options()
--- a/avocado/core/main.py
+++ b/avocado/core/main.py
@@ -60,7 +60,7 @@ def handle_exception(*exc_info):
 
 def main():
     sys.excepthook = handle_exception
-    from avocado.core.app import AvocadoApp    # pylint: disable=E0611
+    from avocado.core.app import AvocadoApp  # pylint: disable=E0611
 
     # Override tmp in case it's not set in env
     for attr in ("TMP", "TEMP", "TMPDIR"):
--- a/avocado/plugins/exec_path.py
+++ b/avocado/plugins/exec_path.py
@@ -16,10 +16,9 @@ Libexec PATHs modifier
 
 import os
 
-from pkg_resources import resource_filename
-
 from avocado.core.output import LOG_UI
 from avocado.core.plugin_interfaces import CLICmd
+from pkg_resources import resource_filename
 
 
 class ExecPath(CLICmd):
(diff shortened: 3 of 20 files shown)


richtja (changes requested):
IIUIC this solution is compatible with isort>=5 but in `requirements-selftests.txt` we are using isort==4.3.21 and pylint==2.5.0 unfortunately pylint 2.5.0 requires isort<5,>=4.2.5 IMO this solution is incompatible with our self tests requirements.

willianrampazzo:
> IIUIC this solution is compatible with isort>=5 but in `requirements-selftests.txt` we are using isort==4.3.21 and pylint==2.5.0 unfortunately pylint 2.5.0 requires isort<5,>=4.2.5 IMO this solution is incompatible with our self tests requirements.

I did some testing and, indeed, the output from `isort` 4 and `isort` 5 are different. `isort` 5 fixes are incompatible with `isort` 4.

I'm fine if we stick to `isort` 4. I'm wondering why Travis installed `isort` 5.

beraldoleal:
> > IIUIC this solution is compatible with isort>=5 but in `requirements-selftests.txt` we are using isort==4.3.21 and pylint==2.5.0 unfortunately pylint 2.5.0 requires isort<5,>=4.2.5 IMO this solution is incompatible with our self tests requirements.
> 
> I did some testing and, indeed, the output from `isort` 4 and `isort` 5 are different. `isort` 5 fixes are incompatible with `isort` 4.
> 
> I'm fine if we stick to `isort` 4. I'm wondering why Travis installed `isort` 5.

If we can install easily isort 5 on major distros, we can also upgrade our requirements to install isort 5.

willianrampazzo:
> > > IIUIC this solution is compatible with isort>=5 but in `requirements-selftests.txt` we are using isort==4.3.21 and pylint==2.5.0 unfortunately pylint 2.5.0 requires isort<5,>=4.2.5 IMO this solution is incompatible with our self tests requirements.
> > 
> > 
> > I did some testing and, indeed, the output from `isort` 4 and `isort` 5 are different. `isort` 5 fixes are incompatible with `isort` 4.
> > I'm fine if we stick to `isort` 4. I'm wondering why Travis installed `isort` 5.
> 
> If we can install easily isort 5 on major distros, we can also upgrade our requirements to install isort 5.

Right, but @richtja mentioned `pylint` 2.5 depends on `isort` 4. We need to also evaluate if we can bump the `pylint` version, if there is a new version.

willianrampazzo:
Initially, I thought the `isort` problem of #4369 was related to the version of `isort`, as I had version 5 locally, and it was also failing due to command syntax. It turned out that #4369 had a failure on `isort` 4, but not on `isort` 5 from my machine. I fixed the code on #4369, and the checks are passing now. In this case, this PR is not needed at this moment.

The open question is still why this PR passed the `isort` check if it was supposed to run on `isort` 4, which should be incompatible with this PR syntax and fixes.

PR #4373, opened 2021-01-06
Inspekt-style fix

richtja:
Inspekt style stopped using working directory as default directory for style check. Because of this problem our style checks didn't work properly for some time. This PR fixing this problem and also fixing forgotten style issues in avocado.


willianrampazzo (approved):
Thanks for checking that!

beraldoleal (approved):
LGTM.

PR #4385, opened 2021-01-19
vmimage: fix fedora image provider for older images

Dovgalyuk:
Old (version<32) fedora images now are stored at different server.
This patch adds urls of that servers for base and secondary
fedora image providers.

Signed-off-by: Pavel Dovgalyuk <Pavel.Dovgalyuk@ispras.ru>

### DIFF
--- a/avocado/utils/vmimage.py
+++ b/avocado/utils/vmimage.py
@@ -199,13 +199,19 @@ class FedoraImageProviderBase(ImageProviderBase):
     """
 
     HTML_ENCODING = 'iso-8859-1'
+    url_old_images = None
 
     def get_image_url(self):
         if int(self.version) >= 28:
             cloud = 'Cloud'
         else:
             cloud = 'CloudImages'
-        self.url_images = self.url_images % cloud
+
+        if self.url_old_images and int(self.version) <= 31:
+            self.url_versions = self.url_old_images
+
+        self.url_images = self.url_versions + '{version}/' \
+            + cloud + '/{arch}/images/'
         return super(FedoraImageProviderBase, self).get_image_url()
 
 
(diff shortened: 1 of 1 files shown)


beraldoleal (approved):
Hi @Dovgalyuk , thanks for this.

LGTM.

beraldoleal (commented):
Hi @Dovgalyuk , thanks for this.

LGTM, but Travis is complaining about lint, and your signed off. Can you please double-check?

Dovgalyuk:
> Hi @Dovgalyuk , thanks for this.
> 
> LGTM, but Travis is complaining about lint, and your signed off. Can you please double-check?

I fixed some the code to allow testing.
Now I see only signed-off complain, but don't understand why it occurs. Maybe because I used different email than in github account.

beraldoleal:
@Dovgalyuk yes, the email mismatch it is a problem here, would be nice to use the same email. (You can add multiple emails to your GH account in your settings)

Dovgalyuk:
> @Dovgalyuk yes, the email mismatch it is a problem here, would be nice to use the same email. (You can add multiple emails to your GH account in your settings)

I did it, but after committing the last change.

Dovgalyuk:
> /packit copr-build

Seems like travis can't compare secondary emails.

beraldoleal:
@Dovgalyuk looks Travis is complaining because only one commits is signed. All your commits must be signed.

Also, we don't need to keep all the changes in history here. You can squash them, to not polute our git history.

Dovgalyuk:
> @Dovgalyuk looks Travis is complaining because only one commits is signed. All your commits must be signed.
> 
> Also, we don't need to keep all the changes in history here. You can squash them, to not polute our git history.

I thought that PRs are usually squashed by maintainers.
However, I did it but travis still complains about signed-off.

Dovgalyuk:
> @Dovgalyuk looks Travis is complaining because only one commits is signed. All your commits must be signed.
> 
> Also, we don't need to keep all the changes in history here. You can squash them, to not polute our git history.

Trying one more time, after fixing letter case in email.

clebergnu (approved):
The only CI failure was about the SOB check, so this LGTM. Thanks!

PR #4393, opened 2021-01-27
signedoff-checks.sh: don't take letter case in account

Dovgalyuk:
Added -i option to grep to allow matching names and emails
when letter cases in them are different.

Signed-off-by: Pavel Dovgalyuk <Pavel.Dovgalyuk@ispras.ru>


clebergnu (approved):
LGTM!

PR #4405, opened 2021-02-06
Enable Python3 checker messages missing for some built-in functions

tiagohonorato:
With the introduction of Python3, some built-in functions changed its return to an iterator type. This PR, related to issue #3466, does the modifications needed to enable checks for `dict.keys, dict.items, filter, range, zip` to warn the use outside an iterator context. In some cases the check had to be disabled as it might be a false positive.


richtja (approved):
LGTM, thanks

willianrampazzo (approved):

avocado/plugins/distro.py
```diff
@@ -353,7 +353,7 @@ def configure(self, parser):
                                  parser=parser,
                                  long_arg='--distro-def-path')
 
-        type_choices = DISTRO_PKG_INFO_LOADERS.keys()
+        type_choices = tuple(DISTRO_PKG_INFO_LOADERS.keys())
```
beraldoleal, line 356: Any reason for using `tuple` instead of `list` here?

tiagohonorato: It was just to follow the docs for `settings.register_option()`, but `list` should work fine as well from what I can remember.

clebergnu (approved):
LGTM, thanks!

PR #4411, opened 2021-02-10
Enable Python3 checker messages for dict.iter*()

tiagohonorato:
Dict methods iteritems(), iterkeys() and itervalues() are longer supported
in Python 3. The check was disabled for some lines as AvocadoParam,
AvocadoParams and ValueDict classes define a method called iteritems()
that triggers the warning.

Related to #3466


richtja (changes requested):
Hi @tiagohonorato LGTM, just can you please resolve the conflicts which formed after the merge of your last PR #4405

tiagohonorato:
Hi @richtja, thanks for pointing that out. I will resolve the conflicts.

richtja (approved):
LGTM, thanks

willianrampazzo (commented):
This PR fixes 3 places and adds local disables in 8 places. I'm wondering if it is worth enabling the check globally without fixing most of the places we have problems with now.

tiagohonorato:
Hi @willianrampazzo, the check was disabled in some places because I see those as false positives. To issue the warning, Pylint seems to just check whether methods named ``iterkeys()``, ``itervalues()`` and ``iteritems()`` are called, not taking the instance in consideration, in other words, if it is a ``dict`` (I will take a closer look at Pylint's codebase and open an issue there if it is confirmed). Consequently, if we define a method named ``iterkeys()``, ``itervalues()`` or ``iteritems()`` it will trigger the error when called, even though we are not referring to an instance of ``dict``, but to, for example, an instance of ``AvocadoParam``.

On the other hand, I see your point and it might be useful to change these names that conflict with ``dict.iter*``methods to avoid confusion when the check is enabled, specially for new code that uses ``AvocadoParam`` and ``AvocadoParams``. Maybe change these methods to something like ``iter_items()``, what do you think?

optional_plugins/varianter_yaml_to_mux/avocado_varianter_yaml_to_mux/mux.py
```diff
@@ -294,7 +294,7 @@ def iteritems(self):
 
     def items(self):
         """ Slower implementation with the use of __getitem__ """
-        return self.iteritems()
+        return self.iteritems()  # pylint: disable=W1620
```
tiagohonorato, line 297: By the way, after reviewing the modifications after @willianrampazzo's comment, we might me able to remove this method entirety, as ``ValueDict`` inherits from ``dict`` and the return from ``dict.items`` is already an iterator.

richtja:
> Hi @willianrampazzo, the check was disabled in some places because I see those as false positives. To issue the warning, Pylint seems to just check whether methods named `iterkeys()`, `itervalues()` and `iteritems()` are called, not taking the instance in consideration, in other words, if it is a `dict` (I will take a closer look at Pylint's codebase and open an issue there if it is confirmed). Consequently, if we define a method named `iterkeys()`, `itervalues()` or `iteritems()` it will trigger the error when called, even though we are not referring to an instance of `dict`, but to, for example, an instance of `AvocadoParam`.
> 
> On the other hand, I see your point and it might be useful to change these names that conflict with `dict.iter*`methods to avoid confusion when the check is enabled, specially for new code that uses `AvocadoParam` and `AvocadoParams`. Maybe change these methods to something like `iter_items()`, what do you think?

@willianrampazzo, @tiagohonorato about this problem I already opened discussion #4414, where we can find out the best approach.

willianrampazzo:
> > Hi @willianrampazzo, the check was disabled in some places because I see those as false positives. To issue the warning, Pylint seems to just check whether methods named `iterkeys()`, `itervalues()` and `iteritems()` are called, not taking the instance in consideration, in other words, if it is a `dict` (I will take a closer look at Pylint's codebase and open an issue there if it is confirmed). Consequently, if we define a method named `iterkeys()`, `itervalues()` or `iteritems()` it will trigger the error when called, even though we are not referring to an instance of `dict`, but to, for example, an instance of `AvocadoParam`.
> > On the other hand, I see your point and it might be useful to change these names that conflict with `dict.iter*`methods to avoid confusion when the check is enabled, specially for new code that uses `AvocadoParam` and `AvocadoParams`. Maybe change these methods to something like `iter_items()`, what do you think?
> 
> @willianrampazzo, @tiagohonorato about this problem I already opened discussion #4414, where we can decide the best approach.

Thanks, @richtja!

willianrampazzo:
> Hi @willianrampazzo, the check was disabled in some places because I see those as false positives. To issue the warning, Pylint seems to just check whether methods named `iterkeys()`, `itervalues()` and `iteritems()` are called, not taking the instance in consideration, in other words, if it is a `dict` (I will take a closer look at Pylint's codebase and open an issue there if it is confirmed). Consequently, if we define a method named `iterkeys()`, `itervalues()` or `iteritems()` it will trigger the error when called, even though we are not referring to an instance of `dict`, but to, for example, an instance of `AvocadoParam`.
> 
> On the other hand, I see your point and it might be useful to change these names that conflict with `dict.iter*`methods to avoid confusion when the check is enabled, specially for new code that uses `AvocadoParam` and `AvocadoParams`. Maybe change these methods to something like `iter_items()`, what do you think?

I see your point and I agree it makes sense.

willianrampazzo (approved):

beraldoleal (changes requested):
@tiagohonorato, thanks for this. LGTM, but my only suggestion is to rebase this on top of the master and remove the "merge" commit.

tiagohonorato:
@beraldoleal, sure!

tiagohonorato:
I decided to keep ``ValueDict.iteritems()`` and ``ValueDict.items()`` here, as that may need further discussion and keeping those shouldn't cause any problems.

beraldoleal (approved):

PR #4427, opened 2021-03-01
Assets: small fix on Asset constructor consistency

beraldoleal:
Commit 27614569 was introduced as a default value but without effect on the
constructor signature. This small change will allow objects to be
constructed without having to pass None or `''` to the first argument
while keeps `''` as the default value.

### DIFF
--- a/avocado/utils/asset.py
+++ b/avocado/utils/asset.py
@@ -50,11 +50,11 @@ class Asset:
     Try to fetch/verify an asset file from multiple locations.
     """
 
-    def __init__(self, name, asset_hash=None, algorithm=None, locations=None,
-                 cache_dirs=None, expire=None, metadata=None):
+    def __init__(self, name=None, asset_hash=None, algorithm=None,
+                 locations=None, cache_dirs=None, expire=None, metadata=None):
         """Initialize the Asset() class.
 
-        :param name: the asset filename. url is also supported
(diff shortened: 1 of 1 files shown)


avocado/utils/asset.py
```diff
@@ -50,11 +50,11 @@ class Asset:
     Try to fetch/verify an asset file from multiple locations.
     """
 
-    def __init__(self, name, asset_hash=None, algorithm=None, locations=None,
-                 cache_dirs=None, expire=None, metadata=None):
+    def __init__(self, name=None, asset_hash=None, algorithm=None,
```
richtja, line 53: Hi @beraldoleal, why are you not using `name=''` as default?

beraldoleal: Just to follow the same pattern on other arguments. In this case, it will be converted to '' anyway. But yes, it could be `name=''`

richtja (approved):

willianrampazzo (approved):

clebergnu (changes requested):
The commit message refers to a commit which is not related to this change.  Maybe you meant 229d64f99579b663fbc224a12c39d740d13b5fc5 ?

Other than that, it LGTM.

beraldoleal:
@clebergnu nice catch. Thanks. Fixed with the force-push.

clebergnu (approved):
LGTM, thanks!

PR #4428, opened 2021-03-01
utils/multipath: Fix the module's sphinx documentation

beraldoleal:
The sphinx generated from utils/multipath.py docstrings
ins't formatted correctly. Fixed it. Also made the text
a little bit more uniform.

This is a one year PR and is not intent to do a huge refactoring on those
docstrings. There is still room for improvements.

Signed-off-by: Wainer dos Santos Moschetta <wainersm@redhat.com>

### DIFF
--- a/avocado/utils/multipath.py
+++ b/avocado/utils/multipath.py
@@ -71,7 +71,9 @@ def device_exists(mpath):
     """
     Checks if a given mpath exists.
 
+    :param mpath: The multipath path
     :return: True if path exists, False if does not exist.
+    :rtype: bool
     """
     cmd = "multipath -ll"
     out = process.run(cmd, ignore_status=True, sudo=True,
@@ -86,7 +88,6 @@ def get_mpath_name(wwid):
     Get multipath name for a given wwid.
 
     :param wwid: wwid of multipath device.
-
     :return: Name of multipath device.
     :rtype: str
     """
@@ -167,8 +168,11 @@ def get_paths(wwid):
 
 def get_multipath_details():
     """
-    Get multipath details as a dictionary, as given by the command:
-    multipathd show maps json
+    Get multipath details as a dictionary.
+
+    This is the output of the following command:
+
+      $ multipathd show maps json
 
     :return: Dictionary of multipath output in json format
     :rtype: dict
@@ -186,7 +190,6 @@ def is_path_a_multipath(disk_path):
     Check if given disk path is part of a multipath.
 
     :param disk_path: disk path. Example: sda, sdb.
-
     :return: True if part of multipath, else False.
     """
     if not process.system("multipath -c /dev/%s" % disk_path, sudo=True,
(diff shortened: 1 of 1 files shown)


avocado/utils/multipath.py
```diff
@@ -71,7 +71,9 @@ def device_exists(mpath):
     """
     Checks if a given mpath exists.
 
+    :param mpath: The multipath path
     :return: True if path exists, False if does not exist.
+    :rtype: Boolean
```
pevogam, line 76: I am not sure what is the de facto standard for avocado docstrings, at least for our plugin we use `bool`, `str`, `int` and so on. Some of the docstrings below seem to be missing `rtype` for return types, as well as type descriptions. But this might be squeezing your contribution too much, just noting down here because the current docstring handling in this module is not uniform.

avocado/utils/multipath.py
```diff
@@ -215,9 +217,10 @@ def get_path_status(disk_path):
 
 def fail_path(path):
     """
-    failing the individual paths
-    :param disk_path: disk path. Example: sda, sdb.
-    :return: True or False
+    Fail the individual paths.
+
+    :param path: disk path. Example: sda, sdb.
+    :return: True if succeeded, False otherwise
```
pevogam, line 223: For instance something like this I would write like this (to minimize redundancy):
```
:returns: whether ... succeeded
:rtype: bool
```
I don't really understand what it means for a path to succeed or not but just visualizing here.

avocado/utils/multipath.py
```diff
@@ -233,9 +236,10 @@ def is_failed():
 
 def reinstate_path(path):
     """
-    reinstating the individual paths
-    :param disk_path: disk path. Example: sda, sdb.
-    :return: True or False
+    Reinstate the individual paths.
+
+    :param path: disk path. Example: sda, sdb.
```
pevogam, line 241: We also use `:param str path: ...` one liners which seems to work in sphinx too but again it depends what your standards here are. Otherwise this should have another line with the path type.

avocado/utils/multipath.py
```diff
@@ -371,7 +380,8 @@ def is_mpath_added():
 
 def remove_path(path):
     """
-    removing the individual paths
+    Remove the individual paths.
```
pevogam, line 383: Very nice that this was fixed, much cleaner to read.

pevogam (commented):
Left some of my previous general comments here in case you are interested to accommodate any of them.

avocado/utils/multipath.py
beraldoleal: I agree, but as I said, I will leave those improvements for future PRs.

avocado/utils/multipath.py
beraldoleal: It is a nice suggestion, but I will change only in a few spots and leave other changes for the future.

pevogam (approved):
Alright, as there is no clearly established standard for the docstrings atm and I doubt already merged docstrings are much stricter, LGTM.

willianrampazzo (approved):

Issue #4425, opened 2021-02-24
Test independent on Job class


richtja:
In the legacy runner Test class have instance of Job which provide  logging for Test. But in the nrunner instrumented tests the test class have to be independent because the nrunner runs separately from avocado core. Right now the instrumented runner deals with this problem by creating a new [job instance](https://github.com/avocado-framework/avocado/blob/master/avocado/core/nrunner_avocado_instrumented.py#L41) for each test, which is not good approach. We have to create Test specific logging for test log files, and make Test class independent on Job to be easily run separately in instrumented nrunner.

PR #4429, opened 2021-03-02
Test independent on Job class

richtja:
This commit removes the Job instance from the Test class. This is
preparation for legacy runner removal. For the nrunner doesn't make
sense to have Test dependent on Job instance, because Test is run inside
runner separated from Job.

Reference: #4425

### DIFF
--- a/avocado/core/test.py
+++ b/avocado/core/test.py
@@ -228,7 +228,7 @@ class Test(unittest.TestCase, TestData):
     timeout = None
 
     def __init__(self, methodName='test', name=None, params=None,
-                 base_logdir=None, job=None, runner_queue=None, tags=None):
+                 base_logdir=None, config=None, runner_queue=None, tags=None):
         """
         Initializes the test.
 
@@ -242,7 +242,9 @@ class Test(unittest.TestCase, TestData):
         :type name: :class:`avocado.core.test.TestID`
         :param base_logdir: Directory where test logs should go. If None
                             provided a temporary directory will be created.
-        :param job: The job that this test is part of.
+        :param config: the job configuration, usually set by command
+                       line options and argument parsing
+        :type config: dict
         """
         self.__phase = 'INIT'
 
@@ -257,8 +259,9 @@ class Test(unittest.TestCase, TestData):
         else:
             self.__name = TestID(0, self.__class__.__name__)
 
-        self.__job = job
         self.__tags = tags
+
+        self.__config = config or settings.as_dict()
 
         self.__base_logdir_tmp = None
         if base_logdir is None:
@@ -282,12 +285,8 @@ class Test(unittest.TestCase, TestData):
 
         self.__outputdir = utils_path.init_dir(self.logdir, 'data')
 
-        # For some reason, sometimes, job.config if None here.
-        # This is the only place that we create a "default" in code.
-        try:
-            self.__sysinfo_enabled = job.config.get('sysinfo.collect.per_test')
-        except AttributeError:
-            self.__sysinfo_enabled = False
+        self.__sysinfo_enabled = self.__config.get('sysinfo.collect.per_test',
+                                                   False)
 
         if self.__sysinfo_enabled:
             self.__sysinfodir = utils_path.init_dir(self.logdir, 'sysinfo')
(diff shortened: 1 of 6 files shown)


clebergnu:
Hi @richtja , I've looked and tested the code, and this looks to be in the right direction.  It will be a relief to finally get rid of the test -> job dependency.

Anyway, what we need here though, is a compatibility handling strategy.  Basically, by changing the signature of the `avocado.core.test.Test` class, it impacts all other code that inherits from it.  You've tackled the changes on this repo, but we have to consider external users too.

For instance, this would be needed to get Avocado-VT working with this change:

```diff
diff --git a/avocado_vt/test.py b/avocado_vt/test.py
index c518d8824..f6d147833 100644
--- a/avocado_vt/test.py
+++ b/avocado_vt/test.py
@@ -85,7 +85,7 @@ class VirtTest(test.Test):
     env_version = utils_env.get_env_version()
 
     def __init__(self, methodName='runTest', name=None, params=None,
-                 base_logdir=None, job=None, runner_queue=None,
+                 base_logdir=None, config=None, runner_queue=None,
                  vt_params=None):
         """
         :note: methodName, name, base_logdir, job and runner_queue params
@@ -112,7 +112,7 @@ class VirtTest(test.Test):
         self.background_errors.clear()
         super(VirtTest, self).__init__(methodName=methodName, name=name,
                                        params=params,
-                                       base_logdir=base_logdir, job=job,
+                                       base_logdir=base_logdir, config=config,
                                        runner_queue=runner_queue)
         self.builddir = os.path.join(self.workdir, 'backends',
                                      vt_params.get("vm_type", ""))

```

But then, it would only work with Avocado versions containing the change you're proposing.  A better version of such a patch would handle the different signatures.  I suggest you do the following next:

1. Work on the Avocado-VT compatibility
2. Add an Avocado-VT Cirrus CI jobs (just for the sake of the review process, a commit that will be throw away) that also uses Avocado containing the change you're proposing

This way, we'd make sure the compatibility is kept with all previous versions, and will be kept when your proposal (this PR) is merged.

avocado/core/test.py
```diff
@@ -257,9 +259,12 @@ def record_and_warn(*args, **kwargs):
         else:
             self.__name = TestID(0, self.__class__.__name__)
 
-        self.__job = job
         self.__tags = tags
 
+        self.__config = config
+        if self.__config is None:
+            self.__config = settings.as_dict()
+
```
beraldoleal, line 267: Suggestion, maybe this:

```python
self.__config = config or settings.as_dict()
```

avocado/core/test.py
```diff
@@ -797,9 +791,9 @@ def _run_avocado(self):
         whiteboard_file = os.path.join(self.logdir, 'whiteboard')
         genio.write_file(whiteboard_file, self.whiteboard)
 
-        if self.job is not None:
-            output_check_record = self.job.config.get('run.output_check_record')
-            output_check = self.job.config.get('run.output_check')
+        if self.__config is not None:
+            output_check_record = self.__config.get('run.output_check_record')
```
beraldoleal, line 795: With my previous suggestion, `self.__config` will always have a dict even empty, right? So maybe this if is unecessary.

richtja:
> But then, it would only work with Avocado versions containing the change you're proposing. A better version of such a patch would handle the different signatures. I suggest you do the following next:
> 
>     1. Work on the Avocado-VT compatibility
> 
>     2. Add an Avocado-VT Cirrus CI jobs (just for the sake of the review process, a commit that will be throw away) that also uses Avocado containing the change you're proposing
> 
>     3. Tell the community (a post to avocado-devel) about the upcoming change, and suggest the Avocado-VT change as a model for other extensions to keep compatibility with Avocado 86.0 and later.
> 
> 
> This way, we'd make sure the compatibility is kept with all previous versions, and will be kept when your proposal (this PR) is merged.


@clebergnu thanks for your suggestions. I will do that.

avocado/core/test.py
richtja: sure, thanks

pevogam (commented):
I have a few very superficial comments and questions regarding this PR, mostly things that arose from the VT-side PR.

pevogam (approved):
These changes are no doubt good, my comments are mostly meta.

avocado/core/test.py
```diff
@@ -282,10 +285,10 @@ def record_and_warn(*args, **kwargs):
 
         self.__outputdir = utils_path.init_dir(self.logdir, 'data')
 
-        # For some reason, sometimes, job.config if None here.
+        # For some reason, sometimes, config if None here.
```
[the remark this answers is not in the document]
beraldoleal: Yes, this is strange. @richtja are you be able to reproduce this easily? Can we document this as an issue?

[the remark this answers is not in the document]
richtja: Thanks for catching this, I miss it in refactoring. This problem is not happening when we use `config` instead of `job.config`. I will change that in force push.

beraldoleal (approved):
LGTM

clebergnu (approved):
LGTM, thanks!

PR #4436, opened 2021-03-04
Remove class inheriting from object

tiagohonorato:
Unlike in python2, python3 classes inheritance from object is implicit. As
a result, since python2 is no longer supported in Avocado, this can be
safely removed.

Related to #3466


richtja (changes requested):
Hi @tiagohonorato, I find another occurrences in [dont_crash.py](https://github.com/avocado-framework/avocado/blob/master/selftests/.data/loader_instrumented/dont_crash.py). Can you please add those as well. It looks like lint check doesn't consider such files. I create issue #4437 about this problem.

beraldoleal (approved):
Besides the @richtja recommendation, this LGTM.

tiagohonorato:
> Hi @tiagohonorato, I find another occurrences in [dont_crash.py](https://github.com/avocado-framework/avocado/blob/master/selftests/.data/loader_instrumented/dont_crash.py). Can you please add those as well. It looks like lint check doesn't consider such files. I create issue #4437 about this problem.

Hi @richtja, thank you for this. I will include it here.

richtja (approved):
LGTM, thanks

PR #4442, opened 2021-03-05
get_disks_in_pci_address: add a path check

abdhaleegit:
If the test system do not have any physical disks, the path /dev/disk/by-path/
is invalid, make sure the path exists before it proceeds

fixes ERROR: [Errno 2] No such file or directory: '/dev/disk/by-path/'

### DIFF
--- a/avocado/utils/pci.py
+++ b/avocado/utils/pci.py
@@ -97,6 +97,8 @@ def get_disks_in_pci_address(pci_address):
     """
     disks_path = "/dev/disk/by-path/"
     disk_list = []
+    if not os.path.exists(disks_path):
+        return disk_list
     for dev in os.listdir(disks_path):
         if pci_address in dev:
             link = os.readlink(os.path.join(disks_path, dev))


abdhaleegit:
fixes this issue https://github.com/avocado-framework-tests/avocado-misc-tests/issues/2019

avocado/utils/pci.py
```diff
@@ -97,10 +97,11 @@ def get_disks_in_pci_address(pci_address):
     """
     disks_path = "/dev/disk/by-path/"
     disk_list = []
-    for dev in os.listdir(disks_path):
-        if pci_address in dev:
-            link = os.readlink(os.path.join(disks_path, dev))
-            disk_list.append(os.path.abspath(os.path.join(disks_path, link)))
```
beraldoleal, line 103: My only suggestion here to reduce complexity is to invert the logic:
```python
if not os.path.exists(disks_path):
   return disk_list
for dev in ....
```

richtja (approved):
Besides the @beraldoleal recommendation, LGTM.

avocado/utils/pci.py
abdhaleegit: @beraldoleal Thanks

beraldoleal (commented):
@abdhaleegit thank you for the update.

This looks good to me, but we could keep the test coverage here. Could you please, add some test for this case ? I guess `selftests/unit/test_utils_pci.py` is the right place here.

willianrampazzo (approved):

PraveenPenguin (approved):

beraldoleal (approved):
@abdhaleegit I just saw that mocking `/dev/disk/by-path` would not be so trivial. So yes, we can leave this for a second round. I will just wait for a Travis run and I will merge.

PR #4444, opened 2021-03-05
results.json: output tags as well

beraldoleal:
Some of QEMU requirements is to have all the tags inside results.json output
file. Since QEMU has intentions to use tags to associate tests with "test
owners", this will allow CI admins to run scripts to print the owners of failed
tests. This is related to #4430.


wainersm:
Hi @beraldoleal , first of all, thanks for promptly working on that!

I've found a problem. A avocado tag on the test can be a `key:value` or just `key` string, but the current implementation only list the 'key' portion. To reproduce that you can use a modifed version of the `passtest.py` as:

```
diff --git a/examples/tests/passtest.py b/examples/tests/passtest.py
index c255c2ab..84f6ec94 100644
--- a/examples/tests/passtest.py
+++ b/examples/tests/passtest.py
@@ -6,10 +6,11 @@ class PassTest(Test):
     """
     Example test that passes.
 
-    :avocado: tags=fast
     """
 
     def test(self):
         """
+
+        :avocado: tags=arch:x86_64
         """
```

Above results in:

```
            "start": 1614976223.4215314,
            "status": "PASS",
            "tags": [
                "arch"
            ],
            "time": 0.002828359603881836,

```

beraldoleal:
@wainersm I got it. What about now?

wainersm:
> @wainersm I got it. What about now?

It is failing with:
```
Error running method "render" of plugin "json": Object of type set is not JSON serializable
```

beraldoleal:
@wainersm probably this is because you are using the legacy runner.  Nice catch.

--test-runner='nrunner' should fix this. But indeed, I need to fix on legacy too. I will investigate better.

avocado/utils/asset.py
```diff
@@ -50,11 +50,11 @@ class Asset:
     Try to fetch/verify an asset file from multiple locations.
     """
 
-    def __init__(self, name, asset_hash=None, algorithm=None, locations=None,
-                 cache_dirs=None, expire=None, metadata=None):
+    def __init__(self, name=None, asset_hash=None, algorithm=None,
+                 locations=None, cache_dirs=None, expire=None, metadata=None):
```
richtja, line 54: Hi @beraldoleal I think that this commit is not related to this PR and was already merged in #4427

richtja (changes requested):

beraldoleal:
@richtja thank you for the review. Fixed with the force-push.

@wainersm can you please test it now?

wainersm (approved):
@beraldoleal Thanks for working on that!

I tested this implementation with following cases:
```
fast
arch:x86_64
arch:x86_64,fast
arch:x86_64,fast,arch:aarch64
```

richtja (approved):

PR #4448, opened 2021-03-10
Failed tests summary

richtja:
Output of selftests check with summary of all failed tests.

Now the check output will look like this:
```
...
 (check-852/956) selftests/lint.sh: FAIL (91.79 s)
RESULTS    : PASS 956 | ERROR 1 | FAIL 3 | SKIP 44 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB HTML   : /home/jarichte/avocado/job-results/job-2021-03-10T15.40-f7a9604/results.html
JOB TIME   : 462.65 s
Failed tests:
'check-417-selftests/unit/test_test.py:MockingTest.test_combination_2'  ERROR
'check-416-selftests/unit/test_test.py:MockingTest.test_combination'  FAIL
'check-772-selftests/functional/test_loader.py:LoaderTestFunctional.test_python_unittest'  FAIL
'check-852-selftests/lint.sh'  FAIL
```

### DIFF
--- a/selftests/check.py
+++ b/selftests/check.py
@@ -3,6 +3,7 @@
 import argparse
 import glob
 import os
+import sqlite3
 import sys
 
 from avocado import Test
@@ -518,6 +519,19 @@ def create_suites(args):
     return suites
 
 
+def print_failed_tests(job_log_dir):
+    jfile = os.path.join(job_log_dir, '.journal.sqlite')
+    db = sqlite3.connect(jfile)
+    cur = db.cursor()
+    cur.execute("SELECT * FROM test_journal WHERE status='FAIL' OR"
+                " status='ERROR';")
+    fail_tests = cur.fetchall()
+    if fail_tests:
+        print("Failed tests:")
+        for status in fail_tests:
+            print("%s  %s" % (status[0], status[3]))
+
+
 def main():
     args = parse_args()
     suites = create_suites(args)
@@ -539,9 +553,12 @@ def main():
     # Job execution
     # ========================================================================
     config = {'core.show': ['app'],
-              'run.test_runner': 'nrunner'}
+              'run.test_runner': 'nrunner',
+              'run.journal.enabled': True}
     with Job(config, suites) as j:
-        return j.run()
+        exit_code = j.run()
+    print_failed_tests(j.logdir)
+    return exit_code
 
 
 if __name__ == '__main__':


selftests/check.py
```diff
@@ -518,6 +519,19 @@ def create_suites(args):
     return suites
 
 
+def print_failed_tests(job_log_dir):
+    jfile = os.path.join(job_log_dir, '.journal.sqlite')
```
clebergnu, line 523: Interesting that you found a use for the journal, and at the same time, it's a kind of extra check *for* the journal.

selftests/check.py
```diff
@@ -539,9 +553,12 @@ def main():
     # Job execution
     # ========================================================================
     config = {'core.show': ['app'],
-              'run.test_runner': 'nrunner'}
+              'run.test_runner': 'nrunner',
+              'run.journal.enabled': True}
     with Job(config, suites) as j:
-        return j.run()
+        exit_code = j.run()
+        print_failed_tests(j.logdir)
```
clebergnu, line 560: Nipick: this and the following line can be put outside of the context manager (and probably should as an extra guarantee the job is finished).

selftests/check.py
```diff
@@ -518,6 +519,19 @@ def create_suites(args):
     return suites
 
 
+def print_failed_tests(job_log_dir):
+    jfile = os.path.join(job_log_dir, '.journal.sqlite')
+    db = sqlite3.connect(jfile)
+    cur = db.cursor()
+    cur.execute("SELECT * FROM test_journal WHERE status='FAIL' OR"
+                " status='ERROR';")
```
beraldoleal, line 527: Is journal a "required plugin"? I was thinking more about getting it directly from the job/tests list. And make this more abstract.  Depending on SQL syntax there might be a yellow flag for future decoupling.
 
In my mind we could do something like this:

```
job = Job()
job.run()
failed = job.get_failed_tests()
```

But of course that I didn't have the chance to see the problems with this approach, so take it with a grant of salt.

richtja: I get your point, true is that I didn't consider the future decoupling, but I don't want to create a new feature because of this. I want to use existing features to show how it can be used. Because I think that if we have a usage for some plugin in our testing we should use it.

richtja: And when we will do the decoupling the journal plugin can be as `selftest-requirements`

beraldoleal: Having an usage for a plugin IMO should not be the motivation. But I understand some of the motivation, so I will not block this, of course. But have you tried to use ` job.result.tests`? Any reason for not using this?

Like I said, if possible I would like to use our objects and proprieties instead of hardcoded SQL. This gives us more flexibility.

willianrampazzo: I'm not against using plugins to run our self-tests. Like @clebergnu mentioned once to me, this is a nice place to show Avocado's capabilities.

OTOH, I agree that we should avoid adding dependencies to the Avocado core other than pure python packages.

clebergnu:
If you mean sqlite3, it's part of the standard library.

selftests/check.py
richtja: > Having an usage for a plugin IMO should not be the motivation. But I understand some of the motivation, so I will not block this, of course. But have you tried to use ` job.result.tests`? Any reason for not using this?

@beraldoleal as I said my motivation was to create and example of usage of existing plugin, but of course I can use `job.result.tests` and create `job.get_failed_tests()`. I just think that we should show what avocado plugins can do.

beraldoleal: @richtja I agree that we should show off our plugins and our potential. But IMO this is a bad example. Redirecting users to use SQL instead of nice python attributes, could be discouraging for new users.

selftests/check.py
```diff
@@ -539,9 +553,12 @@ def main():
...
     # ========================================================================
     config = {'core.show': ['app'],
-              'run.test_runner': 'nrunner'}
+              'run.test_runner': 'nrunner',
+              'run.journal.enabled': True}
     with Job(config, suites) as j:
-        return j.run()
+        exit_code = j.run()
+    print_failed_tests(j.logdir)
+    return exit_code
```
beraldoleal, line 561: For instance: I'm trying to understand why having a journal dependency and the SQL code is better than this:

```python
for test in job.result.tests: 
    if test.get('status') in ['FAIL', 'ERROR']: 
        print(test.get('name'), test.get('status')) 
```

Probably I'm missing something.

selftests/check.py
richtja: Ok, I think you are right, it's too much complicated. I will create a new version.

clebergnu: > Ok, I think you are right, it's too much complicated. I will create a new version.

If you don't mind, I think we can live with this version until we have time to create, and properly document, a results API that is sound enough and well documented.  I don't think it will me a trivial task.

selftests/check.py
clebergnu: There's little structure to the individual `test` instances.  A believe a task in itself is to document what users should expect to find here (and maybe even convert those from dicts into more structured objects).  Like I said elsewhere, I believe this a larger (but good) task.

beraldoleal: I didn't understand your first sentence.

I agree that we should document what users should expect from our API, but I missing the point to reject this solution and stick with the current one, which is getting even deeper, looking at table structures on a SQL database on "tests" results. Not sure if I understood why this more suitable than querying those Python attributes.

Like I said I will not reject this PR, of course, but I still find strange have this SQL code here.

clebergnu: > I didn't understand your first sentence.
> 
> I agree that we should document what users should expect from our API, but I missing the point to reject this solution and stick with the current one, which is getting even deeper, looking at table structures on a SQL database on "tests" results. Not sure if I understood why this is more suitable than querying those Python attributes.
> 

What I mean was that each `test` instance is a dictionary, with keys that well, users can not currently have many expectations about.  The upside of the journal file, is that it's more structured.

Now, we all agree that in the end, we want a nice Python API for the results of a job (and tests).

If you've tried the code snippet you proposed, and that's that, then maybe @richtja will be able to quickly come up with a new version.  What I wouldn't like is for this to be delayed waiting for an alternative implementation.

> Like I said I will not reject this PR, of course, but I still find strange have this SQL code here.

beraldoleal: So we are on the same page here. I tried the code in an isolated job API usage. Not inside this code. 

@richtja is up to you, sr.

richtja:
I created a new version without the journal plugin. #4453

PR #4451, opened 2021-03-12
Refactor linting config files

tiagohonorato:
As we have most warnings supported on pylint enabled, we could make the files more
readable by only disabling the warnings we are not using for now.

Hopefully we will get to a state that we will not need to disable all refactor (R) and 
convention (C) checks (similarly to the current state with warning checks), having most 
enabled by default and only disabling a few checks.

Some checks previously disabled (E1002, E1103) were removed since they don't seem 
to exist any longer on pylint. In addition, F0401 was changed to E0401.

Since python3 checks (W16..) are not enabled by default, those still need to be enabled
manually.

Related to #3466


richtja (approved):
LGTM, thanks

beraldoleal (approved):
LGTM

willianrampazzo (approved):

PR #4474, opened 2021-03-23
Introduce a very basic GH actions pipeline (v2)

beraldoleal:
This will enable GitHub workflows as part of an effort on running our tests in GitHub infrastructure. More details on individual commits.

Since this is not enabled here yet, you can see the output here:

https://github.com/beraldoleal/avocado/actions/runs/680460902

#### Changes from v1:

 * Installing avocado in develop mode and avoiding pylint hook;
 * Exit status is now 128 in case of error during lint command;
 *  `pyenchant==3.2.0`
 * `check_call` without `shell=True`;
 * Adding a separation line for each check command;


beraldoleal:
Wrong branch.

PR #4476, opened 2021-03-23
Make BaseRunner an abstract class

tiagohonorato:
The idea for this PR came during the discussions over #4441. I guess this is more a design decision than anything else. The idea here is to properly enforce the implementation of ``run()`` in subclasses as well as to avoid direct instantiation from ``BaseRunner``.

### DIFF
--- a/avocado/core/nrunner.py
+++ b/avocado/core/nrunner.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 
+import abc
 import argparse
 import base64
 import collections
@@ -312,7 +313,7 @@ class Runnable:
         raise ValueError('Unsupported kind of runnable: %s' % self.kind)
 
 
-class BaseRunner:
+class BaseRunner(metaclass=abc.ABCMeta):
     """
     Base interface for a Runner
     """
@@ -340,8 +341,13 @@ class BaseRunner:
             status.update(additional_info)
         return status
 
+    @abc.abstractmethod
     def run(self):
-        yield {}
+        """Runner main method
+
+        Yields dictionary as output, containing status as well as relevant
+        information concerning the results.
+        """
 
 
 class NoOpRunner(BaseRunner):


tiagohonorato:
@willianrampazzo and @richtja, would this be in line with what we discussed over #4441?

richtja (approved):
LGTM, thanks

clebergnu (approved):
LGTM, thanks!

avocado/core/nrunner.py
```diff
@@ -312,7 +313,7 @@ def pick_runner_class(self, runners_registry=None):
         raise ValueError('Unsupported kind of runnable: %s' % self.kind)
 
 
-class BaseRunner:
+class BaseRunner(metaclass=abc.ABCMeta):
```
pevogam, line 316: @tiagohonorato I wonder if this could simply be
```python
class BaseRunner(abc.ABC):
```
Direct (if not any) usage of metaclasses is discouraged and in this case not necessary since inheriting from `ABC` could rid is of such python internals and keep normal class inheritance on our side.

pevogam (commented):
A bit delayed but still curious to see what you think.

avocado/core/nrunner.py
tiagohonorato: Hi @pevogam, sorry for the wait here and thanks for pointing that out!

You are right. After researching a bit, since python 3.4, inheriting from ``ABC`` defines the metaclass implicitly:

```python

class ABC(metaclass=ABCMeta):
    """Helper class that provides a standard way to create an ABC using
    inheritance.
    """
    __slots__ = ()
```

So, yeah, I agree that keeping normal inheritance would be way better!

tiagohonorato: Currently, from what I could find, these are the only places ```ABCMeta``` is used:

```
avocado/core/plugin_interfaces.py:19:class Plugin(metaclass=abc.ABCMeta):
avocado/core/nrunner.py:316:class BaseRunner(metaclass=abc.ABCMeta):
avocado/utils/datadrainer.py:32:class BaseDrainer(metaclass=abc.ABCMeta):
```

pevogam: Hi @tiagohonorato, sounds good. Nice to see that you are willing to take this a step further and cover other instances of this improvement!

PR #4480, opened 2021-03-25
add function to skip dmesg error messages

smruti77:
some of the dmesg error messages which is not actual failure from test,
so wanted to skip those error messages during collect error message
by level.

### DIFF
--- a/avocado/utils/dmesg.py
+++ b/avocado/utils/dmesg.py
@@ -84,7 +84,7 @@ def collect_errors_dmesg(patterns):
     return error
 
 
-def collect_errors_by_level(output_file=None, level_check=5):
+def collect_errors_by_level(output_file=None, level_check=5, skip_errors=None):
     """Verify dmesg having severity level of OS issue(s).
 
     :param output_file: The file used to save dmesg
@@ -97,6 +97,8 @@ def collect_errors_by_level(output_file=None, level_check=5):
                         4 - emerg,alert,crit,err
                         5 - emerg,alert,crit,err,warn
     :type level_check: int
+    :skip_errors: list of dmesg error messages which want skip
+    :type skip_errors: list
     """
     if not isinstance(level_check, int):
         raise DmesgError("level_check param should be integer")
@@ -106,7 +108,11 @@ def collect_errors_by_level(output_file=None, level_check=5):
                       verbose=False, shell=True)
     if out.exit_status == 0:
         err = "Found failures in dmesg"
-        dmsg_log = "dmesg log:\n%s" % out.stdout_text
+        if skip_errors:
+            dmsg_log = skip_dmesg_messages(out.stdout_text, skip_errors)
+        else:
+            dmsg_log = "dmesg log:\n%s" % out.stdout_text
+    if dmsg_log:
         if output_file:
             with open(output_file, "w+") as log_f:
                 log_f.write(dmsg_log)
@@ -115,3 +121,23 @@ def collect_errors_by_level(output_file=None, level_check=5):
             err += " Please check  dmesg log in debug log."
             LOGGER.debug(dmsg_log)
         raise DmesgError("Test is failed {}".format(err))
+
+
+def skip_dmesg_messages(dmesg_stdout, skip_messages):
+    """Remove some messages from a dmesg buffer.
+
+      This method will remove some lines in a dmesg buffer if some strings are
+      present. Returning the same buffer, but with less lines (in case of match).
+
+      :dmesg_stdout: dmesg messages from which filter should be applied. This
+                     must be a decoded output buffer with new lines.
+      :type dmesg_stdout: string
+      :skip_messages: list of strings to be removed
+      :type skip_messages: list
+    """
+    def filter_strings(line):
+        return not any([string in line for string in skip_messages])
+
+    return '\n'.join(filter(None,
+                            filter(filter_strings,
+                                   dmesg_stdout.splitlines())))


avocado/utils/dmesg.py
```diff
@@ -115,3 +122,19 @@ def collect_errors_by_level(output_file=None, level_check=5):
             err += " Please check  dmesg log in debug log."
             LOGGER.debug(dmsg_log)
         raise DmesgError("Test is failed {}".format(err))
+
+
+def skip_dmesg_error(dmesg_error, skip_dmesg):
+    """remove dmesg error which want to skip.
+
```
beraldoleal, line 129: s/^remove/^Remove/

avocado/utils/dmesg.py
```diff
@@ -115,3 +122,19 @@ def collect_errors_by_level(output_file=None, level_check=5):
...
+    :dmesg_error: dmesg messages
```
beraldoleal, line 130: If this is a string, I' m assuming this is a single dmesg line. Did I understood right?

avocado/utils/dmesg.py
```diff
@@ -84,7 +85,7 @@ def collect_errors_dmesg(patterns):
     return error
 
 
-def collect_errors_by_level(output_file=None, level_check=5):
+def collect_errors_by_level(output_file=None, level_check=5, skip_error=None):
```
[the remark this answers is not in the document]
beraldoleal: This method is getting "too complex" and doing multiple things. My suggestion is to keep using as it was and use your new method to filter:

```python
messages = collect_erros_by_level()
filtered = skip_dmesg_error(messages, filter)
```

avocado/utils/dmesg.py
beraldoleal: If not , we should choose better names here.

beraldoleal (changes requested):

avocado/utils/dmesg.py
smruti77: not single line,its multiple line of dmesg.

avocado/utils/dmesg.py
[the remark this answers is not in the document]
smruti77: this is fine but collect_erros_by_level() will not return the dmesg messages it will raise exception if any messages found. if i will do this approach i need to do some changes in collect_erros_by_level(). so i have included directly in that function.

avocado/utils/dmesg.py
```diff
@@ -115,3 +122,19 @@ def collect_errors_by_level(output_file=None, level_check=5):
             err += " Please check  dmesg log in debug log."
             LOGGER.debug(dmsg_log)
         raise DmesgError("Test is failed {}".format(err))
+
+
+def skip_dmesg_error(dmesg_errors, skip_dmesg):
+    """Remove dmesg error which want to skip.
+
+    :dmesg_errors dmesg messages
+    :type dmesg_errors: string
```
beraldoleal, line 131: Could you explain better what you are passing here? I see that you are expecting commas (,) in that string. Is a list of lines separated by , ?

smruti77: dmesg_errors is lines of dmesg messages, which is output of dmesg error by level, here not expecting commas.
and expecting commas in skip_dmesg what user will pass for remove.

beraldoleal: Could you then, explain better this on the docstrings, please?

smruti77: ok

avocado/utils/dmesg.py
```diff
@@ -115,3 +122,19 @@ def collect_errors_by_level(output_file=None, level_check=5):
...
+    :skip_dmesg: list of dmesg error messages for remove
+    :type skip_dmesg: string
+    """
+    for dmesg in skip_dmesg.split(','):
+        for line in dmesg_errors.splitlines():
+            if dmesg in line:
+                dmesg_errors = dmesg_errors.replace(line, "")
+    dmesg_errors = re.sub(r'\s', '', dmesg_errors)
+    return dmesg_errors
```
beraldoleal, line 140: After reading your comments and the new docstrings again, I think this method needs some refactoring:

1. IIUC, the order of for loops is not efficient, considering that `skip_dmesg` are going to be a smaller list than the number of lines in a dmesg output. 
2. You are using `replace` in order to remove an entire line. This might have some false positives depending on the strings there.
3. You are removing all new lines (`\s`) making the buffer a linear buffer. This is more than what this method is describing. Also your caller method (`collect_errors_by_level()`) will have different output results, depending on the `skip_error` parameter.
4. This code is not only focused on "errors". this could be generic, because is basically doing some filters. So could be useful for other callers. I would recommend removing the "error" string here. 
5. We are inside a python code, would be nice to use proper types (like lists), instead of a "fake string list"
6. When dealing with a buffer like stdouts its is better to handle as bytes instead of strings. Also `process.run()` will return bytes instead of string.

So, again, IIUC, you are doing a `dmesg | egrep -v "some|words|here"`.  If my understanding is correct, IMO this could be an alternative method:

```python
def skip_dmesg_error(dmesg_errors, skip_dmesg):
    """Remove dmesg error messages which want to skip.

    :dmesg_errors: dmesg error messages from which want to filter
    :type dmesg_errors: string
    :skip_dmesg: list of dmesg error messages for remove
    :type skip_dmesg: string
    """
    for dmesg in skip_dmesg.split(','):
        for line in dmesg_errors.splitlines():
            if dmesg in line:
                dmesg_errors = dmesg_errors.replace(line, "")
    dmesg_errors = re.sub(r'\s', '', dmesg_errors)
    return dmesg_errors
```

Of course, we need to change the caller, too:

```diff
-def collect_errors_by_level(output_file=None, level_check=5):
+def collect_errors_by_level(output_file=None, level_check=5, skip_error=None):
     """Verify dmesg having severity level of OS issue(s).
 
     :param output_file: The file used to save dmesg
@@ -97,6 +98,8 @@ def collect_errors_by_level(output_file=None, level_check=5):
                         4 - emerg,alert,crit,err
                         5 - emerg,alert,crit,err,warn
     :type level_check: int
+    :skip_error: dmesg error messages which want skip
+    :type skip_error: string
     """
     if not isinstance(level_check, int):
         raise DmesgError("level_check param should be integer")
@@ -106,7 +109,11 @@ def collect_errors_by_level(output_file=None, level_check=5):
                       verbose=False, shell=True)
     if out.exit_status == 0:
         err = "Found failures in dmesg"
-        dmsg_log = "dmesg log:\n%s" % out.stdout_text
+        if skip_error:
+            dmsg_log = skip_dmesg_error(out.stdout_text, skip_error)
+        else:
+            dmsg_log = "dmesg log:\n%s" % out.stdout_text
+    if dmsg_log:
         if output_file:
             with open(output_file, "w+") as log_f:
                 log_f.write(dmsg_log)
@@ -115,3 +122,19 @@ def collect_errors_by_level(output_file=None, level_check=5):
             err += " Please check  dmesg log in debug log."
             LOGGER.debug(dmsg_log)
         raise DmesgError("Test is failed {}".format(err))
```

What do you think?

beraldoleal (changes requested):

avocado/utils/dmesg.py
smruti77: "replace" i am using to remove entire line,in which line contain the skip dmesg. in whole line with messages there is date time etc. things also there.so i am using replace for that. and after remove the message lines there will create empty line so using "\s" for that.

avocado/utils/dmesg.py
```diff
@@ -106,7 +109,11 @@ def collect_errors_by_level(output_file=None, level_check=5):
                       verbose=False, shell=True)
     if out.exit_status == 0:
         err = "Found failures in dmesg"
-        dmsg_log = "dmesg log:\n%s" % out.stdout_text
+        if skip_error:
+            dmsg_log = skip_dmesg_msg(out.stdout_text, skip_error)
```
beraldoleal, line 113: About `skip_error`: Here you receive a `string`, and `skip_dmesg_msg()` is a `list`. Since this is a "list of dmesg messages for remove" we need to change this signature to a list too or convert this to a list as I suggested in my previous comment.

I would prefer the first approach (make this a list too).

avocado/utils/dmesg.py
```diff
@@ -115,3 +122,19 @@ def collect_errors_by_level(output_file=None, level_check=5):
             err += " Please check  dmesg log in debug log."
             LOGGER.debug(dmsg_log)
         raise DmesgError("Test is failed {}".format(err))
+
+
+def skip_dmesg_msg(dmesg_messages, skip_dmesg):
+    """Remove dmesg messages which want to skip.
+
+    :dmesg_messages: dmesg messages from which want to filter
```
beraldoleal, line 130: dmesg_messages give us the idea of "list of things". Give us the idea of a "list"/plural. Since this is a pure stdout buffer, my suggestion is to rename this to "dmesg_buffer" or "dmesg_stdout".

avocado/utils/dmesg.py
```diff
@@ -115,3 +122,19 @@ def collect_errors_by_level(output_file=None, level_check=5):
...
+    :type dmesg_messages: string
+    :skip_dmesg: list of dmesg messages for remove
+    :type skip_dmesg: list
+    """
+    for dmesg in skip_dmesg:
+        for line in dmesg_messages.splitlines():
+            if dmesg in line:
```
beraldoleal, line 137: I could not run this code, since `dmesg_messages` (`process.stdout`) is a byte buffer object, and each `line` will be a byte as well. So unless I"m missing something, this needs a decode/encode properly.

```python
TypeError: a bytes-like object is required, not 'str'
```

Could you paste a basic use case here or on pastebin? A simple call on how to call this method?

beraldoleal (changes requested):

avocado/utils/dmesg.py
smruti77: skip_error: is list i forget to change here.

avocado/utils/dmesg.py
smruti77: here no need to decode/encode. i can able to run,not getting any error.this is just a sample how called form a test.
```
error = ['uevent: failed to send synthetic uevent', 'failed to send uevent']
dmesg.collect_errors_by_level(skip_error=error)
```

beraldoleal: I'm talking about the `skip_dmesg_msg()` method itself. If it is a public method how users will call this? Can you give an example?

beraldoleal: @smruti77 I missed that you are using `stdout_text` instead of `stdout`. This explains my error.

But I would recommend using `stdout` instead of text to avoid character issues related to `encode`.

Also in your example, shows why we should call this variable `skip_errors` instead of `skip_error`, since this is a list of errors.

smruti77: ```stdout_text``` we are using to avoid  ```encode```, ```decode``` error .  and  in ```skip_dmesg_msg()``` if someone want to pass ```stdout```,can pass with ```stdout.decode("utf-8")```. what i feel no need to handle ```encode/decode``` inside  the ```skip_dmesg_msg()```

beraldoleal: @smruti77 I just double-checked and `avocado.utils.process.run()` is calling `subprocess.run()` with the proper encoding. So, I'm ok with not doing the encoding here.

However, I still have a few suggestions:

1. s/skip_dmesg/skip_messages/g: according to your example, we are going to skip any string here, not only entire dmesg lines.

2. Your regex is not removing properly the empty lines, lefting a few blank lines. Take a very generic example, for instance with `skip_messages = ['usb', 'USB']` and you will see that multiple blank lines will be returned. This is because you are doing the replace on `dmesg_stdout`, and from the `re` PoV there are no multiple lines in this buffer. The EOL (`$`) will be only present the end of the buffer.

3. You don't need the last variable `dmesg_out`, since you can return directly:
```python
return re.sub()...
```
So, in order to filter lines, without leaving blank lines, I would go with something like this (feel free to adap):

```python
  def skip_dmesg_messages(dmesg_stdout, skip_messages):
      """Remove some messages from a dmesg buffer.
  
      This method will remove some lines in a dmesg buffer if some strings are
      present. Returning the same buffer, but with less lines (in case of match).
  
      :dmesg_stdout: dmesg messages from which filter should be applied. This
                     must be a decoded output buffer with new lines.
      :type dmesg_stdout: string
      :skip_messages: list of strings to be removed
      :type skip_messages: list
      """
      def filter_strings(line):
          return not any([string in line for string in skip_messages])
  
      return '\n'.join(filter(None,
                              filter(filter_strings,
                                     dmesg_stdout.splitlines())))

```

Thank you for your patience here.

smruti77: @beraldoleal thak you for suggestion. for space removing i have changed the code,if that is not fine, i will go with this approach.

avocado/utils/dmesg.py
```diff
@@ -115,3 +121,23 @@ def collect_errors_by_level(output_file=None, level_check=5):
...
         raise DmesgError("Test is failed {}".format(err))
+
+
+def skip_dmesg_messages(dmesg_stdout, skip_messages):
+    """Remove some messages form dmesg output.
+
+    This method will remove some lines form dmesg output if some
+    strings are present. Returning the filter output.
+
+    :dmesg_stdout: dmesg messages from which want to filter
```
beraldoleal, line 132: s/filter/filter./g

avocado/utils/dmesg.py
```diff
@@ -115,3 +121,23 @@ def collect_errors_by_level(output_file=None, level_check=5):
...
+                   it should be decoded string.
```
beraldoleal, line 133: s/it/It/g
s/be decoded/be a decoded/g

avocado/utils/dmesg.py
```diff
@@ -115,3 +121,23 @@ def collect_errors_by_level(output_file=None, level_check=5):
...
+    :type dmesg_stdout: string
+    :skip_messages: list of message for remove
+    :type skip_messages: list
+    """
+    for dmesg in skip_messages:
+        for line in dmesg_stdout.splitlines():
+            if dmesg in line:
+                dmesg_stdout = dmesg_stdout.replace(line, "")
+    filter_string = [line for line in dmesg_stdout.split('\n') if line.strip()]
```
beraldoleal, line 142: I don't understand why using this approach instead of a much simpler version. Do you have any reason for that?

Also, this is assuming that new lines boundaries are always going to be a `\n`. Today we have more or less 13 boundaries line separators (`\r`, `\n`, `\r\n`, `\v`, `\x0b`, `\u2028`, etc..) depending on your system. That is one of the reasons that you are using `\s` and `.splitlines()`, so they will take care of this for you.

Returning a `\n` as line separator it is OK, but unless I'm missing something, assuming that all separators are going to be `\n` IMO it is not good.

avocado/utils/dmesg.py
```diff
@@ -115,3 +121,23 @@ def collect_errors_by_level(output_file=None, level_check=5):
...
+    return '\n'.join(filter_string)
```
beraldoleal, line 143: Like I said, here find fine.

beraldoleal (changes requested):

avocado/utils/dmesg.py
smruti77: i don't have any reason,i just simplify that approach it might work out. i am completely fine with the approach what you have suggested will go with that.

beraldoleal (changes requested):
@smruti77 thank you for your patience, again. This LGTM. We just need to do a rebase here. Our new GitHub action is failing because of that.

beraldoleal (approved):

PR #4482, opened 2021-03-26
Remove iteritems() when inheriting from dict

tiagohonorato:
Since ``iteritems()`` is no longer an available method for dictionaries in
python3 and ``items()`` returns an iterator by default, it would be preferable
to use ``items()`` directly.

This was something commented briefly over #4411.


richtja (approved):
LGTM, thanks

willianrampazzo (approved):

PR #4485, opened 2021-03-26
interfaces.py: add function to check if interface is present

cforno12:
interface_present is a function that returns True or False to whether or not
the interface name is present or not.

Signed-off-by: Cristobal Forno <cforno12@linux.ibm.com>


avocado/utils/network/interfaces.py
```diff
@@ -380,3 +380,15 @@ def restore_from_backup(self):
...
+        """
+        Check if interface is present
+        rtype: bool
+        """
+        cmd = 'ip link show dev {}'.format(self.name)
+        try:
+            run_command(cmd, self.host, sudo=True)
+            return True
+        except Exception as ex:
+            return False
```
willianrampazzo, line 394: I wonder if there is a need to return `True/False` in this method as it is just checking if the interface is available. See the following method as an example:

https://github.com/avocado-framework/avocado/blob/ca0bacdbb4d4a52ae2a36b312d4bac691d5740f5/avocado/utils/network/interfaces.py#L242

avocado/utils/network/interfaces.py
```diff
@@ -380,3 +380,15 @@ def restore_from_backup(self):
         else:
             raise NWException(
                 "Backup file not available, could not restore file.")
+
+    def interface_present(self):
```
willianrampazzo, line 384: To follow other methods patterns, I would go with `is_interface_available` or `is_interface_present`.

willianrampazzo: Makes sense!

avocado/utils/network/interfaces.py
willianrampazzo: Okay, I had in mind that we prefer to raise an exception instead of return True/False to be more OO, but I recall it from an old discussion. I don't know to which extent it is still true.

beraldoleal: Your understanding is right, We should always try to use exceptions for upper layers. But in this case, the method name is a question, and users are expecting true/false.

willianrampazzo: Okay, got it!

cforno12: Do you want me to include `ping_check` to return True/False in this PR?

beraldoleal: @cforno12, I guess this could be in another PR since will break the API, we need to make sure will not break user cases.

beraldoleal:
Hi @cforno12, we are going to have a new release this week. Do you have intentions to work on this in time for the next release?

cforno12:
Apologies for the delay. I've made the requested changes.

willianrampazzo:
> Apologies for the delay. I've made the requested changes.

This looks good to me. Could you squash both commits?

cforno12:
Done.

avocado/utils/network/interfaces.py
```diff
@@ -380,3 +380,19 @@ def restore_from_backup(self):
...
+    def is_available(self):
+        """Check if interface is available.
+
+        This method checks if the interface is available.
+
+        rtype: bool
+        """
+        cmd = 'ip link show dev {}'.format(self.name)
+        try:
+            run_command(cmd, self.host, sudo=True)
```
beraldoleal, line 393: afaict, you don't need sudo to run a `ip l show`, right?

avocado/utils/network/interfaces.py
```diff
@@ -380,3 +380,19 @@ def restore_from_backup(self):
...
+            return True
+        except Exception as ex:
```
beraldoleal, line 395: `ex` variable is not being used. My suggestion is to add this to `msg`.

beraldoleal (changes requested):
Hi @cforno12, this is almost good to me. Just two small comments for you.

avocado/utils/network/interfaces.py
cforno12: Correct, I've removed it.

avocado/utils/network/interfaces.py
cforno12: Took your suggestion :)

beraldoleal (approved):
Hi @cforno12 , thank you for the changes. LGTM.

PR #4494, opened 2021-03-31
[NRunner] remove the structure of spawner.check_task_requirements

willianrampazzo:
The original code made to check a task requirement is mapped in a way that if requirements are defined in a test, those listed in the check_task_requirements method should be in the list of requirements of a test.

The mapping should work in the reversed way. If a requirement is defined in the test, NRunner should be able to check if each of the requirements are fulfilled and do not expect extra requirements, if they are not set. This operation should happen in the Requirements Resolver when it is implemented.

The following example of the test illustrates the current workflow:

```python
from avocado import Test


class PassTest(Test):

    """
    Example test that passes.

    :avocado: requirement={"type": "file", "name": "foo.bar"}
    """

    def test(self):
        """
        """
```

The list of requirements is triggered by:

https://github.com/avocado-framework/avocado/blob/333015a04362d95682c011bace6f6aded56103e4/avocado/plugins/spawners/process.py#L49

But as the type `core` and value `avocado` does not exist in the list of requirements, this method returns `False`, failing the test execution.

As I mentioned in the beginning, the current mapping forces those requirements listed into the method to be defined in the test's list of requirements. IMO, the requirements should be handled case by case on a specialized architecture (aka, the Requirements Resolver).

### DIFF
--- a/avocado/core/plugin_interfaces.py
+++ b/avocado/core/plugin_interfaces.py
@@ -288,12 +288,3 @@ class Spawner(Plugin):
         :param runtime_task: wrapper for a Task with additional runtime information
         :type runtime_task: :class:`avocado.core.task.runtime.RuntimeTask`
         """
-
-    @staticmethod
-    @abc.abstractmethod
-    async def check_task_requirements(runtime_task):
-        """Checks if the requirements described within a task are available.
-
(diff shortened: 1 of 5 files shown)


beraldoleal:
@willianrampazzo so, could this be closed now?

PR #4503, opened 2021-04-06
selftests: added new test to the assets plugin

### LINKED ISSUE #4460: add test "by days" to the new asset purge and list methods
Today we have only tests about by size. We need to introduce tests by days too.

beraldoleal:
This is just adding a test to ensure purge by days is working.
This fixes #4460.


clebergnu (approved):
LGTM, thanks.

PS: Travis is not allowing the one failed job to be restarted, but all the other jobs passed, as did my local testing.

PR #4529, opened 2021-04-14
Introduce direct inheritance with abstract classes

tiagohonorato:
As mentioned by @pevogam over #4476, using ``abc.ABC`` is much simpler
and direct when compared with ``metaclass=ABCMeta``. 

Since python 3.4, the abc module contains a helper class called ``ABC`` that
will set the metaclass internally. This makes creating an abstract class
more readable and simple as it is only necessary to inherit from ``ABC``.
Avocado, currently, supports python 3.5+ so version should not be a
problem.

### DIFF
--- a/avocado/core/nrunner.py
+++ b/avocado/core/nrunner.py
@@ -313,7 +313,7 @@ class Runnable:
         raise ValueError('Unsupported kind of runnable: %s' % self.kind)
 
 
-class BaseRunner(metaclass=abc.ABCMeta):
+class BaseRunner(abc.ABC):
     """
     Base interface for a Runner
     """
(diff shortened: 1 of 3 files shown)


richtja (approved):
LGTM, thanks

pevogam (approved):
LGTM and thanks for walking the extra mile for other instances of this problem throughout Avocado!

willianrampazzo (approved):

beraldoleal (approved):
LGTM, thanks.

clebergnu (approved):
LGTM, thanks a lot @tiagohonorato !

PR #4530, opened 2021-04-14
 Log information when running tests

richtja:
Adds the test start and finish information in the `job.log`. So, if a
test "1-foo" passes, that should be easy to spot in the job.log file.

Reference: #4366

### DIFF
--- a/avocado/plugins/testlogs.py
+++ b/avocado/plugins/testlogs.py
@@ -7,7 +7,7 @@ from avocado.core.settings import settings
 from avocado.core.teststatus import STATUSES
 
 
-class TestLogsInit(Init):
+class TestLogsUIInit(Init):
 
     description = "Initialize testlogs plugin settings"
 
@@ -31,7 +31,7 @@ class TestLogsInit(Init):
                                  help_msg=help_msg)
 
 
-class TestLogs(JobPre, JobPost):
+class TestLogsUI(JobPre, JobPost):
 
     description = "Shows content from tests' logs"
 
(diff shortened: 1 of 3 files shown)


avocado/plugins/testlogs.py
```diff
@@ -63,3 +63,39 @@ def post(self, job):
...
+        self.runner = config.get('run.test_runner')
+
+    @staticmethod
+    def _get_name(state):
+        if "name" in state:
+            name = state["name"]
+            name = name.name + name.str_variant
+        else:
+            name = "<unknown>"
+        return name
```
beraldoleal, line 82: What about?
```python
name = state.get('name')
if name is None:
    return "<unknown>"
return name.name + name.str_variant
```

avocado/plugins/testlogs.py
```diff
@@ -63,3 +63,39 @@ def post(self, job):
                 except (FileNotFoundError, PermissionError) as error:
                     LOG_UI.error('Failure to access log file "%s": %s',
                                  path, error)
+
+
+class TestLogging(ResultEvents):
+
+    description = "Test logs for Job"
```
beraldoleal, line 70: Maybe be nice to document here and/or any other place that this is specific for nrunner.

beraldoleal (changes requested):
Hi @richtja , thanks for this. LGTM, just a few small comments.

richtja:
Hi @beraldoleal  thanks for your review, I processed your comments and force pushed the changes. I also appended the tests.

beraldoleal (approved):
LGTM, Thank you.

PR #4533, opened 2021-04-15
dmesg.py: Add a decorator for test to fail on

harish-24:
Create a class decorator that wraps the test* methods of a class with a method that raises avocado core Failure exception based on DmesgError after checking for a failure at given dmesg level

### DIFF
--- a/avocado/utils/dmesg.py
+++ b/avocado/utils/dmesg.py
@@ -20,14 +20,54 @@ import logging
 import os
 import tempfile
 import time
+from functools import wraps
 
 from . import genio, process
 
 LOGGER = logging.getLogger('avocado.test')
 
 
+class TestFail(AssertionError, Exception):
+    """
+    Indicates that the test failed.
+
+    This is here, just because of an impossible circular import.
+    """
+    status = "FAIL"
+
+
 class DmesgError(Exception):
     """Base Exception Class for all dmesg  utils exceptions."""
+
+
+def fail_on_dmesg(level=5):
+    """Dmesg fail method decorator
+
+    Returns a class decorator used to signal the test when DmesgError
+    exception is raised.
+
+    :param level: Dmesg Level based on which the test failure should be raised
+    :type level: int
+    :return: Class decorator
+    :rtype: class
+    """
+    def dmesg_fail(cls):
+        def raise_dmesg_fail(func):
+            @wraps(func)
+            def wrapper(*args, **kwargs):
+                try:
+                    data = func(*args, **kwargs)
+                    collect_errors_by_level(level_check=level)
+                    return data
+                except DmesgError as details:
+                    raise TestFail(repr(details)) from details
+            return wrapper
+
+        for key, value in list(vars(cls).items()):
+            if callable(value) and value.__name__.startswith('test'):
+                setattr(cls, key, raise_dmesg_fail(value))
+        return cls
+    return dmesg_fail
 
 
 def clear_dmesg():


avocado/utils/dmesg.py
```diff
@@ -20,8 +20,10 @@
 import os
 import tempfile
 import time
+from functools import wraps
 
 from . import genio, process
+from ..core import exceptions as core_exceptions
```
beraldoleal, line 26: And here we might have a problem. AFAICT, we are not allowed to import avocado core from utils libraries. This might create a circular reference if someone at the core is importing this.

(If is allowed, my suggestions is to remove the relative import)

@clebergnu can you please, confirm this?

beraldoleal (changes requested):

avocado/utils/dmesg.py
willianrampazzo: `selftests/modules-boundaries.sh` is supposed to catch that.

harish-24: @beraldoleal what would be the alternative if we are not allowed to import? How could we invoke Failure from here.

beraldoleal: @clebergnu do you have any ideas ideas ? IMO we should not add this to the `core/decorators.py` file. But I'm not sure how flexible we can be here.

harish-24: @clebergnu any update??

harish-24: @clebergnu Any update on this?

beraldoleal: Hi @harish-24, sorry if this is taking so long.

So, today I was doing a cleanup on the PRs that we have open, and I was investigating this, maybe you could avoid the need to import the core with this small patch:


```diff
diff --git a/avocado/utils/dmesg.py b/avocado/utils/dmesg.py
index 0ef6fd59..f7a6cf87 100644
--- a/avocado/utils/dmesg.py
+++ b/avocado/utils/dmesg.py
@@ -23,7 +23,6 @@ import time
 from functools import wraps
 
 from . import genio, process
-from ..core import exceptions as core_exceptions
 
 LOGGER = logging.getLogger('avocado.test')
 
@@ -44,7 +43,6 @@ def fail_on_dmesg(level=5):
     :rtype: class
     """
     def dmesg_fail(cls):
-        signal = core_exceptions.TestFail
         def raise_dmesg_fail(func):
             @wraps(func)
             def wrapper(*args, **kwargs):
@@ -53,7 +51,7 @@ def fail_on_dmesg(level=5):
                     collect_errors_by_level(level_check=level)
                     return data
                 except DmesgError as details:
-                    raise signal(repr(details)) from details
+                    return cls.fail(details)
             return wrapper
 
         for key, value in list(vars(cls).items()):

```

What do you think?

harish-24: @beraldoleal Thanks for suggesting this. Although it does work, I am seeing a difference running this

With cls.fail:
```
 (1/1) memory/ndctl.py:NdctlTest.test_write_infoblock_size_unaligned:  FAIL: None (3.32 s)
```
With core_expections:
```
 (1/1) memory/ndctl.py:NdctlTest.test_write_infoblock_size_unaligned:  FAIL: DmesgError('Test is failed Found failures in dmesg Please check  dmesg log in debug log.',) (3.27 s)
```


The `cls.fail` does not seem to be able to represent the failure in anyway. Tried with `cls.fail(details)` and `cls.fail(repr
(details)) from details` as well. Please help me with this.

beraldoleal: Hi @harish-24 , luckily `TestFail` is just a normal exception that, for now, we can redefine it here:

```diff
diff --git a/avocado/utils/dmesg.py b/avocado/utils/dmesg.py
index 0ef6fd59..ad2d8e2c 100644
--- a/avocado/utils/dmesg.py
+++ b/avocado/utils/dmesg.py
@@ -23,11 +23,19 @@ import time
 from functools import wraps
 
 from . import genio, process
-from ..core import exceptions as core_exceptions
 
 LOGGER = logging.getLogger('avocado.test')
 
 
+class TestFail(AssertionError, Exception):
+    """
+    Indicates that the test failed.
+
+    This is here, just because of an impossible circular import.
+    """
+    status = "FAIL"
+
+
 class DmesgError(Exception):
     """Base Exception Class for all dmesg  utils exceptions."""
 
@@ -44,7 +52,6 @@ def fail_on_dmesg(level=5):
     :rtype: class
     """
     def dmesg_fail(cls):
-        signal = core_exceptions.TestFail
         def raise_dmesg_fail(func):
             @wraps(func)
             def wrapper(*args, **kwargs):
@@ -53,7 +60,7 @@ def fail_on_dmesg(level=5):
                     collect_errors_by_level(level_check=level)
                     return data
                 except DmesgError as details:
-                    raise signal(repr(details)) from details
+                    raise TestFail(repr(details)) from details
             return wrapper
 
         for key, value in list(vars(cls).items()):

```

harish-24: @beraldoleal Thank you for the suggestions. Pushed with the suggested changes.

beraldoleal (approved):
LGTM. Thank for waiting so long.

willianrampazzo (approved):

Issue #4437, opened 2021-03-04
lint doesn't check every file


richtja:
Our lint check script doesn't check every python file inside the avocado project. This problem occurs especially in examples and selftests directories. It's happening because pylint assumes that any directory is a package, and fails if not. [[1]](https://github.com/PyCQA/pylint/issues/352) So when directory is not a package all files and subdirectories are not checked by lint.

tiagohonorato:
A possible temporary solution (while pylint doesn't provide a solution to lint directories that are not packages) would be to use something like``git ls-files '*.py'`` to list all tracked python files and use it as input to pylint.

PR #4534, opened 2021-04-15
Lint check for every file

richtja:
The pylint didn't check every file python in avocado project. It's
happening because pylint assumes that any directory is a package, and
fails if not. This is fixing that and also fix the lint errors which was found by this change.

Reference: #4437


selftests/.data/loader_instrumented/imports.py
```diff
@@ -1,4 +1,4 @@
-# docstrings, methods, imports... pylint: disable=C0111,W0403,R0903,C0411,E0611,C0412
+# docstrings, methods, imports... pylint: disable=C0111,R0903,C0411,E0611,C0412
```
beraldoleal, line 1: Here you are removing one warning but the commit message says "Lint E0012 fix". Could, please, elaborate a little bit ?

selftests/lint.sh
```diff
@@ -9,22 +9,22 @@ PYLINT=$(which pylint-3 2>/dev/null || which pylint)
 
 # Those are our main packages, we should follow the .pylintrc file with all
 # enabled by default. Some are disabled, we are working to reduce this list.
```
beraldoleal, line 11: We should update this comment too. replacing  the "packages" word.

beraldoleal (changes requested):
Hi @richtja, well done. Thank you. Besides minor changes, this looks fine.

willianrampazzo (approved):
Including @beraldoleal's comments. LGTM.

willianrampazzo:
Codecov needs to be fixed: `ImportError: cannot import name 'test_suite' from 'selftests' (/home/travis/build/avocado-framework/avocado/selftests/__init__.py)`

selftests/.data/loader_instrumented/imports.py
richtja: Yes the problem is that there are no imports so disabling the `W0403` doesn't make sense. And the `E0012` is complaining about it. `E0012 checks if a bad value for a PyLint option is encountered.` I understand the confusion and I will update the commit message.

selftests/lint.sh
richtja: I missed that, thanks.

richtja:
Hi @beraldoleal and @willianrampazzo, thanks for your comments, I force-pushed the fixes.

beraldoleal (approved):
LGTM

PR #4536, opened 2021-04-16
Instrumented nrunner error handling

richtja:
When the exception is raised on the side of the instrumented runner not
inside the test itself, the runner will crash and don't send any message
to the avocado and such an event is not logged into the error log. This
PR handles such event and logs more information about the occurred
exception.


willianrampazzo (approved):

beraldoleal (approved):
LGTM

PR #4549, opened 2021-04-27
better output message and code when using ignore-missing-references

### LINKED ISSUE #4527: Avocado crash without output
When I run nrunner with the wrong test path avocado crashes without any information. It can be reproduced by:

`avocado run --test-runner=nrunner --ignore-missing-references -- test_utils_partition.py`

beraldoleal:
When using ignore-missing-references, Avocado was returning
AVOCADO_JOB_FAIL without explaining the reason. Since we don't have
exactly an job failed here, I'm changing the return code and adding a
warning message. Fixes #4527.

### DIFF
--- a/avocado/plugins/run.py
+++ b/avocado/plugins/run.py
@@ -296,7 +296,11 @@ class Run(CLICmd):
         try:
             suite = TestSuite.from_config(config, name='')
             if suite.size == 0:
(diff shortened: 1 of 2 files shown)


willianrampazzo (approved):

richtja (approved):
LGTM, thanks

PR #4553, opened 2021-04-28
datadir.paths.cache_dirs now is defined from settings registration

beraldoleal:
The current code is creating a custom logic adding a default value that is hardcoded. data_dir module was created way before our new settings logic was introduced, default values now are defined at registration time.  Also, with the current logic "avocado config --datadirs" and "avocado config" output are not consistent (when looking to the cache_dirs).

On top of that, when users define a custom value, in some cases Avocado is replacing this value with a default hardcoded value, instead of respecting users' choice (for instance when users have no permissions).

My proposal here is to use `~/avocado/data/cache` as the default cache dir since usually, a cache dir is useful regardless of the method of installation. This is part of #4443, and if approved I will work on the remaining datadir.paths.*
    
### DIFF
--- a/avocado/core/__init__.py
+++ b/avocado/core/__init__.py
@@ -82,11 +82,15 @@ def register_core_options():
                          help_msg=help_msg,
                          default=60)
 
+    # Let's assume that by default, cache will be located under the user's
+    # umbrella. This will make it easy for our deployments and it is a common
+    # place for other applications too.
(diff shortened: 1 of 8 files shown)


beraldoleal:
@clebergnu, can you please take a look here?

clebergnu (approved):
This makes things much simpler and more predictable.  Thanks a lot @beraldoleal !

PR #4554, opened 2021-04-29
Task identifier as TestID [v2]

richtja:
Some `ResultEvents` plugins expecting that when task is a test that the
even the default identifier of Task is not `TestID` but it is a string.
This commit adds the ability to convert `Task.identifier` to `TestID`
before is used by `ResultEvents` plugins.

---
Changes from v1 (#4550)
* The convertor moved from the Task to Messages.

### DIFF
--- a/avocado/core/messages.py
+++ b/avocado/core/messages.py
@@ -14,6 +14,23 @@
 
 import os
 
+from .test_id import TestID
+
+
+def _convert_identifier_to_test_id(identifier):
+    """
+    It converts Task identifier to TestID
+
+    :param identifier: Task identifier
+    :type: :class:`avocado.core.nrunner.Task.identifier`
+    :return: TestID
+    :rtype :class:`avocado.core.test_id.TestID
+    """
+    if type(identifier) is TestID:
+        return identifier
+    else:
+        return TestID(str(identifier), "task")
+
 
 class BaseMessageHandler:
     """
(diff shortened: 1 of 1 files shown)


avocado/core/messages.py
```diff
@@ -14,6 +14,23 @@
...
 
+from .test_id import TestID
+
+
+def _convert_identifier_to_test_id(identifier):
+    """
+    It converts Task identifier to TestID
+
+    :param identifier: Task identifier
+    :type: :class:`avocado.core.nrunner.Task.identifier`
```
beraldoleal, line 25: nitpicking: Today, task.identifier has not "type" defined. Can be TestID or string. So one possibility, is that this type could be pointing directly to the end types.

avocado/core/messages.py
```diff
@@ -14,6 +14,23 @@
...
+    :return: TestID
+    :rtype :class:`avocado.core.test_id.TestID
+    """
+    if type(identifier) is TestID:
+        return identifier
+    else:
+        return TestID(str(identifier), "task")
+
```
beraldoleal, line 33: Another suggestion is to move this method to the TestID class as a classmethod, something like:

```python
@classmethod
def from_identifier(cls, identifier):
    if type(identifier) is TestID:
        return identifier
    else:
        return cls(str(identifier), "task")
```

In that way, the responsibility of creating a TestID object would be of its own class:

```python
identifier = TestID.from_identifier('foo')
```

beraldoleal (commented):
Hi, @richtja, just left two small comments for you.

avocado/core/messages.py
richtja: Yes I wasn't sure with this. I wanted to point out that it is related to the Task.

avocado/core/messages.py
richtja: I am not sure with this, because I see this change related to the problem of incompatibility of the `Task.identifier` with  `ResultEvents` plugins caused by undefined  type of `Task.identifier`. It is strange to me that `TestID` will be solving the problem of undefined  type of `Task.identifier`. Maybe when the legacy runner will be deprecated it will make more sense or we will be able to remove `TestID` itself. But for now it looks strange to me.

beraldoleal: Well, I understand your point and I'm ok without this change. So, please, feel free to ignore my suggestion.

But IMO makes sense to delegate to TestID most of the things (if not anything) related to TestID, this could be related to the "cohesion of the class or module". One way that I like to think is that if TestID is removed in the future, we have to remove methods/functions from the core besides the objects calls.

It is a common practice to delegate to the class any related helpers and conversion methods that is related to that class/object. We could delegate to that class when it comes: "Please, give me one object of your class, with this information", regardless if this is "information" is a "task.identifier" or something else, TestID should be able to do the conversion.

richtja: You convinced me, thanks for the explanation. It will be easier for future refactoring with legacy runner removal. I will create a new version.

PR #4555, opened 2021-04-29
Task identifier as TestID [v3]

richtja:
Some `ResultEvents` plugins expecting that when task is a test that the
even the default identifier of Task is not `TestID` but it is a string.
This PR adds the ability to convert `Task.identifier` to `TestID`
before is used by `ResultEvents` plugins.

---
Changes from v1 (#4550)
* The convertor moved from the Task to Messages.

Changes from v2 (#4554)
* The convertor moved from the Messages to TestID.

### DIFF
--- a/avocado/core/messages.py
+++ b/avocado/core/messages.py
@@ -14,6 +14,8 @@
 
 import os
 
+from .test_id import TestID
+
 
 class BaseMessageHandler:
     """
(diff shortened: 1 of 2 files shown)


beraldoleal (approved):
This LGTM, thank you @richtja.

willianrampazzo (approved):

clebergnu (approved):
LGTM, thanks!

pevogam:
Hi @richtja, while this PR solves one issue I am still confused about the general ambiguity here. What should be the preferred approach for other plugins? Should they rely and specify TestID-s or strings? Is the plan to deprecate TestID and use strings (considering the direction here) or vice versa?

PR #4562, opened 2021-05-04
plugins.list: Adding a machine-readable output for 'avocado list'

### LINKED ISSUE #4538: Add machine-readeable output to `avocado list`
So that scripts could more easily use its output to implement complex workflows. Examples of workflows can be found at https://github.com/avocado-framework/avocado/issues/4430

Currently `avocado list` prints to stdout as human-readable text. If you use options like `--verbose` it will print detailed information about the tests, however, this isn't easily consumed as one would need to implement a parser. Thus, an ouput on json format would be very welcomed.

beraldoleal:
Introducing the --write-to-json option to the 'list' command.
Fixes #4538.

### DIFF
--- a/avocado/plugins/list.py
+++ b/avocado/plugins/list.py
@@ -14,6 +14,7 @@
 # Author: Beraldo Leal <bleal@redhat.com>
 
 import os
+import json
 
 from avocado.core import exit_codes, loader, parser_common_args
 from avocado.core.output import LOG_UI, TERM_SUPPORT
@@ -176,6 +177,21 @@ class List(CLICmd):
         return test_matrix
 
     @staticmethod
+    def _save_to_json(matrix, filename, verbose=False):
+        result = []
+        type_label_mapping = loader.loader.get_type_label_mapping()
+        for line in matrix:
+            if verbose:
+                result.append({'Type': type_label_mapping[line[0]],
+                               'Test': line[1],
+                               'Tags': line[2]})
+            else:
+                result.append({'Type': type_label_mapping[line[0]],
+                               'Test': line[1]})
+        with open(filename, 'w') as fp:
+            json.dump(result, fp, indent=4)
+
+    @staticmethod
     def save_recipes(suite, directory, matrix_len):
         fmt = '%%0%uu.json' % len(str(matrix_len))
         index = 1
@@ -230,9 +246,20 @@ class List(CLICmd):
                                  parser=parser,
                                  long_arg='--write-recipes-to-directory')
 
+        help_msg = 'Writes output to a json file.'
+        settings.register_option(section='list',
+                                 key='write_to_json_file',
+                                 default=None,
+                                 metavar='JSON_FILE',
+                                 help_msg=help_msg,
+                                 parser=parser,
+                                 long_arg='--write-to-json')
+
         parser_common_args.add_tag_filter_args(parser)
 
     def run(self, config):
+        verbose = config.get('core.verbose')
+        write_to_json_file = config.get('list.write_to_json_file')
         resolver = config.get('list.resolver')
         runner = 'nrunner' if resolver else 'runner'
         config['run.references'] = config.get('list.references')
@@ -257,6 +284,8 @@ class List(CLICmd):
             else:
                 matrix = self._get_test_matrix(suite)
                 self._display(suite, matrix)
+            if write_to_json_file:
+                self._save_to_json(matrix, write_to_json_file, verbose)
         except KeyboardInterrupt:
             LOG_UI.error('Command interrupted by user...')
             return exit_codes.AVOCADO_FAIL
--- a/avocado/plugins/list.py
+++ b/avocado/plugins/list.py
@@ -47,7 +47,32 @@ class List(CLICmd):
     description = 'List available tests'
 
     @staticmethod
-    def _display(suite, matrix):
+    def _apply_colors_on_test_type(matrix, verbose=False):
+        try:
+            type_label_mapping = loader.loader.get_type_label_mapping()
+            decorator_mapping = loader.loader.get_decorator_mapping()
+        except RuntimeError:
+            # When running with --resolver this will fall here, and should use
+            # the default mapping
+            pass
+
+        colored_matrix = []
+        for item in matrix:
+            cls = item[0]
+            try:
+                decorator = decorator_mapping[cls]
+                type_label = decorator(type_label_mapping[cls])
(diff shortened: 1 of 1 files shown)


avocado/plugins/list.py
```diff
@@ -143,11 +171,26 @@ def _get_resolution_matrix(suite):
...
+        type_label_mapping = loader.loader.get_type_label_mapping()
+        for line in matrix:
+            if verbose:
+                result.append({'Type': type_label_mapping[line[0]],
+                               'Test': line[1],
+                               'Tags': line[2]})
+            else:
+                result.append({'Type': type_label_mapping[line[0]],
+                               'Test': line[1]})
+        with open(filename, 'a') as fp:
```
richtja, line 191: I wouldn't use append mode here. I think that the json should be related to the updated list results and shouldn't be mixed with other results.

avocado/plugins/list.py
```diff
@@ -230,9 +246,20 @@ def configure(self, parser):
...
                                  long_arg='--write-recipes-to-directory')
 
+        help_msg = 'Writes output to a json file.'
+        settings.register_option(section='list',
+                                 key='write_to_json_file',
+                                 default=None,
+                                 metavar='JSON_FILE',
+                                 help_msg=help_msg,
+                                 parser=parser,
+                                 long_arg='--write-to-json')
```
willianrampazzo, line 256: we already have `--json FILE` arguments across the code. I would go with that to keep the pattern.

richtja (commented):
Hi @beraldoleal, thanks for this LGTM, I have just a small comment here, but it's not necessary to change that.

avocado/plugins/list.py
beraldoleal: Makes sense.

avocado/plugins/list.py
beraldoleal: `--json` it was introduced by the `plugins/jsonresult.py` and it is about `job.run.results`. Assuming that plugins should be independent, IMO we should choose different command line options names to avoid confusion.  IDK, but having `--json` that generates one output with the `run` and a similar `--json` that provides a different output with `list` commands, seems strange to me. Also `--json` is too generic, IMO.

What do you think?

beraldoleal:
> Hi @beraldoleal, thanks for this LGTM, I have just a small comment here, but it's not necessary to change that.

Thank you @richtja , Since it was a small change, I just force-pushed here.

avocado/plugins/list.py
willianrampazzo: I like consistency. Usually when I use a command and I want to check if there is some kind of JSON output I look for the `-j` or `--json` argument, just like in the `ip` command. Having two different options in the same software to write the output to JSON would make me really mad. I know the plugins are different, but they will do the same when using the argument, write to a JSON file.

wainersm:
@beraldoleal first of all, thanks for taking yet another of my requirements to work on :)

I'm testing it right now. So stuffs I noted so far:

 1) Passing `--write-to-json` it properly write the json file AND keep printing to stdout. Maybe the output to stdout should be omitted?
 2) Suppose `--write-to-json tests.json`. If `tests.json` exists then it will append to that file. @richtja noticed that on his review;
 3) Suppose `avocado --verbose list tests/acceptance/ --write-to-json tests.json`
 3.1) The *test types summary* and *test tags summary* aren't wrote to json. I don't have a strong opinion if they should.
 3.2) The *Tags* field is a string. I think it should be a hash field just like in `job-results/latest/results.json`
Currently:
  ```
     {
        "Type": "INSTRUMENTED",
        "Test": "tests/acceptance/boot_linux.py:BootLinuxX8664.test_pc_i440fx_tcg",
        "Tags": "machine(pc),accel(tcg),arch(x86_64)"
    }
 ```
 In `job-results/latest/results.json` it would be:
 ```
            "tags": {
                "machine": ["pc"],
                "accel": ["tcg"],
                "arch": ["x86_64"]
            },
 ```

avocado/plugins/list.py
beraldoleal: Ok, I don't like much, but I understand your point. I will change on a v2.

beraldoleal:
> 1. Passing --write-to-json it properly write the json file AND keep printing to stdout. Maybe the 2. output to stdout should be omitted?

I left there because is usefull and users can remove with `> /dev/null`. Also it is the same behaviour as `run --json`

> 2. Suppose --write-to-json tests.json. If tests.json exists then it will append to that file. @richtja noticed that on his review;

I fixed that, is in `w` mode. No append to avoid mess inside the file.

> 3. Suppose avocado --verbose list tests/acceptance/ --write-to-json tests.json
> 3.1) The test types summary and test tags summary aren't wrote to json. I don't have a strong opinion if they should.

I'm omitting this intentionally. Unless it is a strong requirement, we can add this in future PRs.

> 3.2) The Tags field is a string. I think it should be a hash field just like in job-results/latest/results.json

Yes, makes sense. I will do like we did a few PRs ago.

avocado/plugins/list.py
wainersm: While `--write-to-json` keeps it similar to the existing `--write-recipes-to-directory`, I do think that `--json` is more broadly used on tools. So I agree with @willianrampazzo

avocado/plugins/list.py
wainersm: 1+

wainersm:
> > 1. Passing --write-to-json it properly write the json file AND keep printing to stdout. Maybe the 2. output to stdout should be omitted?
> 
> I left there because is usefull and users can remove with `> /dev/null`. Also it is the same behaviour as `run --json`
> 

So it is consistent with `run --json`. Cool.

> > 1. Suppose --write-to-json tests.json. If tests.json exists then it will append to that file. @richtja noticed that on his review;
> 
> I fixed that, is in `w` mode. No append to avoid mess inside the file.
> 
> > 1. Suppose avocado --verbose list tests/acceptance/ --write-to-json tests.json
> >    3.1) The test types summary and test tags summary aren't wrote to json. I don't have a strong opinion if they should.
> 
> I'm omitting this intentionally. Unless it is a strong requirement, we can add this in future PRs.

And make senses to keep it simple. User tools can generate those data from the json anyway.
 
> 
> > 3.2) The Tags field is a string. I think it should be a hash field just like in job-results/latest/results.json
> 
> Yes, makes sense. I will do like we did a few PRs ago.

beraldoleal:
Closing this in favor of a v2.

PR #4577, opened 2021-05-11
removing task-state-machine.py from examples

beraldoleal:
IIUC, this is just a small simulation for explaining how a state machine
should work. IMO examples/ folder should have only code that could be
used by users and contributors for reuse.


clebergnu (approved):
Fair enough, it's in the wrong place and we don't need it anymore! Thanks!

willianrampazzo (approved):

clebergnu (approved):
LGTM, thanks!

PR #4595, opened 2021-05-15
ssh.py: disable pseudoterminal by default.

cforno12:
Ran into this issue when trying to connect to an AIX partition. The ControlPath was not created since ssh stated this: "Pseudo-terminal will not be allocated because stdin is not a terminal.". Passing in the '-T' flag supressed this message and the ControlPath was created.

Signed-off-by: Cristobal Forno <cforno12@linux.ibm.com>

### DIFF
--- a/avocado/utils/ssh.py
+++ b/avocado/utils/ssh.py
@@ -105,7 +105,7 @@ class Session:
         else:
             options += (('PasswordAuthentication', 'yes'),
                         ('NumberOfPasswordPrompts', '1'),)
-        return self._ssh_cmd(options, ('-n',))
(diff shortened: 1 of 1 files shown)


willianrampazzo (approved):
I don't think we expect this utility to have any interaction with a terminal. In this case, LGTM.

clebergnu (approved):
Seems a fair assumption to me.  Tested using the previous usual scenarios (connecting to a QEMU based VM within a test) and saw no side effects.

Thanks!

beraldoleal (approved):
LGTM

Issue #4575, opened 2021-05-11
The --dry-run is not working with nrunner


richtja:
You can reproduce it with:

```
avocado run --test-runner=nrunner  --dry-run /bin/true
JOB ID     : 0000000000000000000000000000000000000000
JOB LOG    : /tmp/avocado-dry-run-6u51u5lh/job-2021-05-11T11.48-0000000/job.log
 (1/1) /bin/true: STARTED
 (1/1) /bin/true: PASS (0.01 s)
RESULTS    : PASS 1 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB HTML   : /tmp/avocado-dry-run-6u51u5lh/job-2021-05-11T11.48-0000000/results.html
JOB TIME   : 1.17 s

```

PR #4598, opened 2021-05-17
Dry-run for nrunner

richtja:
This commit introduces the `--dry-run` option for nrunner. When you will
use `--dry-run` it will change every runnable kind to `dry-run` and the
tests will be run inside `DryRunRunner` which mark them as canceled.

Reference: #4575

### DIFF
--- a/selftests/unit/test_job.py
+++ b/selftests/unit/test_job.py
@@ -337,6 +337,17 @@ class JobTest(unittest.TestCase):
         self.job.run()
         self.assertEqual(len(self.job.get_failed_tests()), 1)
 
+    def test_job_dryrun(self):
+        config = {'run.references': ['/bin/true', '/bin/false'],
+                  'run.results_dir': self.tmpdir.name,
+                  'run.dry_run.enabled': True,
+                  'core.show': ['none']}
+        suite = TestSuite.from_config(config)
+        self.job = job.Job(config, [suite])
+        self.job.setup()
+        self.job.run()
+        self.assertEqual(self.job.result.cancelled, 2)
+
     def tearDown(self):
         data_dir._tmp_tracker.unittest_refresh_dir_tracker()
         self.tmpdir.cleanup()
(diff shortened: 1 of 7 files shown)


selftests/unit/test_job.py
```diff
@@ -329,7 +329,6 @@ def test_job_make_test_suite_resolver(self):
     def test_job_get_failed_tests(self):
```
beraldoleal, line 329: Not sure if I understood why `run.dry_run.enabled` makes no sense here?

IIUC, this test is testing failed test counts, and should not be changed or converted to dry_run. a /bin/true + /bin/false should continue to work and having only 1 failed.

If we need to test the dry-run, this could be in a different test.

beraldoleal (changes requested):

selftests/unit/test_job.py
willianrampazzo: That is what the test is doing, checking how many tests failed. `run.dry_run.enabled` here will result in 2 canceled, not one pass and one canceled.

willianrampazzo (approved):
I find it an interesting approach. The code looks good to me

willianrampazzo:
rpm-build failed:

error: Installed (but unpackaged) file(s) found:
   /usr/bin/avocado-runner-dry-run
    Installed (but unpackaged) file(s) found:
   /usr/bin/avocado-runner-dry-run

selftests/unit/test_job.py
richtja: As @willianrampazzo said, before the 442d069629dc3efbe9ed17aa64b5206badfe025f the `run.dry_run.enabled` wasn't changing the behavior of nrunner and the test was running like a standard test. But now without this commit the job will be run as `dry-run` and cancel all tests. That is why we have to remove `run.dry_run.enabled` or we can change it to `False` but that is the default value.

richtja:
Hi @willianrampazzo and @beraldoleal, thanks for your comments I fixed the rpm-build issue and added some tests.

selftests/unit/test_job.py
beraldoleal: Nevermind, I got it.

beraldoleal (approved):
Thank you @richtja . LGTM.

PR #4602, opened 2021-05-19
Refresh avocado man-page (v2)

ana:
Full refresh and update of the avocado man page.

### DIFF
--- a/man/avocado.rst
+++ b/man/avocado.rst
@@ -13,13 +13,13 @@ DESCRIPTION
 ===========
 
 Avocado is a modern test framework that is built on the experience
-accumulated with `autotest` (`http://autotest.github.io`).
+accumulated with `autotest` (https://autotest.github.io).
 
 `avocado` is also the name of its test runner command line tool,
 described in this man page.
 
 For more information about the Avocado project, please check its
-website: http://avocado-framework.github.io/
+website: https://avocado-framework.github.io/
 
 OPTIONS
 =======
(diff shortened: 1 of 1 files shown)


man/avocado.rst
```diff
@@ -304,79 +306,103 @@ Options for subcommand `diff` (`avocado diff --help`)::
...
+
+    By default, a textual diff report is generated in the standard output.
 
 Options for subcommand `distro` (`avocado distro --help`)::
 
     optional arguments:
       -h, --help            show this help message and exit
       --distro-def-create   Creates a distro definition file based on the path
-                            given
+                            given.
```
willianrampazzo, line 321: This is the only place you added the period. As most places do not have a period, I would follow the pattern and let this line without it.

willianrampazzo (approved):
So, I see some inconsistency related to the use of periods in the options' descriptions. We need to review them, but this can be done in another PR. In this case, LGTM.

ana:
@willianrampazzo agree, we need to check it directly in the code. This update shows exactly the present output.
What do we prefer periods or not? :-)

willianrampazzo:
> @willianrampazzo agree, we need to check it directly in the code. This update shows exactly the present output.

Okay, that makes sense.

> What do we prefer periods or not? :-)

I don't have a strong preference. It looks like using periods makes it grammatically more correct.

beraldoleal (approved):
LGTM, thank you @ana .

PR #4605, opened 2021-05-21
Empty reference in resolver [v3]

richtja:
Some test might be able to be resolved without a reference. Config
files or other parameters from user can be used for that. As an example
is avocado vt which can use Cartesian configuration for resolving tests.
This commit adds the ability to discover tests from different data
without the test references.

---
Changes from v1 (#4574)
Instead of creating `EmptyReference`. This PR introduces `Discoverer` plugin interface.

Changes from v2 (#4591)
* Removed commit about `list --resolver`
* Added example with Magic Tests
* Documentation update

### DIFF
--- a/avocado/core/plugin_interfaces.py
+++ b/avocado/core/plugin_interfaces.py
@@ -234,6 +234,24 @@ class Resolver(Plugin):
         """
 
 
+class Discoverer(Plugin):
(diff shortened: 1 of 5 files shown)


clebergnu (approved):
LGTM, thanks!

PR #4612, opened 2021-05-25
rpm: Get string type of output of installed version to be compared with

chloerh:
Current code got `inst_version` as `byte` type and cannot compare with
a string or error would be throw out. Replace with `process.run()` to
get string type of output to fix the problem.


beraldoleal (approved):
LGTM, thank you.

clebergnu (approved):
LGTM, thanks!

PR #4615, opened 2021-05-26
docs: list all packages required for development

ana:


willianrampazzo (approved):
I don't think Cirrus' failures are related to this PR. LGTM!

setup.py
```diff
@@ -46,7 +46,7 @@ def run(self):
         call(('rm -rf MANIFEST BUILD BUILDROOT SPECS RPMS SRPMS SOURCES',
               'PYPI_UPLOAD ./build ./dist'), shell=True)
         call('rm -rf ./man/avocado.1 ./docs/build', shell=True)
-        call('rm -rf /var/tmp/avocado* /tmp/avaocado*', shell=True)
+        call('rm -rf /var/tmp/avocado* /tmp/avocado*', shell=True)
```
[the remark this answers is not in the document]
beraldoleal: Nice catch! Thank you!

docs/source/guides/contributor/chapters/environment.rst
```diff
@@ -3,14 +3,18 @@ Development environment
...
-.. warning:: TODO: Needs improvment here. i.e: virtualenvs, GPG, etc.
+.. warning:: TODO: Needs improvement here. i.e: virtualenvs, GPG, etc.
 
 Installing dependencies
 -----------------------
 
 You need to install few dependencies before start coding::
 
- $ sudo dnf install gcc libvirt-devel
+ $ sudo dnf install gcc libvirt-devel python-devel python-enchant python-pip pylint
```
beraldoleal, line 13: Ideally this should be a list of system distro dependencies only. `pyenchant` and `pylint` are already on `requirements-selftests.txt`.

docs/source/guides/contributor/chapters/environment.rst
```diff
@@ -3,14 +3,18 @@ Development environment
...
+
+Optionally, if you want to build the manpage install::
+
+ $ sudo dnf install python-docutils
```
beraldoleal, line 17: Maybe we could add this requirement to `requirements-selftests.txt`

beraldoleal (changes requested):
Hi @ana, thank you for this. I left two small comments for you here.

docs/source/guides/contributor/chapters/environment.rst
willianrampazzo: Personally, I use all packages from the system and in the beginning, I always had a hard time trying to find which package was missing.

beraldoleal: I agree that is confusing, but having a few packages here and others in `requirements-selftests.txt` IMO make things more confusing.

In an ideal python package, requirements should be installed automatically as much as possible for automatic tasks, deployments and follow some conventions for new contributors.

Maybe the installing dependencies section should  instruct developers to run `make requirements-selftests` or `pip install -r requirements-selftests.txt`, in that way we will also be less prune to future mistakes here in case we forget to update the docs.

willianrampazzo: > Maybe the installing dependencies section should instruct developers to run `make requirements-selftests` or `pip install -r requirements-selftests.txt`, in that way we will also be less prune to future mistakes here in case we forget to update the docs.

This or all the packages listed here would help. Leaving one or two packages here and no other information does not help much.

ana: I tend to use directly the packages from the system, but this is for the setup of a development environment so it makes sense to point developers to use `pip` with `requirements-selftests.txt`. 
I'll update the list of packages from the system anyway because `python-devel` is required and add a comment about   `pip` with `requirements-selftests.txt`

docs/source/guides/contributor/chapters/environment.rst
ana: pip install the binary as `rst2man.py` and `Makefile` looks for `rst2man`. So this will require a small update in the `Makefile`
to use both.

PR #4630, opened 2021-05-31
utils.vmimage: deprecate vmimage.get() in favor of Image.from_parameters

beraldoleal:
get() method is returning an Image object, let's delegate this to the Image
class.

@clebergnu, I'm aware that we are using this on qemu acceptance tests, if
this is accepted, then we should update there too.


richtja (approved):

clebergnu (approved):
LGTM, thanks.

PR #4673, opened 2021-06-11
 utils.software_manager: introduce extract_from_package() - v3

### LINKED ISSUE #3549: utils/software_management: Add extract_from_deb and extract_from_rpm operations

beraldoleal:
Fixes #3549.

#### Changes from v2:
 - Some typo fixes
 - Using assertIn
 - Remove unreachable code
 - Small code refactor to return directly the condition result
 
#### Changes from v1:
 - Moving tests to functional tests;
 - Using TestCaseTmpDir


clebergnu (approved):
LGTM, thanks!

willianrampazzo (approved):

PR #4679, opened 2021-06-14
setup.py: replace subprocess.call with python

ana:
This is a more pythonic approach to avoid the security implications of using shell=True .
Use remove_tree for removing directories recursively even when they're not empty.
Introduce the module pathlib to import bash globbing.
    
### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,11 +45,12 @@ class Clean(clean):
 
     def run(self):
         super().run()
-        cleaning_list = ["MANIFEST", "BUILD", "BUILDROOT", "SPECS",
-                         "RPMS", "SRPMS", "SOURCES", "PYPI_UPLOAD",
-                         "./build", "./dist",
-                         "./man/avocado.1", "./docs/build",
-                         "/tmp/avocado/"]
+        cleaning_list = [Path("MANIFEST"), Path("BUILD"), Path("BUILDROOT"),
+                         Path("SPECS"), Path("RPMS"), Path("SRPMS"),
+                         Path("SOURCES"), Path("PYPI_UPLOAD"),
+                         Path("./build"), Path("./dist"),
+                         Path("./man/avocado.1"), Path("./docs/build"),
+                         Path("/tmp/avocado/")]
 
         cleaning_list += list(Path('/var/tmp/').glob(".avocado-task*"))
         cleaning_list += list(Path('/var/tmp/').glob("avocado*"))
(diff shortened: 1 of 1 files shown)


ana:
The failure on Debian 9.13  is because the python version is 3.5 and `os.path` operations hasn't been updated to work with PosixPath until Python 3.6.

On the other hand, in `setup.py` we claim:
`          
python_requires='>=3.6',
`

And this should be lowered to 3.5 if we want to keep around Debian 9.

setup.py
```diff
@@ -43,15 +45,27 @@ class Clean(clean):
...
+        cleaning_list += list(Path('.').rglob("*.egg-info"))
+        cleaning_list += list(Path('.').rglob("*.pyc"))
+        cleaning_list += list(Path('.').rglob("__pycache__"))
+        cleaning_list += list(Path('./docs/source/api/').rglob("*.rst"))
+
+        for e in cleaning_list:
+            if os.path.exists(e) and os.path.isfile(e):
+                os.remove(e)
+            if os.path.exists(e) and os.path.isdir(e):
+                remove_tree(e)
```
beraldoleal, line 65: Any reason for using remove_tree instead of `os.removedirs()`? I'm asking this because we are trying to remove distutils, since as pointed on the docs: 

> ... most Python users will not want to use this module directly, but instead use the cross-version tools maintained by the Python Packaging Authority. In particular, setuptools is an enhanced alternative to distutils ...

beraldoleal (commented):
Hi @ana, thank you for this. Very welcome. I have just left one question for you.

setup.py
ana: @beraldoleal  `os.removedirs()` doesn't work if the directory is not empty. However, I can use instead `shutil.rmtree`

beraldoleal:
> The failure on Debian 9.13 is because the python version is 3.5 and `os.path` operations hasn't been updated to work with PosixPath until Python 3.6.
> 
> On the other hand, in `setup.py` we claim:
> ` python_requires='>=3.6',`
> 
> And this should be lowered to 3.5 if we want to keep around Debian 9.

Yes, you are right, and I'm ok with keeping things on >=3.6.  So, if you hit some wall with #4682 , we can alternatively update here with:

```python
os.path.exists(str(e))
```

But, yes, lets try to push things to the newer versions.

ana:
I'll do a v2 with more commits once #4682 is merged.

PR #4681, opened 2021-06-14
Nrunner File message

richtja:
This PR adds a new type of nrunner messages, the `FileMessage`. This message comes with runners ability to create runner specific files and subdirectories in avocado Task directory. This feature can be used when you want to log some runner specific data. It will be used in sysinfo runner.

### DIFF
--- a/avocado/core/messages.py
+++ b/avocado/core/messages.py
@@ -84,12 +84,11 @@ class StartMessageHandler(BaseMessageHandler):
     It will create the test base directories and triggers the 'start_test'
     event.
 
-    This is triggered when the runner starts the test.
+    This have to be triggered when the runner starts the test.
 
-    The started message properties:
-    param status: 'started'
-    param time: start time of the test
-    type time: float
+    :param status: 'started'
+    :param time: start time of the test
+    :type time: float
 
     example: {'status': 'started', 'time': 16444.819830573}
     """
@@ -120,12 +119,14 @@ class FinishMessageHandler(BaseMessageHandler):
 
     This is triggered when the runner ends the test.
 
-    The finished message properties:
-    param status: 'finished'
-    param result: test result
-    type result: `avocado.core.teststatus.STATUSES`
-    param time: end time of the test
-    type time: float
+    :param status: 'finished'
+    :param result: test result
+    :type result: `avocado.core.teststatus.STATUSES`
+    :param time: end time of the test
+    :type time: float
+    :param fail_reason: Optional parameter for brief specification, of the
+                       failed result.
+    :type fail_reason: string
 
     example: {'status': 'finished', 'result': 'pass', 'time': 16444.819830573}
     """
@@ -209,13 +210,12 @@ class LogMessageHandler(BaseRunningMessageHandler):
 
     It will save the log to the debug.log file in the task directory.
 
-    The log message properties:
-    param status: 'running'
-    param type: 'log'
-    param log: log message
-    type log: string
-    param time: Time stamp of the message
-    type time: float
+    :param status: 'running'
+    :param type: 'log'
+    :param log: log message
+    :type log: string
+    :param time: Time stamp of the message
+    :type time: float
 
     example: {'status': 'running', 'type': 'log', 'log': 'log message',
              'time': 18405.55351474}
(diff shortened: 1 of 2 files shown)


avocado/core/messages.py
```diff
@@ -301,3 +302,36 @@ class WhiteboardMessageHandler(BaseRunningMessageHandler):
...
+    :param time: Time stamp of the message
+    :type time: float
+
+    example: {'status': 'running', 'type': 'file', 'path':'foo/runner.log',
+             'log': 'this will be saved inside file',
+             'time': 18405.55351474}
+    """
+
+    def handle(self, message, task, job):
+        filename = os.path.relpath(os.path.join("/", message['path']), "/")
```
willianrampazzo, line 332: If the `message['path']` should be the relative path, like in the example `foo/runner.log`, this line returns the same as `message['path']`. So, what case are you covering here?

avocado/core/messages.py
```diff
@@ -79,17 +79,16 @@ def process_message(self, message, task, job):
 
 class StartMessageHandler(BaseMessageHandler):
     """
-    Handler for started message.
+    Started message.
```
beraldoleal, line 83: If this is not a handler, then for consistency, we should change the class name too. Otherwise I think we should keep the handler here.

avocado/core/messages.py
```diff
@@ -70,7 +70,8 @@ def __init__(self):
         self._handlers = {'log': [LogMessageHandler()],
```
beraldoleal, line 70: The commit says File_Message, but this is actually FileMessage, right?

avocado/core/messages.py
```diff
@@ -301,3 +302,36 @@ class WhiteboardMessageHandler(BaseRunningMessageHandler):
     def handle(self, message, task, job):
         self._save_message_to_file('whiteboard', message['log'], task,
                                    message.get('encoding', None))
+
+
+class FileMessageHandler(BaseRunningMessageHandler):
+    """
+    Handler for file message.
```
beraldoleal, line 309: If others are not handler, is this a handler?

docs/source/guides/contributor/chapters/runners.rst
```diff
@@ -592,89 +592,19 @@ information will be processed by the avocado core.
...
-:type time: float
-:example: {'status': 'running', 'type': 'whiteboard',
-         'log': b'whiteboard message', 'time': 18405.55351474}
+.. autoclass:: avocado.core.messages.LogMessageHandler
+
+.. autoclass:: avocado.core.messages.StdoutMessageHandler
+
+.. autoclass:: avocado.core.messages.StderrMessageHandler
+
+.. autoclass:: avocado.core.messages.WhiteboardMessageHandler
```
beraldoleal, line 610: Cool!

avocado/core/messages.py
richtja: I think that is important to ensure that this files will be under `Task directory`. This covers cases as: 
`../test`
`../../test`
`../../foo/../test`
`/foo/test`
`./foo/test`

avocado/core/messages.py
richtja: Yes, my bad

avocado/core/messages.py
richtja: I wasn't sure with this change. All of them are handlers, but this docstring, will be copied to the message documentation, so I thought that the word `Handler` might be confusing in documentation.

avocado/core/messages.py
richtja: No, it's mistake, All of them have to be same.

avocado/core/messages.py
beraldoleal: I think that is fine to have "Handler" on the documentation if that makes sense.

richtja:
@beraldoleal, thanks for your comments, I did fixes with force-push.

beraldoleal (approved):
LGTM

avocado/core/messages.py
willianrampazzo: Got it, thanks!

willianrampazzo (approved):

Issue #4680, opened 2021-06-14
Cirrus: replace debian 9 with debian 11


ana:
Cirrus is running right now for old debian 9 (now only with LTS support until June 2022) and current debian stable: debian  10. 
Debian 11 is being released next month and it's about time to switch debian 9 for debian 11.

PR #4682, opened 2021-06-14
cirrus: update debian images

ana:
Remove Debian 9 (stretch) and add instead Debian 11 (Bullseye),
that will be released next month.
Update tags to use codenames instead of versions. The codename
always point to the latest release, this saves us to update
the tag every time there is a new point release.

Fixes: #4680

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -33,8 +33,8 @@ debian_version_task:
 
     container:
         matrix:
-          - image: debian:9.13
-          - image: debian:10.8
+          - image: debian:10.9
+          - image: debian:bullseye-20210511
           - image: ubuntu:18.04
           - image: ubuntu:20.04
 


willianrampazzo (approved):

.cirrus.yml
```diff
@@ -33,8 +33,8 @@ debian_version_task:
 
     container:
         matrix:
-          - image: debian:9.13
-          - image: debian:10.8
+          - image: debian:buster
+          - image: debian:bullseye
```
beraldoleal, line 37: Usually we try to pin the versions to avoid false positives errors on CI. This will make CI update more frequently and can cause problems. We can revisit that, and discuss this, of course.

beraldoleal (commented):
Hi @ana I'm ok with updating our debian version on cirrus, just one small consideration about to pin vs not to pin.

.cirrus.yml
ana: I was thinking this kind of comment was a possible answer to this PR while making it. I still went ahead and used the codename because I consider Debian to be extremely conservative in updates for point releases and I'm not expecting a lot of problems.
That said, I'm ok to continue pinning to the versions exactly for the same reasons: point releases are very conservative and even if we don't update often, it won't change a lot the result of the tests.

beraldoleal: I see your point, it is just a matter of deciding what we should follow. Even with a conservative distro, like debian, buster had 9 updates, since mid-2019. I'm ok with that approach too, I just would like to make sure that we are agreeing on this change and giving more visibility to this change.

So, let see with others has to say. @clebergnu, thoughts ?

richtja (approved):

.cirrus.yml
ana: FWIW, I did a force-push using version instead of codename (except for bullseye, that'll be a timestamp until is released)

beraldoleal (approved):

PR #4686, opened 2021-06-15
Extend runnable config options

willianrampazzo:
This PR does 2 two changes to allow passing all the config options to a runnable:

* This handles the nrunner serialization/deserialization process involved in the config. It introduces a mechanism that converts sets into lists inside a dictionary. The dictionary key `__encoded_set__` is used in the decoding process to convert the option back to a set.
* Change the suite to pass all the config options to a runnable instead of filtering.

### DIFF
--- a/avocado/core/nrunner.py
+++ b/avocado/core/nrunner.py
@@ -72,6 +72,35 @@ def check_runnables_runner_requirements(runnables, runners_registry=None):
     return (ok, missing)
 
 
+class ConfigDecoder(json.JSONDecoder):
+    """
+    JSON Decoder for config options.
+    """
+
+    @staticmethod
+    def decode_set(config_dict):
+        for k, v in config_dict.items():
+            if isinstance(v, dict):
+                if '__encoded_set__' in v:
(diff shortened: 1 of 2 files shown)


willianrampazzo:
Is anyone able to reproduce the `inspekt-style` fail from GitHub Actions? I could not reproduce it locally. Also, looking at the code, I don't see a real reason for the failure.

willianrampazzo:
BTW, where is Travis? :)

willianrampazzo:
> Is anyone able to reproduce the `inspekt-style` fail from GitHub Actions? I could not reproduce it locally. Also, looking at the code, I don't see a real reason for the failure.

Problem solved. There was a missing blank line after the class docstring. Interesting that locally it does not fail.

willianrampazzo:
> BTW, where is Travis? :)

Found it. From travis-ci.org: _Since June 15th, 2021, the building on travis-ci.org is ceased. Please use travis-ci.com from now on._

willianrampazzo:
Closing to check if Travis is triggered now.

PR #4687, opened 2021-06-15
Extend runner config options

willianrampazzo:
This PR does 2 two changes to allow passing all the config options to a runnable:

* This handles the nrunner serialization/deserialization process involved in the config. It introduces a mechanism that converts sets into lists inside a dictionary. The dictionary key `__encoded_set__` is used in the decoding process to convert the option back to a set.
* Change the suite to pass all the config options to a runnable instead of filtering.


willianrampazzo:
Trying again :)

PR #4690, opened 2021-06-16
[nrunner] introduce avocado-runner-requirement-asset [v2]

willianrampazzo:
This runner handles assets using the Avocado Asset utility. It can be used as a requirement type on tests that define requirements.

Changes from v1 (#4600):
- Skip the test if `name` is empty and do not process anything else.
- Correctly handle the `cache_dirs` based on the runnable config.


beraldoleal (approved):

clebergnu (approved):
LGTM, thanks!

PR #4693, opened 2021-06-16
github actions: initial implementation

ana:
The checks have been copied almost verbatim from .travis.yml.

Remove previous ci.yml experiment, the checks made in this file
have been integrated on linux.yml

Here you can see how it looks in a PR: https://github.com/ana/avocado/pull/1


.github/workflows/ci.yml
```diff
@@ -1,26 +0,0 @@
-name: Avocado Gating CI
```
richtja, line 1: I think we shouldn't delete those checks. If I remember correctly, the idea is to run a style check before the unit and functional tests and if it fails, the other tests wouldn't be run. Unfortunately, I wasn't able to find an issue about it. So removing this file is going backwards in this effort. Maybe @beraldoleal  can give more info.

richtja (changes requested):

.github/workflows/ci.yml
beraldoleal: Yes @richtja.

The original ideal in order to save some time running those checks that we discussed is having a nice pipeline (for instance: style checks -> unit -> functional -> smoke checks) with some dependencies.

If styles and signed-off tests are failing, then we could exit and save some time. It is a tradeoff but with most of the CI systems having high price/compute hour, we concluded that we would give this approach a try, and see what happens.

This stage must be there, and being the first of all others. if this fails, then we won't proceed to the next stages.

beraldoleal (changes requested):

.github/workflows/code_coverage.yml
```diff
@@ -0,0 +1,41 @@
...
+          python-version: ${{ matrix.python-version }}
+      - name: Install and pre script
+        env:
+          SELF_CHECK_CONTINUOUS: "yes"
+          CC_TEST_REPORTER_ID: "387887b88a76f31c2c376219fc749689ea5975c8fe7fcd9609f1dcc139e053a6"
+        run: |
+         pip install -r requirements-selftests.txt
+         curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
+         chmod +x ./cc-test-reporter
+         ./cc-test-reporter before-build
```
beraldoleal, line 29: I understand that this was a copy-past from travis, and I' m ok with that approach, since we are running against the clock here. But would be nice if we had that Makefile/setup targets mapping here. (for future improvements)

.github/workflows/ci.yml
beraldoleal: Yes @richtja.

The original ideal in order to save some time running those checks that we discussed is having a nice pipeline (for instance: style checks -> unit -> functional -> smoke checks) with some dependencies.

If styles and signed-off tests are failing, then we could exit and save some time. It is a tradeoff but with most of the CI systems having high price/compute hour, we concluded that we would give this approach a try, and see what happens.

This stage must be there, and being the first of all others. if this fails, then we won't proceed to the next stages.

beraldoleal (changes requested):
Hi @ana, this is very welcome, but I left a main comment here for you.

.github/workflows/linux.yml
```diff
@@ -0,0 +1,37 @@
...
+         pip install -r requirements-selftests.txt
+      - name: Installing Avocado in develop mode
+        run: python3 setup.py develop --user
+      - name: Avocado version
+        run: avocado --version
+      - name: Make check
+        env:
+          AVOCADO_LOG_DEBUG: "yes"
+          AVOCADO_CHECK_LEVEL: "1"
+        run:  make check
```
willianrampazzo, line 36: Instead of running the full check, this step could run `selftests/check.py --disable-static-checks` and make this job dependent on the original static check job.

This avoids unnecessary multiple runs of the static checks.

ana: Thanks willian, make check is what we're currently using in travis.

willianrampazzo: Right. If we take the opportunity to do some small reorg, that would be good too, but I'm fine with migration then reorg to be on the safe side.

ana:
v2 a https://github.com/avocado-framework/avocado/pull/4698

ana:
Please, find v2 at https://github.com/avocado-framework/avocado/pull/4698

PR #4694, opened 2021-06-16
[sogtware-manager] Avoid using --output from `ar` command

willianrampazzo:
Some versions of `ar` do not have the --output parameter available. This patch uses a different approach changing the directory to the destination and then extracting the source file.


beraldoleal (approved):
LGTM, thank you.

clebergnu (approved):
LGTM, thanks!

PR #4698, opened 2021-06-17
GitHub Actions: initial implementation (v2)

ana:
The checks have been migrated almost verbatim from .travis.yml 

This is a first implementation and we'll improve form here, I'll open a RFC discussion so we can share ideas there.
There is a previous test on my github userspace here: https://github.com/ana/avocado/pull/3


.github/workflows/linux.yml
```diff
@@ -0,0 +1,79 @@
+name: Linux
+
+on: [pull_request]
```
beraldoleal, line 3: Any reason for removing "push" here? I mean, according to our contributor guide we should send multiple versions, but sometimes do push on minor changes.

.github/workflows/linux.yml
```diff
@@ -0,0 +1,79 @@
...
+        run: python -m avocado run passtest.py
+      - name: Quick lint checks
+        run: python setup.py lint
+      - name: Tree static check, unittests and fast functional tests (3.9 only)
+        if: ${{ matrix.python-version == '3.9' }}
+        env:
+          AVOCADO_LOG_DEBUG: "yes"
+          AVOCADO_CHECK_LEVEL: "1"
+        run: make check
+      - run: echo "🥑 This job's status is ${{ job.status }}."
```
beraldoleal, line 43: Is this avocado icon intentional?

.github/workflows/linux.yml
```diff
@@ -0,0 +1,79 @@
...
+        with:
+          fetch-depth: 0
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Display Python version
+        run: python -c "import sys; print(sys.version)"
+      - name: Install dependencies
+        run: |
```
beraldoleal, line 27: Do you plan to have multiple line commands here?

beraldoleal (changes requested):
Hi @ana, I left a few minor comments for you.

Besides that, the only thing that we could improve is the presentation (maybe is personal style, and I will be the only one on this):

In general I prefer one single pipeline instead of multiples, this might improve a little bit the presentation on PR and also when reading the pipeline. For instance, by adding multiple workflows I find difficult to see the overall picture of the build (Maybe I'm just lost on GH interface). I was thinking about adding all the steps on the ci.yaml (already defined) we have only one workflow.

I'm not saying that this is the way to go, we just need to try-experiment and see what is best.

In general, nice work here.

beraldoleal:
And we can see already a huge improvement, 25 minutes vs a few minutes (sometimes seconds) from the GH actions.

.github/workflows/linux.yml
ana: No, it's a left over from separating steps.

willianrampazzo:
> In general I prefer one single pipeline instead of multiples, this might improve a little bit the presentation on PR and also when reading the pipeline. For instance, by adding multiple workflows I find difficult to see the overall picture of the build (Maybe I'm just lost on GH interface). I was thinking about adding all the steps on the ci.yaml (already defined) we have only one workflow.

+1 for a single pipeline for pull requests!

.github/workflows/linux.yml
beraldoleal: Fancy!

willianrampazzo: It shows up on my vim! :)

richtja:
> > In general I prefer one single pipeline instead of multiples, this might improve a little bit the presentation on PR and also when reading the pipeline. For instance, by adding multiple workflows I find difficult to see the overall picture of the build (Maybe I'm just lost on GH interface). I was thinking about adding all the steps on the ci.yaml (already defined) we have only one workflow.
> 
> +1 for a single pipeline for pull requests!

+1 from me too

ana:
> Hi @ana, I left a few minor comments for you.
> 
> Besides that, the only thing that we could improve is the presentation (maybe is personal style, and I will be the only one on this):
> 
> In general I prefer one single pipeline instead of multiples, this might improve a little bit the presentation on PR and also when reading the pipeline. For instance, by adding multiple workflows I find difficult to see the overall picture of the build (Maybe I'm just lost on GH interface). I was thinking about adding all the steps on the ci.yaml (already defined) we have only one workflow.
> 
> I'm not saying that this is the way to go, we just need to try-experiment and see what is best.
> 
> In general, nice work here.

Do you mean you prefer having it all in the same tab when checking in https://github.com/avocado-framework/avocado/pull/4698/checks  ? Because the appeareance in the "main tab" of the pull request won't change a lot.

I see your point of having everything in the same file, but then they're all the same action for github and we can't pause/pause if there is problems with one the actions or for example, we can't browse very easily to see why an action was working before and why is not working now.
In fact, as I see github actions, we might end up having more files if we add more type of tests and this will give us the possibility of which tests to enable/disable when working in our user spaces.
My idea of improvements would rather go in the direction of creating dependent jobs and for example start the tests on python 3.9 and if they work then running in older releases.

.github/workflows/linux.yml
ana: Yes, it's intentional. if you update the PR with a force-push it will rebuild. 
The idea here is to avoid automatic builds every time you push something in your own fork or when someone forks stuff. As far I know all changes in avocado are merged by a PR, aren't they?

ana:
And we got a prettier workflow at https://github.com/avocado-framework/avocado/pull/4703. Closing.

PR #4699, opened 2021-06-17
Nrunner helper methods for sending messages [v2]

richtja:
It brings helper methods for creating avocado messages. This will
simplify the usage of avocado messages for runners developers. Because
they will be able to use a messaging system without knowledge of the
internal structure.

Reference: #4627

---
Changes form v1 (#4662)
* Documentation and docstring fixes.
* Removal of message helpers with base runners because of podman issue
* Usage of a OOP approach

### DIFF
--- a/avocado/core/runners/utils/messages.py
+++ b/avocado/core/runners/utils/messages.py
@@ -1,0 +1,209 @@
+import logging
+import sys
+import time
+
+from ... import output
+
+
+class GenericMessage:
+    message_status = None
+
+    @classmethod
+    def _prepare_message(cls, additional_info=None):
+        """Prepare a message dict with some basic information.
+
+        This will add the keyword 'status' and 'time' to all messages.
+
+        :param: addional_info: Any additional information that you
+                               would like to add to the message.
+        :type additional_info: dict
+        :return: message dict which can be send to avocado server
+        :rtype: dict
+        """
+        status = {}
+        if additional_info is not None:
+            status = additional_info
+        status.update({'status': cls.message_status,
+                       'time': time.monotonic()})
+        return status
+
+    @classmethod
+    def get(cls, **kwargs):
+        """Creates message base on it's type with  all necessary information.
+
+        :return: message dict which can be send to avocado server
+        :rtype: dict
+        """
+        kwargs = {key: value for (key, value) in kwargs.items()
+                  if value is not None}
+        return cls._prepare_message(additional_info=kwargs)
+
+
+class StartedMessage(GenericMessage):
+    message_status = 'started'
+
+
+class RunningMessage(GenericMessage):
+    """Creates running message without any additional info."""
+    message_status = 'running'
+
+
+class FinishedMessage(GenericMessage):
+    message_status = 'finished'
+
+    @classmethod
+    def get(cls, result, fail_reason=None, returncode=None):  # pylint: disable=W0221
+        """Creates finished message with  all necessary information.
+
+        :param result: test result
+        :type result values for the statuses defined in
+                     :class: avocado.core.teststatus.STATUSES
+        :param fail_reason: parameter for brief specification, of the failed
+                            result.
+        :type fail_reason: str
+        :param returncode: exit status of runner
+        :return: finished message
+        :rtype: dict
+        """
+        return super().get(result=result,
+                           fail_reason=fail_reason,
+                           returncode=returncode)
+
+
+class GenericRunningMessage(GenericMessage):
+    message_status = 'running'
+    message_type = None
+
+    @classmethod
+    def _get_running_message(cls, msg):
+        """Prepare a message dict with necessary information for specific type.
+
+        :param msg: message data. If the message is str, it will be encoded
+                    with utf-8.
+        :type msg: str, bytes
+        :return: message dict which can be send to avocado server
+        :rtype: dict
+        """
+        message = {'type': cls.message_type, 'log': msg}
+        if type(msg) is not bytes:
+            msg = msg.encode('utf-8')
+            message.update({'log': msg, 'encoding': 'utf-8'})
+        return message
+
+    @classmethod
+    def get(cls, msg, **kwargs):  # pylint: disable=W0221
+        """Creates running message with  all necessary information.
+
+        :param msg: log of running message
+        :type msg: str
+        :return: running message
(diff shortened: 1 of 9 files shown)


clebergnu (approved):
@richtja This LGTM.  But, we need to track the still pending changes to the runners in `avocado/core/nrunner.py`.  When we replace the distribution model for the nrunner, as per issue #4336.

clebergnu (changes requested):
Actually, I just noticed the comparison spotted by LGTM, and it looks like it's spot on.  Let me know if there's a reason for using the identify comparison (I couldn't figure out the reason).

richtja:
@clebergnu yes you are right, I fixed it with `force-push`

clebergnu (approved):
Hi @richtja, LGTM, but can you please create an issue about the pending migration in `avocado/core/nrunner.py`? Thanks.

willianrampazzo (approved):

pevogam:
Hi @richtja will this need some adaptation on the Avocado VT side? If there is a corresponding PR there I could test these two together (without blocking merging for you of course).

avocado/core/runners/utils/messages.py
```diff
@@ -0,0 +1,212 @@
...
+
+        :param: addional_info: Any additional information that you
+                               would like to add to the message.
+        :type additional_info: dict
+        :return: message dict which can be send to avocado server
+        :rtype: dict
+        """
+        status = {}
+        if isinstance(additional_info, dict):
+            status = additional_info
```
beraldoleal, line 25: You are saying on the docstring that `additional_info` should be a dict, and with this check here, gives us the impression that another type could be supported. Because of the docstring, Maybe you could consider using `if is not None` here.

richtja:
> Hi @richtja will this need some adaptation on the Avocado VT side? If there is a corresponding PR there I could test these two together (without blocking merging for you of course).

Hi @pevogam this shouldn't have affect behavior of avocado-vt. When this will be merged, I am planing to use it in avocado-vt, because it will decrease the maintaining of vt runner, but it will work without it. I can let you know when the vt PR will be created, if you are interested.

avocado/core/runners/utils/messages.py
```diff
@@ -0,0 +1,212 @@
...
+class RunningMessage(GenericMessage):
+    """Creates running message without any additional info."""
+    message_status = 'running'
+
+
+class FinishedMessage(GenericMessage):
+    message_status = 'finished'
+
+    @classmethod
+    def get(cls, result, fail_reason=None, returncode=None):  # pylint: disable=W0221
```
beraldoleal, line 55: nitpick: I would be happier if the methods signature were the same and we don't need this disable.

avocado/core/runners/utils/messages.py
```diff
@@ -0,0 +1,212 @@
...
+                            result.
+        :type fail_reason: str
+        :param returncode: exit status of runner
+        :return: finished message
+        :rtype: dict
+        """
+        return super(FinishedMessage, cls).get(result=result,
+                                               fail_reason=fail_reason,
+                                               returncode=returncode)
+
```
beraldoleal, line 71: Unless I'm missing something, this could be: `super().get(...` Any reason for that?

avocado/core/runners/utils/messages.py
```diff
@@ -0,0 +1,212 @@
...
+    def get(cls, msg, **kwargs):  # pylint: disable=W0221
+        """Creates running message with  all necessary information.
+
+        :param msg: log of running message
+        :type msg: str
+        :return: running message
+        :rtype: dict
+        """
+        kwargs.update(cls._get_running_message(msg))
+        return super(GenericRunningMessage, cls).get(**kwargs)
```
beraldoleal, line 103: same here.

avocado/core/runners/utils/messages.py
```diff
@@ -0,0 +1,212 @@
...
+        super(FileMessage, cls).get(msg=msg, path=path)
+
+
+def _get_message_by_type(message_type):
+    """Get the message class based on message_type"""
+
+    messages = [LogMessage, StdoutMessage, StderrMessage, WhiteboardMessage]
+    for message in messages:
+        if message_type == message.message_type:
+            return message
```
beraldoleal, line 140: nitpick: maybe here we could use a dict:

```python
spported_types = {'log': LogMessage,
                  'file': FileMessage' # that is missing here on your list
                  'stderr': StderrMessage}
return supported_types[message_type]
```

And because of that, maybe this method would be unnecessary.

beraldoleal (commented):
Hi @richtja, thanks for the changes and again, this is a such nice improvements.

I left a few comments for you.

avocado/core/runners/utils/messages.py
```diff
@@ -0,0 +1,212 @@
+import logging
+import sys
+import time
+
+from ... import output
```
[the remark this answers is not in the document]
richtja: sure

avocado/core/runners/utils/messages.py
richtja: Yes, you are right, I will change that.

avocado/core/runners/utils/messages.py
richtja: I know that it would be better to have the same method's signature, but each message needs different attributes and I want to have those attributes to be explicitly set. Because I think, this will be easier for runner creators, who will be those methods using. 

But maybe I am missing some solution which can do both, so if you have some ideas I would love to hear them :)

avocado/core/runners/utils/messages.py
richtja: Yes, I will fix that.

avocado/core/runners/utils/messages.py
richtja: True, this is a better solution.

avocado/core/runners/utils/messages.py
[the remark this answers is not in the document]
richtja: unfortunately, I find out that I can't do it because of modules-boundaries check

richtja:
@beraldoleal  thanks for your comments, I resolved almost all of them with `force-push`.

beraldoleal (approved):
LGTM

PR #4703, opened 2021-06-18
GitHub Actions: initial implementation (v3)

ana:
I have added all the checks to the same workflow and also dependencies between jobs. You can see it graphically at: https://github.com/ana/avocado/actions/runs/949165905

There are 3 main jobs running first: Windows, Linux and OS X with latest python. 
In the case of Windows and OS X, all the test run in the latest Python stable release - 3.9 right now - and if everything  goes fine, then it'll run the smokechecks in Python 3.6  - 3.8. If during the run of the main job in Python 3.9 any of the steps fails, everything is canceled.

For Linux,  all the test run in the latest Python stable release 3.9 and if everything goes fine, then it'll run the checks in Python 3.6-3.8 and the code coverage test that is a long one.

What happens when the test fails? Here an example where the signed off has been purposely forgotten:
https://github.com/ana/avocado/actions/runs/949216389

We need to discuss if we want to test all supported python versions extensively and if we should add more dependencies between jobs. In the meantime, I think this is an improvement over travis already :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,26 +1,149 @@
-name: Avocado Gating CI
+name: GH Actions
 
-on: [push, pull_request]
+on: [pull_request]
 
 jobs:
-  lint:
-    name: Logical, stylistic, analytical and formatter checks
+
+  full-smokecheck-linux:
+
+    name: Linux with Python ${{ matrix.python-version }}
     runs-on: ubuntu-20.04
 
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8, 3.9]
+      fail-fast: false
+
     steps:
-      - name: Checkout the code
+      - run: echo "Job triggered by a ${{ github.event_name }} event on branch is ${{ github.ref }} in repository is ${{ github.repository }}, runner on ${{ runner.os }}"
+      - name: Check out repository code
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
-
-      - name: Installing selftests requirements
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Display Python version
+        run: python -c "import sys; print(sys.version)"
+      - name: Install dependencies
         run: pip install -r requirements-selftests.txt
-
       - name: Installing Avocado in develop mode
         run: python3 setup.py develop --user
-
       - name: Avocado version
         run: avocado --version
+      - name: Avocado smoketest
+        run: python -m avocado run passtest.py
+      - name: Quick lint checks
+        run: python setup.py lint
+      - name: Tree static check, unittests and fast functional tests
+        env:
+          AVOCADO_LOG_DEBUG: "yes"
+          AVOCADO_CHECK_LEVEL: "1"
+        run: make check
+      - run: echo "🥑 This job's status is ${{ job.status }}."
 
-      - name: Running lint checks
-        run: python3 setup.py lint
+  code-coverage:
+
+    name: Code Coverage
+    runs-on: ubuntu-20.04
+
+    strategy:
+      matrix:
+        python-version: [3.9]
+
+    steps:
+      - run: echo "Job triggered by a ${{ github.event_name }} event on branch is ${{ github.ref }} in repository is ${{ github.repository }}, runner on ${{ runner.os }}"
+      - name: Check out repository code
+        uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install and pre script
+        env:
+          SELF_CHECK_CONTINUOUS: "yes"
+          CC_TEST_REPORTER_ID: "387887b88a76f31c2c376219fc749689ea5975c8fe7fcd9609f1dcc139e053a6"
+        run: |
+         pip install -r requirements-selftests.txt
+         curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
+         chmod +x ./cc-test-reporter
+         ./cc-test-reporter before-build
+      - name: Run script
+        env:
+          SELF_CHECK_CONTINUOUS: "yes"
+          CC_TEST_REPORTER_ID: "387887b88a76f31c2c376219fc749689ea5975c8fe7fcd9609f1dcc139e053a6"
+        run: make develop && ./selftests/run_coverage
+      - name: post script
+        env:
+          SELF_CHECK_CONTINUOUS: "yes"
+          CC_TEST_REPORTER_ID: "387887b88a76f31c2c376219fc749689ea5975c8fe7fcd9609f1dcc139e053a6"
+        run: ./cc-test-reporter after-build
+      - run: echo "🥑 This job's status is ${{ job.status }}."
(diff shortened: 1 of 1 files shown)


richtja (approved):
This is great, thank you.

richtja:
And for the discussion, I think we should test all supported python versions extensively, because most of the time we have failiers on different versions inside functional and unit tests which smoke test don't catch. I don't remember that we ever catch a version issue which can be reproduced by  `avocado run passtest.py`

willianrampazzo:
Hi, @ana, this is a nice improvement! 

I also agree with @richtja regarding extensive tests at least on Linux. It happened to me sometimes to have failed tests in a lower version of python and not in the newest because I was using a new syntax. This should be reduced starting from 3.6, but I think it is worth checking the code if not in all python versions, at least in the lowest and in the newest versions.

For the smoke checks on OSX and Windows, I think testing in one version is enough as those are not officially supported platforms.

ana:
Hi,
I have just done a force-push taking into account the latest comments:
- All the python version perform all the test. Once the test are OK in Python 3.9, it'll test Python 3.5-3.8
-  Only test in the latest version in Window and OSX

The actions are still running while I write this, they look like this https://github.com/avocado-framework/avocado/actions/runs/956784255

beraldoleal (commented):
Hi @ana,

I like the idea you introduced about running older versions of python only if latest python is working. However, after staring for the duplicated code here for a while, I was wondering why not adding all the python versions on the same stage, avoiding duplicated code at same time?

There is other reason for that, besides saving run time?

willianrampazzo:
> * All the python version perform all the test. Once the test are OK in Python 3.9, it'll test Python 3.5-3.8

What do you think about running just the unit and functional tests on 3.6-3.8? I don't think the other tests will catch anything different on other python versions.

Other than that, it looks good to me.

willianrampazzo:
> Hi @ana,
> 
> I like the idea you introduced about running older versions of python only if latest python is working. However, after staring for the duplicated code here for a while, I was wondering why not adding all the python versions on the same stage, avoiding duplicated code at same time?
> 
> There is other reason for that, besides saving run time?

I think the idea is to reduce the processing time spent if a test fails. If a unit/functional test fails on 3.9, there is a good chance it fails on other python versions.

beraldoleal:
> > Hi @ana,
> > I like the idea you introduced about running older versions of python only if latest python is working. However, after staring for the duplicated code here for a while, I was wondering why not adding all the python versions on the same stage, avoiding duplicated code at same time?
> > There is other reason for that, besides saving run time?
> 
> I think the idea is to reduce the processing time spent if a test fails. If a unit/functional test fails on 3.9, there is a good chance it fails on other python versions.

I see. However I'm not sure if the opposite is true.

If we have 'unlimited time' to run this, why wait for the 3.9 if we could run everything in parallel and spot the issues quicker?

Based on the time reported here, Python 3.9 took ~ 13min, and only then another 13min was needed to run others checks on < 3.9. In total we needed ~20min. And IIUC, we could save half of the time.

(Btw, I'm not against this approach, just trying to save some time here and reduce the duplicated lines of code, at same time)

willianrampazzo:
> > > Hi @ana,
> > > I like the idea you introduced about running older versions of python only if latest python is working. However, after staring for the duplicated code here for a while, I was wondering why not adding all the python versions on the same stage, avoiding duplicated code at same time?
> > > There is other reason for that, besides saving run time?
> > 
> > 
> > I think the idea is to reduce the processing time spent if a test fails. If a unit/functional test fails on 3.9, there is a good chance it fails on other python versions.
> 
> I see. However I'm not sure if the opposite is true.
> 
> If we have 'unlimited time' to run this, why wait for the 3.9 if we could run everything in parallel and spot the issues quicker?
> 
> Based on the time reported here, Python 3.9 took ~ 13min, and only then another 13min was needed to run others checks on < 3.9. In total we needed ~20min. And IIUC, we could save half of the time.
> 
> (Btw, I'm not against this approach, just trying to save some time here and reduce the duplicated lines of code, at same time)

Okay, I understood your point. So, as GH Actions is unlimited right now (https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration#usage-limits) I don't see a problem running everything in the same stage, nor the approach proposed by @ana :) except for the longer execution time showed by @beraldoleal.

ana:
@willianrampazzo @beraldoleal I'm making a single comment to reply everything pending.

There is some context missing in how GitHub Actions (GHA) works that helps explaining some of the choices here.
A job matrix running all the tests for Python 3.5-3.9 is quick to implement and does not duplicate code. However if one of the job fails, GHA will kill stop running all the remaining jobs from that same job matrix no matter the status. 
For example, if the job fails in Python 3.5, it will kill all 3 others jobs running Python 3.6-3.9 and in most cases we won't know if the problem is only in 3.5, some of them or all of them. 
When I was doing tests initially with a job matrix for Python 3.5-3.9 and `make check` was failing at the signedoff-check, it was random what job would arrive at this check first, fail and then GHA would kill all the others. I didn't find a way to privilege one of job (3.9) to run with more priority than the others.
The current implementation is my middle ground between duplicating code (only 2 times) and being certain it works in 3.8 before  running a job matrix with all the other versions.  
Having the 4 version of python to run at the same time independently (not in a job matrix) would require the code duplicated 4 times. GHA doesn't support things like custom functions or yaml anchors to avoid this.

With this on mind, what solution would you prefer?

beraldoleal:
Hi @ana,

> For example, if the job fails in Python 3.5, it will kill all 3 others jobs running Python 3.6-3.9 and in most cases we won't know if the problem is only in 3.5, some of them or all of them.

Today, unless I'm missing something, with your proposal we have this twice right? First 3.6-3.8 depending on 3.9, then on the 3.6-3.8 job matrix as you described. Is my understanding right? If so, IMO we are ok to go with a single job matrix 3.6-3.9.

Besides that, most of the common errors (but syntax on new versions) are going to be duplicated regardless the version.

In the worst case we will have to run the CI job again after an initial fix. And this could also be possible here (specially with unsupported syntax when using newer styles).

> Having the 4 version of python to run at the same time independently (not in a job matrix) would require the code duplicated 4 times. GHA doesn't support things like custom functions or yaml anchors to avoid this.

Have you played with `strategy.fail-fast`? Based on the description seems to useful (but I didn't tested, so I'm not sure):

> strategy.fail-fast:
> When set to true, GitHub cancels all in-progress jobs if any matrix job fails. Default: true 

Please let me know if I understood it right and if this makes sense to you.

ana:
> Have you played with `strategy.fail-fast`? Based on the description seems to useful (but I didn't tested, so I'm not sure):
> 
> > strategy.fail-fast:
> > When set to true, GitHub cancels all in-progress jobs if any matrix job fails. Default: true

No! I didn't see that option. If it works as announced it solves the dilemma from my previous comment.

ana:
> > Have you played with `strategy.fail-fast`? Based on the description seems to useful (but I didn't tested, so I'm not sure):
> > > strategy.fail-fast:
> > > When set to true, GitHub cancels all in-progress jobs if any matrix job fails. Default: true
> 
> No! I didn't see that option. If it works as announced it solves the dilemma from my previous comment.

@beraldoleal  it worked, thanks!
I just did a force-push and GHA are currently building, see https://github.com/avocado-framework/avocado/actions/runs/960792103

willianrampazzo (approved):

clebergnu (approved):
Yay! This looks great! Thanks a lot!

beraldoleal (approved):
Thank you @ana for the changes.

PR #4707, opened 2021-06-21
selftests/check_tmp_dirs: only check for dirs

ana:
This was checking for both files and directories, with this change
it'll only check for directories.


willianrampazzo (approved):

clebergnu (approved):
LGTM, thanks!

PR #4714, opened 2021-06-23
.travis.yml: update following move to GHA

ana:
Update distribution from Xenial LTS (16.06) to Focal LTS (20.04).
Update to Python 3.9
Only leave non-amd64 builds: arm64, ppc64le and s390x

This will make Travis to look like this https://travis-ci.com/github/ana/avocado/builds/230556863

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,17 @@
 os: linux
-dist: xenial
+dist: focal
 language: python
 cache: pip
-arch: amd64
+arch:
+  - s390x
+  - ppc64le
+  - arm64
 
 python:
-    - "3.6"
-    - "3.7"
-    - "3.8"
+    - "3.9"
 
 env:
-      AVOCADO_LOG_DEBUG=yes
-      AVOCADO_CHECK_LEVEL=1
+    - AVOCADO_PARALLEL_LINT_JOBS=1
 
 addons:
   apt:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,13 +1,11 @@
 os: linux
-dist: xenial
+dist: focal
 language: python
 cache: pip
 arch: amd64
```
willianrampazzo, line 5: I think this can also be removed.

.travis.yml
```diff
@@ -52,7 +35,7 @@ jobs:
       # Disable pip cache dir temporarily
       install:
        - pip install --no-cache-dir -r requirements-selftests.txt
-    - python: "3.8"
+    - python: "3.9"
       arch: ppc64le
```
willianrampazzo, line 39: So, ppc64le is one of the architectures that consume credits. The free archs are s390x and arm. Are we keeping ppc? (cc @beraldoleal @clebergnu @richtja)

ana: @willianrampazzo  During my build tests of this PR in my userspace, none of the three arches have consumed credits. According to the documentation, these are offered for free from partners:
https://docs.travis-ci.com/user/billing-overview/#partner-queue-solution

.travis.yml
ana: @willianrampazzo yes! I just removed it and replaced it with a list of all the supported arches. This made the YAML file shorter :)

ana:
Sorry, following @willianrampazzo 's comment, I realized I could make a shorter YAML file and pushed an update.
It looks exactly the same with less code https://travis-ci.com/github/ana/avocado/builds/230572352

.travis.yml
willianrampazzo: Ah, thanks for checking that! When I first read this documentation, I thought IBM was supporting only s390x. So, we are good here!

willianrampazzo (approved):
Much better, thanks!

clebergnu (approved):
LGTM, thanks!

PR #4719, opened 2021-06-23
Forcing python3 as default environment alias

beraldoleal:
We are using python3 as dependency, and some distributions don't have 'python' symlink. Let's use python3 for all envs.


willianrampazzo (approved):
Thanks for that!

clebergnu (approved):
Yes, this matches PEP 394:

"When packaging third party Python scripts, distributors are encouraged to change less specific shebangs to more specific ones. This ensures software is used with the latest version of Python available, and it can remove a dependency on Python 2."

Thanks!

PR #4725, opened 2021-06-24
Add custom setuptools command for 'develop' and 'develop-external' - v2

ana:
This is a v2 of https://github.com/avocado-framework/avocado/pull/4711

Migrate the extra steps of:
* `make develop`
* `make develop`
* `make develop-external`
to be performed by a customized 'develop' and 'developexternal' commands with setuptools.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,6 +37,15 @@ def get_long_description():
     return readme_contents
 
 
+def walk_plugins_setup_py(action_name="UNNAMED", action=None,
+                          directory=os.path.join(BASE_PATH, "optional_plugins")):
+
+    for plugin in list(Path(directory).rglob("*/setup.py")):
+        parent_dir = plugin.parent
+        print(">> {} {}".format(action_name, parent_dir))
+        run([sys.executable, "setup.py"] + action, cwd=parent_dir, check=True)
+
+
 class Clean(clean):
     """Our custom command to get rid of junk files after build."""
 
@@ -69,11 +78,7 @@ class Clean(clean):
 
     @staticmethod
     def clean_optional_plugins():
-        for plugin in list(Path(os.getcwd()).rglob("./optional_plugins/*/setup.py")):
-            parent_dir = plugin.parent
-            print(">> CLEANING {}".format(parent_dir))
-            run('{} setup.py clean --all'.format(sys.executable),
-                shell=True, cwd=parent_dir, check=True)
+        walk_plugins_setup_py(action_name="CLEANING", action=["clean", "--all"])
 
 
 class SimpleCommand(Command):
--- a/setup.py
+++ b/setup.py
@@ -104,6 +104,26 @@ class Develop(setuptools.command.develop.develop):
             super().run()
 
 
+class DevelopExternal(setuptools.command.develop.develop):
+    """Install Avocado's external plugins in develop mode.."""
+
+    description = "Install Avocado's external plugins in develop mode. \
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -81,6 +82,28 @@ def clean_optional_plugins():
...
+
+        # python setup.py develop --user
+        # we install the plugins after installing avocado
+        elif self.user and self.uninstall:
+            walk_plugins_setup_py(action_name="UNLINK", action=["develop", "--uninstall", "--user"])
+            setuptools.command.develop.develop.run(self)
+
+        # other cases: do nothing and call parent function
+        else:
+            setuptools.command.develop.develop.run(self)
```
beraldoleal, line 104: `super().run()`

setup.py
```diff
@@ -81,6 +82,28 @@ def clean_optional_plugins():
...
 
+class Develop(setuptools.command.develop.develop):
+    """Custom develop command."""
+
+    def run(self):
+
+        # python setup.py develop --user --uninstall
+        # we uninstall the plugins before uninstalling avocado
+        if self.user and not self.uninstall:
+            setuptools.command.develop.develop.run(self)
```
beraldoleal, line 93: `super.run()`

setup.py
```diff
@@ -81,6 +82,28 @@ def clean_optional_plugins():
...
+            walk_plugins_setup_py(action_name="LINK", action=["develop", "--user"])
+
+        # python setup.py develop --user
+        # we install the plugins after installing avocado
+        elif self.user and self.uninstall:
+            walk_plugins_setup_py(action_name="UNLINK", action=["develop", "--uninstall", "--user"])
+            setuptools.command.develop.develop.run(self)
```
beraldoleal, line 100: `super.run()`

setup.py
```diff
@@ -290,6 +310,7 @@ def run(self):
           python_requires='>=3.6',
           cmdclass={'clean': Clean,
                     'develop': Develop,
+                    'developexternal': DevelopExternal,
```
beraldoleal, line 313: I'm ok with having a new command, but since that you already defined a custom `develop` command, what about instead of using another command, if we had the `--external` option to the develop command? Have you considered that?

beraldoleal (changes requested):
Hi @ana, this is very welcome. Thanks. I left a few comments for you.

setup.py
ana: @beraldoleal I think that's going to make the develop command too complicated for little gain. I also find `--external` might be confusing. The name of the command should something like develop-external-plugins to make obvious you don't need it unless you're developing an external plugin.
A possible solution here is that if you have defined `AVOCADO_EXTERNAL_PLUGINS_PATH` then when you use the develop command also will install the external plugins. This would entirely remove the command. It would also reinstall everything else, but it's a quick action so it doesn't seem like a big deal. What do you think?

setup.py
ana: Thanks, I have updated this and the others two.

setup.py
beraldoleal: I think this approach is more prune to errors, since devs can have the env variable defined but don't remember, so IMO it is not so explicit.

I also don't think `--external` will make things complicated if they are documented. Take `python setup.py develop --help` and see all the options for develop for instance. We are just adding a new one. In fact, maybe we should respect `--no-deps` there, not sure.

But as I said, I'm ok with this approach as well, so I will let you choose. ;)

ana:
See v3 at #4731

PR #4734, opened 2021-06-25
[nrunner] inject run.test_parameters option on Test params

willianrampazzo:

### DIFF
--- a/avocado/core/runners/avocado_instrumented.py
+++ b/avocado/core/runners/avocado_instrumented.py
@@ -104,12 +104,18 @@ class AvocadoInstrumentedTestRunner(nrunner.BaseRunner):
             module_path, klass_method = runnable.uri.split(':', 1)
 
             klass, method = klass_method.split('.', 1)
+
+            # inject the run.test_parameters option in the Test params
+            # this handles the -p command-line argument
+            params = TreeNode().get_node("/", True)
+            params.value = dict(runnable.config.get('run.test_parameters', []))
+
             test_factory = [klass,
                             {'name': TestID(1, klass_method),
                              'methodName': method,
                              'config': runnable.config,
                              'modulePath': module_path,
-                             'params': (TreeNode(), []),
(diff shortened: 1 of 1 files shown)


willianrampazzo:
I'll handle the exec-test similar problem in a different PR.

avocado/core/runners/avocado_instrumented.py
```diff
@@ -104,12 +104,20 @@ def _run_avocado(runnable, queue):
...
 
             klass, method = klass_method.split('.', 1)
+
+            # inject the run.test_parameters option in the Test params
+            # this handles the -p command-line argument
+            test_parameters = {name: value for name, value in
+                               runnable.config.get('run.test_parameters', [])}
+            params = TreeNode().get_node("/", True)
+            params.value = test_parameters
+
```
beraldoleal, line 114: IIUC, `run.test_parameters` is a list of lists. Something like:

```python
"run.test_parameters": [["foo", "1"]]
```

So what do you think about this:

```python
params = TreeNode().get_node("/", True)
params.value = dict(runnable.config.get('run.test_parameters'))
```

Note: AFAICT, config.get() should return all the values, and the default value for `run.test_parameters` is already a empty list. So, when getting, the default `[]` wouldn't be necessary. But if that is not the case here, then you can keep the `[]`.

beraldoleal (commented):
Hi @willianrampazzo this LGTM, just one small comment for you.

avocado/core/runners/avocado_instrumented.py
willianrampazzo: Much better, thanks! I force-pushed the change.

beraldoleal (approved):
LGTM.

willianrampazzo:
@beraldoleal, I had to add the default value back as some tests running with the Job API are failing.

clebergnu (approved):
LGTM, thanks!

PR #4688, opened 2021-06-15
Extend runner config options

willianrampazzo:
This PR does 2 two changes to allow passing all the config options to a runnable:

* This handles the nrunner serialization/deserialization process involved in the config. It introduces a mechanism that converts sets into lists inside a dictionary. The dictionary key `__encoded_set__` is used in the decoding process to convert the option back to a set.
* Change the suite to pass all the config options to a runnable instead of filtering.

Issue #4352, opened 2020-12-08
Task with default status_uri


richtja:
When we're initializing a new `Task` object, we can have default `status_uri` same as default `nrunner.status_server_uri`. We are doing the same in `TestSuite.from_config()`

beraldoleal:
@richtja I'm revisiting this, and I have doubts if this makes sense.

For instance, status server URI could be something like '0.0.0.0:8080' (to be listening in multiple interfaces). In this case, makes no sense to send messages to this address. I would also prefer to avoid corner cases (and ifs) here.

Makes sense?

If you agree, we can close this.

richtja:
@beraldoleal AFAIK every time when we are creating the Task we are setting the `status_uri` from `nrunner.status_server_uri`. So if you set `nrunner.status_server_uri` to `0.0.0.0:8080` the `Task.status_services` will be set to `0.0.0.0:8080` this is won't be changed by this issue. My point is to not setting `status_uri` every time when we are creating `Task` as we are doing now, but have it as default.

beraldoleal:
@richtja I got confused with the status_server_listen. Sorry for the mess. I will try to implement this.

beraldoleal:
@richtja I did a second round of debugging here. Do you have any use case where you are creating a Task object with an empty `status_uris`?

AFAICT, today, during our runs, we are passing `status_uris` with the same value as  `nrunner.status_server_uri`.

Now, If you are creating an object Task manually, then I think that I understood the point here. IIUC, my suggestion is to remove `status_uris`  argument completely and use it as part of the configuration options/file. I will try to explain...

Today, we are passing all configuration options starting with `runner.` as runnable config. So since we are moving towards removing legacy runners, and having only one runner. We could also pass all configuration options starting with `nrunner.` to the config as well. And we have already there the `nrunner.status_server_uri`. This makes sense to me, what do you think?

richtja:
@beraldoleal For the context, I created this issue when we had to create Task inside TestSuite creation wchich was changed in 93f93764f311ef2a988dc23bdbac80db11cfdc3f that is why I'm mentioning the `TestSuite.from_config()` here. 

I don't know any use case with the empty `status_uris`.

AFAIK right now we are using one status server, but Tasks are able to send messages to the multiple status servers and I don't know what is the expected behavior of this. If two different tasks should be able to send messages to the two different status servers than I think your solution won't work. But I am not sure what exactly is the purpose of multiple status serves. But now we have this ability.

[...]

PR #4740, opened 2021-06-28
Task status_server_uri

richtja:
After the #4688 the `Task` has access to the 'nrunner.status_server_uri'
config variable, and it can be used for creating `Task.staus_services`.

Reference: #4352

### DIFF
--- a/avocado/core/nrunner.py
+++ b/avocado/core/nrunner.py
@@ -743,7 +743,10 @@ class Task:
         #: will not be accounted for on a Job's test results.
         self.category = category
         self.status_services = []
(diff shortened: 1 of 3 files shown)


avocado/core/nrunner.py
```diff
@@ -743,7 +743,10 @@ def __init__(self, runnable, identifier=None, status_uris=None,
         #: will not be accounted for on a Job's test results.
         self.category = category
         self.status_services = []
+        status_uris = status_uris or self.runnable.config.get('nrunner.status_server_uri')
```
beraldoleal, line 746: Maybe in a more radical change,  we could even, remove `status_uris` parameter.

beraldoleal (approved):
LGTM

PR #4744, opened 2021-06-28
[nrunner] inject run.test_parameters option on exec-test

willianrampazzo:


richtja (approved):
LGTM

beraldoleal (approved):
Ok @willianrampazzo, this makes sense. Current runner, is doing the same. So yes, thanks.

PR #4746, opened 2021-06-29
Config variable for resolvers [v2]

richtja:
The resolvers don't have access to the config variable, and they are
using settings.as_dict(). Unfortunately, this method returns only
config from command line and config files. When the configuration is
created by Job API, the settings.as_dict() doesn't respect that. This
leads to the incompatibility between resolvers config and suite config
created by TestSuite.from_config() in Job API.

This problem was hidden because avocado default resolvers don't need
config for resolving references and TestSuite.from_config() rewrites
the runnables config created by resolvers.

It is not such a big problem for avocado default resolvers, but it is a
problem for vt resolver which is using config in resolving process.

---
Changes from v1 (#4713):
* Passing the configuration automatically at the class initialization time.
* TAP resolver description fix

### DIFF
--- a/avocado/core/resolver.py
+++ b/avocado/core/resolver.py
@@ -19,6 +19,7 @@ Test resolver module.
 import os
 from enum import Enum
 
+from .. import settings
 from .enabled_extension_manager import EnabledExtensionManager
 from .exceptions import JobTestSuiteReferenceResolutionError
 
@@ -80,6 +81,13 @@ class ReferenceResolution:
                           self.info, self.origin)
 
 
+class ResolverMixin:
+    """Common utilities for Resolver implementations."""
+
+    def __init__(self, config=None):
+        self.config = config or settings.as_dict()
+
+
 class Resolver(EnabledExtensionManager):
 
     """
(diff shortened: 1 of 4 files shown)


avocado/plugins/resolvers.py
```diff
@@ -125,7 +126,9 @@ def resolve(reference):
         if criteria_check is not True:
             return criteria_check
 
+        runnable = Runnable('tap', reference,
+                            config=settings.filter_config(self.config,
+                                                          r'^runner\.'))
```
beraldoleal, line 131: After #4688, we are passing the entire config to the runners. So what you think about using the same approach here?

richtja: I was thinking about it, but I wanted to keep the same functionality as it is and just fix the bug. I see this as a different problem, so I don't wanted to  put it in the same PR, but if you this that is ok I can also delete the filtering.

clebergnu (approved):
LGTM, thanks!

richtja:
Hi @beraldoleal, thanks for your comment, I did an update with force-push.

willianrampazzo (approved):

avocado/core/resolver.py
```diff
@@ -80,6 +81,15 @@ def __repr__(self):
                           self.info, self.origin)
 
 
+class ResolverMixin:
+    """Common utilities for Resolver implementations."""
+
+    def __init__(self, config=None):
+        if config is None:
+            config = settings.as_dict()
+        self.config = config
```
beraldoleal, line 90: Maybe:

```python
self.config = config or settings.as_dict()
```

beraldoleal:
Besides a small comment (optional)  and the failures on packit/travis, LGTM.

beraldoleal (approved):

Issue #4492, opened 2021-03-31
Removing the relative imports


richtja:
After the discussion #3525 we should change the relative imports to absolute one if it is possible. And start using absolute imports only policy in the whole Avocado.

Issue #3525, opened 2020-01-22
RFC: Relative Imports vs Absolute Imports


beraldoleal:
Hi all,

I noticed that we have some selftests that checks for absolute imports on core (`selftests/modules-boundaries.sh`).

From PEP8:

> Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured (such as when a directory inside a package ends up on sys.path)...

> However, explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose.

> Standard library code should avoid complex package layouts and always use absolute imports.

I just would like to know what you think about that. What do you prefer and why?

Does anyone have any strong objections to removing relative imports?

willianrampazzo:
My preference is always for absolute imports, as they make the code clear to read.

I understand that relative imports can help in the development process while setting up a development environment, but, if possible, I would go with absolute imports.

In case it is not a pain and do not breaks what we have now, I agree with removing the relative imports.

apahim:
+1 for absolute imports between sub-packages, nor +1 or -1 for absolute imports between modules of the same sub-package, as the use cases for relative imports there are stronger.

beraldoleal:
@apahim can you elaborate on the "relative imports there are stronger", please?

apahim:
https://www.python.org/dev/peps/pep-0328/#rationale-for-relative-imports

apahim:
But as I said, nor +1 or -1. If you change it, awesome, if you don't, fine :)

beraldoleal:
Haa ok, thanks!

beraldoleal:
@clebergnu are you ok with moving from relative to absolute ? I would like to remove the `module-boundaries.sh` at least for `avocado/core`. What do you think?

PR #4747, opened 2021-06-29
`modules-boundaries` check in `setup.py lint`

richtja:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -105,6 +105,7 @@ class Linter(SimpleCommand):
             check_call('selftests/lint.sh')
             check_call('selftests/signedoff-check.sh')
             check_call('selftests/spell.sh')
+            check_call('selftests/modules-boundaries.sh')
         except CalledProcessError as e:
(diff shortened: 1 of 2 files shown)


ana (approved):

setup.py
```diff
@@ -105,6 +105,7 @@ def run(self):
             check_call('selftests/lint.sh')
             check_call('selftests/signedoff-check.sh')
             check_call('selftests/spell.sh')
+            check_call('selftests/modules-boundaries.sh')
```
beraldoleal, line 108: Hi @richtja, thanks for this. Maybe since you are adding this, we could take this opportunity to remove the need for relative imports on core, as we discussed here: #4492  and #3525 . What you think?

richtja: Right, I will update the script.

beraldoleal (approved):
LGTM.

willianrampazzo (approved):

PR #4752, opened 2021-06-29
Added support for ssh over sudo

otisdog8:
This allows commands to be run over sudo in ssh without
configuring passwordless sudo or using the root account to ssh.
This also solves the issue where avocado fails to configure
net interfaces on a remote machine that uses password-based sudo.

Signed-off-by: Jacob Root <otis@otisroot.com>

### DIFF
--- a/avocado/utils/network/common.py
+++ b/avocado/utils/network/common.py
@@ -7,6 +7,4 @@ def run_command(command, host, sudo=False):
     if host.__class__.__name__ == 'LocalHost':
         return process.system_output(command, sudo=sudo).decode('utf-8')
 
-    if sudo:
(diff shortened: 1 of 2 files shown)


willianrampazzo:
I believe this change is very welcome, but It brings back the discussion we had some months (maybe years?) whether we should add functionalities to the avocado ssh utility or use aexpect.

Thoughts @clebergnu, @beraldoleal, @richtja, @ana, and @wainersm (as you were involved in the discussion last time)?

richtja:
Since there is no activity on this topic for more than 2 years, I am closing this. @otisdog8 if you want to continue in this, feel free to reopen it.

PR #4761, opened 2021-07-02
Allow to run the selftests without testing the plugins

ana:
Updates to be able to opt-out all the plugins tests so we can split the plugins and still keep the tests in the avocado repository.


willianrampazzo (approved):

richtja (approved):
LGTM, thanks

PR #4775, opened 2021-07-13
core.data_dir: make datadir.paths.base_dir consisent

### LINKED ISSUE #4443: datadir.* with inconsist values

beraldoleal:
data_dir module was created way before our new settings logic was
introduced, default values now are defined at registration time. When
users define a custom value, in some cases Avocado is replacing this
value with a default hardcoded value, instead of respecting users'
choice (for instance when users have no permissions). base_dir is
already defined as default at `~/avocado/`, let's get the values from
the config dict here too. This fixes #4443.

### DIFF
--- a/avocado/core/__init__.py
+++ b/avocado/core/__init__.py
@@ -94,7 +94,7 @@ def register_core_options():
                          help_msg=help_msg)
 
     help_msg = 'Base directory for Avocado tests and auxiliary data'
-    default = prepend_base_path('/var/lib/avocado')
+    default = prepend_base_path('~/avocado')
     stgs.register_option(section='datadir.paths',
                          key='base_dir',
                          key_type=prepend_base_path,
--- a/avocado/core/data_dir.py
+++ b/avocado/core/data_dir.py
@@ -136,6 +136,9 @@ def get_data_dir():
         * GPG files
         * VM images
         * Reference bitmaps
+
+    .. warning:: This method is deprecated, get values from
+                 settings.as_dict() or self.config
     """
     warnings.warn(("get_data_dir() is deprecated, get values from "
                    "settings.as_dict() or self.config"), DeprecationWarning)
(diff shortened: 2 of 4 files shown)


willianrampazzo:
I don't have a strong opinion about where Avocado should allocate the base dir, so the code looks good to me.

beraldoleal:
> I don't have a strong opinion about where Avocado should allocate the base dir, so the code looks good to me.

This is just following the same approach as previous PRs on this.

richtja (approved):

beraldoleal:
Ops, I had the impression that @willianrampazzo gave a go-a-head, my bad.

@clebergnu let me know if we need a revert here.

willianrampazzo:
> Ops, I had the impression that @willianrampazzo gave a go-a-head, my bad.
> 
> @clebergnu let me know if we need a revert here.

Sorry, I added the comment and forgot to approve. I'm fine with the changes.

PR #4777, opened 2021-07-14
logging: we should not remove handlers when using the root log

beraldoleal:
We are messing up with the root log. After finishing the run, we
need to restore back level  and handlers. Because logging is a singleton,
according to the documentation we should be using a context manager for
selective logging, but unfortunately this will not be trivial today. So this should fix it.


willianrampazzo (approved):

richtja (commented):
Hi @beraldoleal LGTM, I just think that we have to do the changes also on the runners [side](https://github.com/avocado-framework/avocado/blob/master/avocado/core/runners/utils/messages.py#L182), because we also don't want to break the environment where the runners are spawned. But that is just a proposal, I am not 100% sure if it's a problem.

PR #4779, opened 2021-07-15
selftests: allow to only use half of the cpus for some tests

ana:
This should help with the current problem in travis on arm64.
Reference: https://github.com/avocado-framework/avocado/issues/4768


richtja:
I am closing this because the #4768 is already solved by #4782.

PR #4753, opened 2021-06-30
Instrumented tests with fail reason

richtja:
The avocado-instrumented runner wasn't gathering the fail-reason of
tests. This fix the problem and adds the appropriate test to catch
future bugs.

Reference: #4749
Signed-off-by: Jan Richter <jarichte@redhat.com>

PR #4780, opened 2021-07-15
Instrumented tests with fail reason [v2]

richtja:
The avocado-instrumented runner wasn't gathering the fail-reason of
tests. This fix the problem and adds the appropriate test to catch
future bugs.

Reference: #4749

---
Changes from v1 (#4753):
* Test fix - create its own status server.

### DIFF
--- a/avocado/core/runners/avocado_instrumented.py
+++ b/avocado/core/runners/avocado_instrumented.py
@@ -61,11 +61,13 @@ class AvocadoInstrumentedTestRunner(nrunner.BaseRunner):
             queue.put(early_state)
             instance.run_avocado()
             state = instance.get_state()
+            fail_reason = state.get('fail_reason')
             queue.put(messages.WhiteboardMessage.get(state['whiteboard']))
-            queue.put(messages.FinishedMessage.get(state['status'].lower()))
-        except Exception:
+            queue.put(messages.FinishedMessage.get(state['status'].lower(),
+                                                   fail_reason=fail_reason))
+        except Exception as e:
             queue.put(messages.StderrMessage.get(traceback.format_exc()))
-            queue.put(messages.FinishedMessage.get('error'))
+            queue.put(messages.FinishedMessage.get('error', fail_reason=str(e)))
 
     def run(self):
         yield messages.StartedMessage.get()
@@ -105,9 +107,9 @@ class AvocadoInstrumentedTestRunner(nrunner.BaseRunner):
                         yield message
                     if message.get('status') == 'finished':
                         break
-        except Exception:
+        except Exception as e:
             yield messages.StderrMessage.get(traceback.format_exc())
-            yield messages.FinishedMessage.get('error')
+            yield messages.FinishedMessage.get('error', fail_reason=str(e))
 
 
 class RunnerApp(nrunner.BaseRunnerApp):
--- a/selftests/functional/plugin/test_jsonresult.py
+++ b/selftests/functional/plugin/test_jsonresult.py
@@ -32,6 +32,19 @@ class JsonResultTest(TestCaseTmpDir):
             expected_logfile = path.join(test_data['logdir'], 'debug.log')
             self.assertEqual(expected_logfile, test_data['logfile'])
 
+    def test_fail_reason(self):
+        cmd_line = ('%s --config %s run --test-runner=nrunner '
+                    'examples/tests/failtest.py '
+                    '--job-results-dir %s --disable-sysinfo ' %
+                    (AVOCADO, self.config_file.path, self.tmpdir.name))
+        process.run(cmd_line, ignore_status=True)
+        json_path = path.join(self.tmpdir.name, 'latest', 'results.json')
+        with open(json_path, 'r') as json_file:
+            data = json.load(json_file)
+            test_data = data['tests'].pop()
+            self.assertEqual('This test is supposed to fail',
+                             test_data['fail_reason'])
+
     def tearDown(self):
         super(JsonResultTest, self).tearDown()
         self.config_file.remove()


ana (approved):

avocado/core/runners/avocado_instrumented.py
```diff
@@ -61,11 +61,13 @@ def _run_avocado(runnable, queue):
...
+            fail_reason = state.get('fail_reason')
             queue.put(messages.WhiteboardMessage.get(state['whiteboard']))
-            queue.put(messages.FinishedMessage.get(state['status'].lower()))
-        except Exception:
+            queue.put(messages.FinishedMessage.get(state['status'].lower(),
+                                                   fail_reason=fail_reason))
+        except Exception as e:
             queue.put(messages.StderrMessage.get(traceback.format_exc()))
-            queue.put(messages.FinishedMessage.get('error'))
+            queue.put(messages.FinishedMessage.get('error', fail_reason=str(e)))
```
beraldoleal, line 70: Aren't we loosing the first `fail_reason` here?  Let's say that  line 64 returns a non None value  and the exception cause wasn't  that line. So, we are replacing t he original reason, with the cause of the exception,  right?

What I'm missing here?

beraldoleal (commented):
Just one  small question for you:

avocado/core/runners/avocado_instrumented.py
richtja: Yes you are right this might happen. But if some test will fail and after that some error will occur, the test will be marked with error and IMO the reason of failure should be the error not the test results, because the error is more important. Also, saving the first `fail_reason` when error is occurred might be confusing, because you will get an error and as reason some assertionerror which didn't cause the real problem.

beraldoleal: Hi @richtja , I understand your point, but I'm still having mix feelings.

IMO it feels strange to get an exception (that probably is not related to my instrumented test) in the output of my run. Users should not care about errors on Avocado code.

Let's say that we have an exception while adding the state to the queue. This is an internal Avocado error, and IMO could be better placed at the debug log.

I might accept this, but as I said, I still have the feeling that we could do better here. What do you think?

richtja: So IIUC you are proposing to remove the `fail_reason` argument from this exception? Because in that case, I think we will be hiding some information from the user and I don't understand why to do that.

Also, I think we can't use the test `fail_reason` in this exception, because if the runner will crash the test results are compromised, and we can not use them as valid results.

beraldoleal: The point is that this block of code is too big for an exception. We don't know for sure if this is an Avocado error or a runner error with the test syntax itself.

One suggestion would be to keep `fail_reason` as the reason of the test (even that is None), *and* log as debug the exception and traceback on `debug.log` or specific test log, so we don't loose the details.

richtja: > The point is that this block of code is too big for an exception. We don't know for sure if this is an Avocado error or a runner error with the test syntax itself.

And the first clue what error it is, could be the `fail_reason`

> 
> One suggestion would be to keep `fail_reason` as the reason of the test (even that is None), _and_ log as debug the exception and traceback on `debug.log` or specific test log, so we don't loose the details.

We are already saving the traceback to the test stderr and as I said earlier I think It will be pity to have `fail_reason` empty when we know the error. Because we can't use  `fail_reason` as the reason of the test when the error it occurred.

beraldoleal: > > The point is that this block of code is too big for an exception. We don't know for sure if this is an Avocado error or a runner error with the test syntax itself.
> 
> And the first clue what error it is, could be the `fail_reason`

I mean programmatically on the code block.

> > One suggestion would be to keep `fail_reason` as the reason of the test (even that is None), _and_ log as debug the exception and traceback on `debug.log` or specific test log, so we don't loose the details.
> 
> We are already saving the traceback to the test stderr and as I said earlier I think It will be pity to have `fail_reason` empty when we know the error. Because we can't use `fail_reason` as the reason of the test when the error it occurred.

I'm just trying to separate what is a test error and what is an Avocado code error. Right now we are mixing those together what is very hard to debug issues. IMO MessagesFinished|StdErrMessage could have only test related errors. Adding there something like an generic `IndexError` trace back from an Avocado code would lead users to confusion even with a trace back saying that this is part of Avocado.

(Please, note that I'm not rejecting this, and this is marked as comment.)

willianrampazzo: We already have ERROR status and FAIL status. The FAIL is related to the test, and the ERROR is related to code problem/avocado. In this case, the solution here makes sense to me.

willianrampazzo (approved):

avocado/core/runners/avocado_instrumented.py
richtja: > I'm just trying to separate what is a test error and what is an Avocado code error. Right now we are mixing those together what is very hard to debug issues. IMO MessagesFinished|StdErrMessage could have only test related errors. Adding there something like an generic `IndexError` trace back from an Avocado code would lead users to confusion even with a trace back saying that this is part of Avocado.

Ok now I see your point and I think it is valid. But that is a problem of all runners, because we are using the MessagesFinished|StdErrMessage for both types of errors (avocado and test) and if we didn't send them, the test will be skipped without any information about the error. So I created an issue #4789 about this problem. I am not sure if I documented it well, so feel free to edit the description.

clebergnu (approved):
The points raised here are covered by #4796, so LGTM.

beraldoleal (approved):

PR #4782, opened 2021-07-15
selftests/check.py: workaround for travis on arm64

ana:
Use only half of the CPUS on arm64 for some tests to avoid problems
with the Travis arm64 builds.  Fixes #4786

References: https://github.com/avocado-framework/avocado/issues/4768

I think this is the best solution for the referenced issue, and it's better to do a  separate PR for other modifications in `selftests/check.py`

### DIFF
--- a/selftests/check.py
+++ b/selftests/check.py
@@ -2,7 +2,9 @@
 
 import argparse
 import glob
+import multiprocessing
 import os
+import platform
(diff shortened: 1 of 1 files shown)


clebergnu (approved):
Nice work! Thank you so much for getting this solved!

PR #4783, opened 2021-07-16
[utils] distro.py: enable util to detect distro of a remote machine

cforno12:
Enable the distro util to have the option to take a populated Session object
that represents a remote machine and detect the remote machine's distro.

### DIFF
--- a/avocado/utils/distro.py
+++ b/avocado/utils/distro.py
@@ -17,8 +17,11 @@ This module provides the client facilities to detect the Linux Distribution
 it's running under.
 """
 
+import logging
 import os
 import re
+
+LOGGER = logging.getLogger('avocado.test')
 
 __all__ = ['LinuxDistro',
            'UNKNOWN_DISTRO_NAME',
@@ -89,7 +92,9 @@ UNKNOWN_DISTRO = LinuxDistro(UNKNOWN_DISTRO_NAME,
 class Probe:
 
     """
-    Probes the machine and does it best to confirm it's the right distro
+    Probes the machine and does it best to confirm it's the right distro.
+    If given an avocado.utils.ssh.Session object representing another machine, Probe
+    will attempt to detect another machine's distro via an ssh connection.
     """
     #: Points to a file that can determine if this machine is running a given
     #: Linux Distribution. This servers a first check that enables the extra
@@ -109,8 +114,23 @@ class Probe:
     #: :attr:`CHECK_FILE_EXISTS`
     CHECK_VERSION_REGEX = None
 
-    def __init__(self):
+    def __init__(self, session=None):
         self.score = 0
+        self.session = session
+
+    def check_for_remote_file(self, file_name):
+        """
+        Checks if provided file exists in remote machine
+
+        :param file_name: name of file
+        :type file_name: str
+        :returns: whether the file exists in remote machine or not
+        :rtype: bool
+        """
+        if self.session and self.session.cmd("test -f %s" % file_name).exit_status == 0:
+            return True
+        else:
+            return False
 
     def check_name_for_file(self):
         """
@@ -133,7 +153,9 @@ class Probe:
         Get the distro name if the :attr:`CHECK_FILE` is set and exists
         """
         if self.check_name_for_file():
-            if os.path.exists(self.CHECK_FILE):
+            if self.check_for_remote_file(self.CHECK_FILE):
+                return self.CHECK_FILE_DISTRO_NAME
+            elif os.path.exists(self.CHECK_FILE):
                 return self.CHECK_FILE_DISTRO_NAME
 
     def check_name_for_file_contains(self):
@@ -161,11 +183,23 @@ class Probe:
         Get the distro if the :attr:`CHECK_FILE` is set and has content
         """
         if self.check_name_for_file_contains():
-            if os.path.exists(self.CHECK_FILE):
-                with open(self.CHECK_FILE) as check_file:
-                    for line in check_file:
-                        if self.CHECK_FILE_CONTAINS in line:
-                            return self.CHECK_FILE_DISTRO_NAME
+            check_file = None
+            if self.check_for_remote_file(self.CHECK_FILE):
+                check_file = self.session.cmd("cat %s" % self.CHECK_FILE).stdout_text.split('/n')
+            elif os.path.exists(self.CHECK_FILE):
+                try:
+                    check_file = open(self.CHECK_FILE)
+                except IOError as err:
+                    LOGGER.debug("Could not open %s", self.CHECK_FILE)
+                    LOGGER.debug("Exception: %s", str(err))
+                    return None
+
+            if not check_file:
+                return None
+
+            for line in check_file:
+                if self.CHECK_FILE_CONTAINS in line:
+                    return self.CHECK_FILE_DISTRO_NAME
 
     def check_version(self):
         """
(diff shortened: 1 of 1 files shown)


ana:
Hi @cforno12 .
Thanks this looks good to me. However two questions:
1. In what module is defined the Session object you're using here?
2. Do you feel like updating the selftests at https://github.com/avocado-framework/avocado/blob/master/selftests/unit/utils/test_distro.py to test this option as well?

cforno12:
Hi @ana , thanks for reviewing my PR! 

1. The Session object that I am using here is defined from avocado.utils.ssh.Session. Would you like me to strictly enforce that `session` is an instance of avocado.utils.ssh.Session?
2. Sure! It would be my first time writing tests for this framework, so I may ask some questions on this PR regarding that :)

ana:
> 
> 1. The Session object that I am using here is defined from avocado.utils.ssh.Session. Would you like me to strictly enforce that `session` is an instance of avocado.utils.ssh.Session?

Updating the comments in the functions you have modified so the API reference is cleat about what type the Session object is should be enough. This way it'll get nicely documented in the API reference https://avocado-framework.readthedocs.io/en/89.0/api/utils/avocado.utils.html#module-avocado.utils.distro

> 2. Sure!
Awesome, thanks!

cforno12:
Hi @ana, 

After scratching my head for a while, I'm not sure how I can write tests for this option without having a remote machine to retrieve files from. If you have any ideas on how to write a test(s) please let me know.

Thanks, Cris

ana:
Hi @cforno12 
 
> After scratching my head for a while, I'm not sure how I can write tests for this option without having a remote machine to retrieve files from. If you have any ideas on how to write a test(s) please let me know.

Indeed, this is tricky :-)
I have tested your changes manually and they worked fine. For reference, my test was:

```
from avocado.utils import ssh, distro
 
sss = ssh.Session('192.168.x.x', user='ana', password='ana')
sss.connect()
 
print(distro.detect(sss)) 
```

II wouldn't mind if you squashed all your commits into the first one since they  all  belong to the same change. Otherwise this looks good to me!

cforno12:
Thanks for testing my patch! Squashed all my commits as requested, thanks for reviewing @ana!

ana (approved):
LGTM, thanks!

avocado/utils/distro.py
```diff
@@ -109,8 +111,23 @@ class Probe:
     #: :attr:`CHECK_FILE_EXISTS`
     CHECK_VERSION_REGEX = None
 
-    def __init__(self):
+    def __init__(self, session=None):
         self.score = 0
+        self.session = session
+
+    def check_for_remote_file(self, file_name):
+        '''
```
beraldoleal, line 119: nitpick: to make it consistent with other comments, please use `"""`.

avocado/utils/distro.py
```diff
@@ -161,11 +180,18 @@ def name_for_file_contains(self):
...
-            if os.path.exists(self.CHECK_FILE):
-                with open(self.CHECK_FILE) as check_file:
-                    for line in check_file:
-                        if self.CHECK_FILE_CONTAINS in line:
-                            return self.CHECK_FILE_DISTRO_NAME
+            check_file = None
+            if self.check_for_remote_file(self.CHECK_FILE):
+                check_file = self.session.cmd("cat %s" % self.CHECK_FILE).stdout_text.split('/n')
+            elif os.path.exists(self.CHECK_FILE):
+                check_file = open(self.CHECK_FILE)
```
beraldoleal, line 187: What happen if `open()` raises `PermissionError`?

avocado/utils/distro.py
```diff
@@ -220,8 +253,11 @@ def release(self):
                 release = match.groups()[1]
         return release
 
-    def get_distro(self):
+    def get_distro(self, session=None):
```
beraldoleal, line 256: IIUC, this object has already a `self.session` that is set during `__init__()`. What is the difference between the first and this?

avocado/utils/distro.py
```diff
@@ -230,6 +266,9 @@ def get_distro(self):
 
         distro = None
 
+        if session:
+            self.session = session
```
beraldoleal, line 270: Unless I'm missing something, this is unnecessary if we remove the `session` argument.

avocado/utils/distro.py
```diff
@@ -342,8 +384,8 @@ class SUSEProbe(Probe):
     # and tests are looking for it in distro.name. So keep using it
     CHECK_FILE_DISTRO_NAME = 'SuSE'
 
-    def get_distro(self):
-        distro = Probe.get_distro(self)
+    def get_distro(self, session=None):
+        distro = Probe.get_distro(self, session)
```
beraldoleal, line 388: Same here, IIUC, we can remove the argument and use `super()`:
```
distro = super().get_distro(self.session)
```

avocado/utils/distro.py
```diff
@@ -407,18 +449,22 @@ def register_probe(probe_class):
...
 
+    :param session: ssh connection between another machine
+    :type session: avocado.utils.ssh.Session
     :returns: the detected :class:`LinuxDistro` or :data:`UNKNOWN_DISTRO`
     :rtype: :class:`LinuxDistro`
     """
     results = []
 
     for probe_class in REGISTERED_PROBES:
         probe_instance = probe_class()
```
beraldoleal, line 466: If my previous comments makes sense, then:

```python
probe_instance = probe_class(session)
distro_result = probe_instance.get_distro()
```

beraldoleal (changes requested):
Hi @cforno12, thank you for the improvements. 

IMO this entire utility module needs a refactoring, but for sure not introduced by your PR. So I just left a few comments for you.

cforno12:
Thanks for your suggestions @beraldoleal , I've implemented all of them. Much appreciated.

Looks like GH actions is complaining about how I imported `logging` (to log if distro.py couldn't open a file). Any idea on what formatting it expects?

ana:
> Thanks for your suggestions @beraldoleal , I've implemented all of them. Much appreciated.
> 
> Looks like GH actions is complaining about how I imported `logging` (to log if distro.py couldn't open a file). Any idea on what formatting it expects?

Hi @cforno12

The static check (isort) is expecting Python import definitions alphabetically. You need to put the logging import in the top like in this change:

```
--- a/avocado/utils/distro.py
+++ b/avocado/utils/distro.py
@@ -17,9 +17,9 @@ This module provides the client facilities to detect the Linux Distribution
 it's running under.
 """
 
+import logging
 import os
 import re
-import logging
 
 LOGGER = logging.getLogger('avocado.test')
 

```

You can run the static checks locally with `python3 setup.py lint`. The checks will require some requirements that can be installed with `pip3 install -r requirements-dev.txt`. This used to be `pip3 install -r requirements-selftests.txt` until a few days ago, but you are missing a lot commits. It'd be a good idea rebasing on top of master  before updating this PR.
Hope this helps!

avocado/utils/distro.py
```diff
@@ -222,6 +268,9 @@ def release(self):
 
     def get_distro(self):
         """
+        :param session: ssh connection between another machine
+        :type session: avocado.utils.ssh.Session
```
beraldoleal, line 272: There is no session parameter anymore. We need to remove the docstring.

avocado/utils/distro.py
```diff
@@ -343,7 +395,7 @@ class SUSEProbe(Probe):
     CHECK_FILE_DISTRO_NAME = 'SuSE'
 
     def get_distro(self):
-        distro = Probe.get_distro(self)
+        distro = super().get_distro(self.session)
```
beraldoleal, line 398: Since `get_distro()` doesn't have session as argument, this first argument is about self which is already passed:

```python
distro = super().get_distro()
```

avocado/utils/distro.py
```diff
@@ -407,17 +459,21 @@ def register_probe(probe_class):
 register_probe(OpenEulerProbe)
 
 
-def detect():
+def detect(session=None):
     """
-    Attempts to detect the Linux Distribution running on this machine
+    Attempts to detect the Linux Distribution running on this machine.
+    If given an avocado.utils.ssh.Session object, it will attempt to detect the
```
beraldoleal, line 465: Please, insert an empty line between the short description and summary paragraph to comply with PEP257.

beraldoleal (changes requested):
Hi @cforno12, thanks for accepting the changes. This looks almost good to me, just a few comments to you.

cforno12:
> You can run the static checks locally with python3 setup.py lint. The checks will require some requirements that can be installed with pip3 install -r requirements-dev.txt. This used to be pip3 install -r requirements-selftests.txt until a few days ago, but you are missing a lot commits. It'd be a good idea rebasing on top of master before updating this PR.
> Hope this helps!

Ah should have read the contributing guide more closely. Thanks!

@beraldoleal , I took all your suggestions once again. Good eye finding these imperfections :)

beraldoleal (approved):
Thanks @cforno12 for the changes. LGTM.

PR #4762, opened 2021-07-05
show correct start time in HTML reports

ana:
References: https://github.com/avocado-framework/avocado/issues/4709

PR #4785, opened 2021-07-19
show correct start time in HTML reports (v2)

ana:
This is a v2 from https://github.com/avocado-framework/avocado/pull/4762
Reference bug: https://github.com/avocado-framework/avocado/issues/4709

The first and third commit are identical to v1 and what changes is the middle commit that adds both actual_time_start and actual_time_end with nrunner.

The problem with monotonic for the legacy runner was introduced in 14937705f15c8334220599edc1047e7563d04c80 from May this year. The problem with the new runner was introduced in b600a5fe08112ee15b41b1150d849fd63ef276f6 from September 2020, which means it has been around for 10 months without anyone noticing!
I'm wondering how we can test this - because it's a "visual" test - to avoid this problem or a similar one happening a third time.

### DIFF
--- a/avocado/core/messages.py
+++ b/avocado/core/messages.py
@@ -13,6 +13,7 @@
 # Authors: Jan Richter <jarichte@redhat.com>
 
 import os
+import time
 
 from .test_id import TestID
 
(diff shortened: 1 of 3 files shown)


beraldoleal (approved):
LGTM, thanks.

PR #4786, opened 2021-07-19
Misc updates to setup.py, check.py and weekly ci

ana:
I have grouped 3 different changes in the same PR because I think they're relatively trivial. Tell me if you find it's a bad idea :-)

At bit of context to the changes in `check.py`. While I don't expect to use the disable options often in our internal CI, they are useful locally when debugging things.

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         # see list of available Python versions at https://github.com/actions/python-versions/blob/main/versions-manifest.json
-        python-version: [3.10.0-beta.3]
+        python-version: [3.10.0-beta.4]
       fail-fast: false
 
     steps:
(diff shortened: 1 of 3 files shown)


selftests/check.py
```diff
@@ -201,6 +201,22 @@ def parse_args():
     parser.add_argument('--disable-plugin-checks',
                         help='Disable checks for a plugin (by directory name)',
                         action='append', default=[])
+    parser.add_argument('--disable-selftests-nrunner',
```
willianrampazzo, line 204: We need a better naming here. I had the impression it was disabling all the `nrunner` tests, but that is not true.

.github/workflows/weekly.yml
```diff
@@ -46,5 +46,5 @@ jobs:
         env:
           AVOCADO_LOG_DEBUG: "yes"
           AVOCADO_CHECK_LEVEL: "1"
-        run: make check
+        run: python selftests/check.py --disable-plugin-checks robot
```
willianrampazzo, line 49: I see `develop` is executed above. All is good here.

selftests/check.py
ana: Good point. I'm going to rename it to `--disable-selftests-nrunner-interface`,

willianrampazzo (approved):

richtja (approved):
LGTM, thanks

clebergnu (approved):
LGTM, thanks!

PR #4787, opened 2021-07-19
logging: we should not remove handlers when using the root log (v2)

beraldoleal:
We are messing up with the root log. After finishing the run, we
need to restore back level and handlers. Because logging is a singleton,
according to the documentation we should be using a context manager for
selective logging, but unfortunately this will not be trivial today. So this should fix it.

### Changes from v2:

 * Added last commit to handle runners as well.

### DIFF
--- a/avocado/core/output.py
+++ b/avocado/core/output.py
@@ -643,6 +643,8 @@ def add_log_handler(logger, klass=logging.StreamHandler, stream=sys.stdout,
 
 
 def disable_log_handler(logger):
+    if not logger:
+        return
     if isinstance(logger, str):
         logger = logging.getLogger(logger)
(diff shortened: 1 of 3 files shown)


ana (approved):

richtja (approved):

avocado/core/output.py
```diff
@@ -643,6 +643,8 @@ def add_log_handler(logger, klass=logging.StreamHandler, stream=sys.stdout,
 
 
 def disable_log_handler(logger):
+    if not logger:
+        return
```
pevogam, line 647: @willianrampazzo I can narrow my problem in https://github.com/avocado-framework/avocado/issues/4797 to these two lines, if I revert those (keeping all other changes from this PR) everything works again.

PR #4798, opened 2021-07-23
Add command plugin and update Makefile

ana:
`setup.py`: Add custom command "plugin". This will allow users to install optional plugins easily.

`Makefile`: remove target requirements-plugins because requirements are already handled in the setup.py file of each plugin and add new target `install-user`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,18 +19,19 @@ RPM_BASE_NAME=python-avocado
 
 all:
 	@echo
+	@echo "User related targets:"
+	@echo "install-user:      Install avocado for the local user"
+	@echo
 	@echo "Development related targets:"
-	@echo "check:             Runs tree static check, unittests and fast functional tests"
-	@echo "smokecheck:        Runs the simplest possible avocado test execution"
 	@echo "develop:           Runs 'python setup.py --develop' on this tree alone"
 	@echo "develop-external:  Install Avocado's external plugins in develop mode."
 	@echo "                   You need to set AVOCADO_EXTERNAL_PLUGINS_PATH"
+	@echo "check:             Runs tree static check, unittests and fast functional tests"
+	@echo "smokecheck:        Runs the simplest possible avocado test execution"
 	@echo "clean:             Get rid of build scratch from this project and subprojects"
+	@echo "requirements-selftests:  Install runtime and selftests requirements"
 	@echo "variables:         Show the value of variables as defined in this Makefile or"
 	@echo "                   given as input to make"
-	@echo
-	@echo "Package requirements related targets"
-	@echo "requirements-selftests:  Install runtime and selftests requirements"
 	@echo
 	@echo "Platform independent distribution/installation related targets:"
 	@echo "source:       Create single source package with commit info, suitable for RPMs"
@@ -39,7 +40,7 @@ all:
 	@echo "pypi:         Create both source and binary wheel packages and show how to"
 	@echo "              upload them to PyPI"
 	@echo "python_build: Installs the build package, needed for source-pypi and wheel"
-	@echo "install:      Install on local system"
+	@echo "install:      Install on local system for all users (needs root)"
 	@echo "uninstall:    Uninstall Avocado and also subprojects"
 	@echo "man:          Generate the avocado man page"
 	@echo
@@ -91,6 +92,13 @@ python_build: pip
 
 clean:
 	$(PYTHON) setup.py clean --all
+
+install-user:
+	$(PYTHON) setup.py install --user
+	@echo
+	@echo "To install the optional plugins, please use: python setup.py plugin --install=<plugin_name>"
+	@echo
+	$(PYTHON) setup.py plugin --list
 
 uninstall:
 	$(PYTHON) setup.py develop --uninstall $(PYTHON_DEVELOP_ARGS)
@@ -144,4 +152,4 @@ propagate-version:
 		else echo ">> Skipping $$DIR"; fi;\
 	done
 
-.PHONY: source source-pypi wheel pypi install clean uninstall requirements-selftests smokecheck check develop develop-external propagate-version variables
+.PHONY: source source-pypi wheel pypi install install-user clean uninstall requirements-selftests smokecheck check develop develop-external propagate-version variables
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,6 @@ all:
 	@echo
 	@echo "Package requirements related targets"
 	@echo "requirements-selftests:  Install runtime and selftests requirements"
-	@echo "requirements-plugins:    Install plugins requirements"
 	@echo
 	@echo "Platform independent distribution/installation related targets:"
 	@echo "source:       Create single source package with commit info, suitable for RPMs"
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -223,6 +223,39 @@ def run(self):
...
+        directory = os.path.join(BASE_PATH, "optional_plugins")
+        for plugin in list(Path(directory).glob("*/setup.py")):
+            plugins_list.append(plugin.parts[-2])
+
+        if self.list or (not self.list and not self.install):
+            print("List of available plugins:\n ", "\n  ".join(plugins_list))
+            return
+
+        if self.install in plugins_list:
+            action = ["install", "--user"]
```
beraldoleal, line 251: Some may prefer to install over a virtual env, where `--user` might not be needed. Maybe we could let users/developers to decide if --user should be used or not. What do you think?

setup.py
```diff
@@ -223,6 +223,39 @@ def run(self):
...
+            parent_dir = os.path.join(directory, self.install)
+            run([sys.executable, "setup.py"] + action, cwd=parent_dir, check=True)
```
beraldoleal, line 253: With the current method,  we support `requirements.txt` inside each plugin, if any requirement is defined there, they will be installed. According to the documentation `install_requires` keyword is supposed  to be minimal and has some limitations and might not cover all the cases.

With this change, what is the plan?

Makefile
```diff
@@ -92,6 +93,13 @@ python_build: pip
 clean:
 	$(PYTHON) setup.py clean --all
 
+install-user:
+	$(PYTHON) setup.py install $(PYTHON_DEVELOP_ARGS)
```
beraldoleal, line 97: I'm  trying to understand why are you using DEVELOP_ARGS on a normal user install. IMO,  DEVELOP_ARGS should be used  when creating links. DEVELOP_ARGS is checking if the `develop` subcommand has `--user`  or not. And here you are using this with another subcommand.

Maybe I'm missing something.

beraldoleal (changes requested):
Hi @ana, thank you for the changes.  I left a few comments.

setup.py
ana: > Some may prefer to install over a virtual env, where `--user` might not be needed. Maybe we could let users/developers to decide if --user should be used or not. What do you think?

I'd expect people using a virtual-env to use the target 'install' instead of 'install-user' The idea of install-user is when you install directly in your local user.

setup.py
ana: My understanding is` install_requires` has all the requirements you need to run the plugin and you need `requirements.txt` for development. In the case of avocado optional plugin's, given it's the same base code, I'd expect to have all the plugins requirements for development in the main `requirements.txt' file.

Makefile
ana: > I'm trying to understand why are you using DEVELOP_ARGS on a normal user install. IMO, DEVELOP_ARGS should be used when creating links. DEVELOP_ARGS is checking if the `develop` subcommand has `--user` or not. And here you are using this with another subcommand.
> 
> Maybe I'm missing something.

No, this should use `--user`` directly.

setup.py
beraldoleal: But,  then you should query if --user is available on `install` instead of `develop`, right?

ana: > But, then you should query if --user is available on `install` instead of `develop`, right?

Sorry, I don't get what you mean.  This new target is unrelated to `develop`, it's aimed to install avocado for testing/trying to people who don't necessarily knows a lot of about Python. And they will install it directly in their system or more likely in a test Virtual Machine.

setup.py
beraldoleal: Unless I'm missing something, `requirements.txt` is not only for development. It can be used to specify *any* known-working environment  (i.e: dev, test, prod).

My only concern is this note on the documentation about install_requires:

> It is not considered best practice to use install_requires to pin dependencies to specific versions, or to specify sub-dependencies (i.e. dependencies of your dependencies). This is overly-restrictive, and prevents the user from gaining the benefit of dependency upgrades.
>
> Lastly, it’s important to understand that install_requires is a listing of “Abstract” requirements, i.e just names and version restrictions that don’t determine where the dependencies will be fulfilled from (i.e. from what index or source). 

About your comment:

> In the case of avocado optional plugin's, given it's the same base code, I'd expect to have all the plugins requirements for development in the main `requirements.txt' file.

TBH, I do have mix-feelings here. For instance, if someone is hacking the list plugin, not sure if makes sense to install all other requirements (specially if they are optional).

Anyway,  I just would  like to confirm the plans, since we don't have  requirements today, probably we could postpone  this discussion, and I don't see as blocking for this PR.

setup.py
beraldoleal: Sorry, ignore my last comment.

My question is: How someone using virtual env will install plugins? Sometimes system admins/users uses virtual env to isolate installs to avoid a wrong update will break some apps. IMO virtualenvs, are useful not only for development. Forcing users to have a VM just to try avocado is not the way to go.

ana: @beraldoleal I'm not forcing anyone to have a VM, I'm trying to add a new option to cater a type of users I feel are common. 
This doesn't change things for users of virtual envs, they can continue using the target `install` and in fact, they get a new option because they can use `python setup.py plugin` to install plugins!

beraldoleal: > This doesn't change things for users of virtual envs, they can continue using the target install and in fact, they get a new option because they can use python setup.py plugin to install plugins!

This is not what I'm getting. Back to my original comment: this new option (I'm talking about the setup.py Plugin command) is not working for virtual env users.

When I type `python3 setup.py --help-commands  | grep plugin`, I will assume this will work, but is not:
```
(avocado-py3.7) [local@laptop avocado]$ python3 setup.py --help-commands | grep plugin
  plugin            Handle plugins
(avocado-py3.7) [local@laptop avocado]$ python3 setup.py plugin --install=html
...

Please make the appropriate changes for your system and try again.
  File "setup.py", line 394, in <module>
    install_requires=['setuptools'])
  File "/home/local/src/venvs/avocado-py3.7/lib64/python3.7/site-packages/setuptools/__init__.py", line 145, in setup
    return distutils.core.setup(**attrs)
  File "/usr/lib64/python3.7/distutils/core.py", line 148, in setup
    dist.run_commands()
  File "/usr/lib64/python3.7/distutils/dist.py", line 966, in run_commands
    self.run_command(cmd)
  File "/usr/lib64/python3.7/distutils/dist.py", line 985, in run_command
    cmd_obj.run()
  File "setup.py", line 253, in run
    run([sys.executable, "setup.py"] + action, cwd=parent_dir, check=True)
  File "/usr/lib64/python3.7/subprocess.py", line 512, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['/home/local/src/venvs/avocado-py3.7/bin/python3', 'setup.py', 'install', '--user']' returned non-zero exit status 1.
```

As a Python contributor, the first thing that I will look is for the `setup.py`, ignoring the Makefile. I understand that some users will look first to Makefile targets, ignoring setup.py, but unless I'm missing something, we could do things a little bit more predictable/consistent here.

ana: @beraldoleal Could you please share all the steps? It works here with a small caveat: I have to setup PYTHONPATH, which I agree is not great, but I don't get that kind of error messages

beraldoleal: I tried on a fresh environment Fedora 33:

```
$ python3.7 -m venv venv
$ source venv/bin/activate
$ git clone -b plugin_update_makefile https://github.com/ana/avocado.git
$ cd avocado/
$  python3 setup.py plugin --install=html
```
And If I remove `--user` here, it works fine.

I you can't reproduce, I can try to investigate better, the real problem here.

ana: @beraldoleal Thanks! This is what I get https://gist.github.com/ana/c07ebceeedaba5ebb8f65d9351d74f87
Don't bother trying to reproduce yet, let's first sync the changes made to setup.py/Makefile in the last weeks and how we could finish this epic.

ana:
Closing, see https://github.com/avocado-framework/avocado/pull/4817

PR #4807, opened 2021-07-29
Resolver and Discoverer plugins initialization fix

richtja:
In the 43286b5fa3083fc0d3fb4381a4a40024d120eb81 the incompatibility with Setting plugin was introduced. Because of that, avocado doesn't respect plugin specific configuration. This commit adds `settings` variable import to the `plugin_interfaces.py` Because of that, the `circular import` bug is created when the plugins and Settings is initialized during avocado initialization. This will make this import local to specified plugins, which fix this problem.

### DIFF
--- a/avocado/core/plugin_interfaces.py
+++ b/avocado/core/plugin_interfaces.py
@@ -15,8 +15,6 @@
 
 import abc
 
-from .resolver import ResolverMixin
-
 
 class Plugin(abc.ABC):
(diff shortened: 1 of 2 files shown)


beraldoleal (approved):
LGTM, thanks.

willianrampazzo (approved):

clebergnu (approved):
LGTM, thanks!

PR #4813, opened 2021-07-30
nrunner: adds support to failfast (v3)

### LINKED ISSUE #3870: NRunner plugin: support failfast feature

beraldoleal:
have already started when an error occurs, because of that, failfast will
work here in a "best-effort" assumption. This change will move all
non-started tasks to the finished queue when failfast is enabled. Fixes #3870

### Changes from v2:

 - Using `self._state_machine._status_repo.result_stats` as source of truth
 - Assuming that each test suite has its own StateMachine, so no need to double-check
 
### Changes from v1:

 - Complete new implementation using state-machine (Thanks to @willianrampazzo)

### DIFF
--- a/avocado/core/task/statemachine.py
+++ b/avocado/core/task/statemachine.py
@@ -47,6 +47,52 @@ class TaskStateMachine:
                            self._ready, self._started])
         return not pending
 
+    async def abort(self, status_reason=None):
+        """Abort all non-started tasks.
+
+        This method will move all non-started tasks to finished with a specific
+        reason.
+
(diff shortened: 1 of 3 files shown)


richtja (approved):
LGTM

clebergnu (approved):
LGTM, thanks!

PR #4815, opened 2021-08-02
interfaces.py: Support get the interface details without IP version

PaulYuuu:
Currently, if the interface does have not an IP address,
"_get_interface_details" will return an empty list, so this is a problem
to set default version=4. Most of the time, we only need the properties
of the interface itself(e.g: get_mtu), only "get_ipaddrs" will use IP addresses in
interface.py.

### DIFF
--- a/avocado/utils/network/interfaces.py
+++ b/avocado/utils/network/interfaces.py
@@ -61,8 +61,10 @@ class NetworkInterface:
             raise NWException(msg)
         return "{}/ifcfg-{}".format(path, self.name)
 
-    def _get_interface_details(self, version=4):
-        cmd = "ip -{} -j address show {}".format(version, self.name)
+    def _get_interface_details(self, version=None):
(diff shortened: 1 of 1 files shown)


PaulYuuu:
For instance, if a interface is a source nic of the bridge:
```
$ ip -4 -j address show enp5s0
[]

$ ip -j link show enp5s0
[{"ifindex":2,"ifname":"enp5s0","flags":["BROADCAST","MULTICAST","UP","LOWER_UP"],"mtu":1500,"qdisc":"fq_codel","master":"switch","operstate":"UP","linkmode":"DEFAULT","group":"default","txqlen":1000,"link_type":"ether","address":"6c:b3:11:0f:99:e9","broadcast":"ff:ff:ff:ff:ff:ff"}]
```

beraldoleal (approved):
LGTM, thank you.

richtja (approved):
LGTM, thanks

PR #4818, opened 2021-08-02
nrunner: failfast: make failfast return AVOCADO_JOB_INTERRUPTED and nice message

beraldoleal:
It will make it consistent with the current runner returning
AVOCADO_JOB_INTERRUPTED, also this is adding a
message to the user about the failfast.

### DIFF
--- a/avocado/core/exceptions.py
+++ b/avocado/core/exceptions.py
@@ -155,6 +155,17 @@ class TestFail(TestBaseException, AssertionError):
     status = "FAIL"
 
 
+class TestFailFast(TestBaseException):
+
+    """
+    Indicates that the test has failed because failfast is enabled.
+
+    Should be thrown when a test has failed and failfast is enabled. This will
+    indicate that other tests will be skipped.
+    """
+    status = "SKIP"
+
+
 class TestWarn(TestBaseException):
 
     """
--- a/avocado/core/task/statemachine.py
+++ b/avocado/core/task/statemachine.py
@@ -3,6 +3,8 @@ import collections
 import multiprocessing
 import time
 
+from avocado.core.exceptions import TestFailFast
+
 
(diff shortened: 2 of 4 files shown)


clebergnu (changes requested):
There's one major problem introduced here: the indiscriminately writing to the "UI" log.   If you run:

```
avocado run --test-runner=nrunner --failfast --nrunner-max-parallel-tasks=1 --tap - -- /bin/true /bin/false /bin/true
```

You should *not* get the last line:

```
1..3
ok 1 /bin/true
not ok 2 /bin/false
Interrupting job (failfast).
```

The `runner` implementation avoid that by setting the failure reason to the job, and letting the human UI plugin handle that when it controls the stdout/stderr.

beraldoleal:
> There's one major problem introduced here: the indiscriminately writing to the "UI" log. If you run:
> 
> ```
> avocado run --test-runner=nrunner --failfast --nrunner-max-parallel-tasks=1 --tap - -- /bin/true /bin/false /bin/true
> ```
> 
> You should _not_ get the last line:
> 
> ```
> 1..3
> ok 1 /bin/true
> not ok 2 /bin/false
> Interrupting job (failfast).
> ```
> 
> The `runner` implementation avoid that by setting the failure reason to the job, and letting the human UI plugin handle that when it controls the stdout/stderr.

Thanks for the hint. I wasn't sure about it. Fixed on the force-push [here](https://github.com/avocado-framework/avocado/compare/114ae9042aacad261b442b486d7823608bcc760e..e7bfa4bbf636335044852a2a67f0f94fa4831e4a)

beraldoleal:
@clebergnu ping.

clebergnu (approved):
LGTM, thanks!

PR #4823, opened 2021-08-05
utils.download: fixes root logging

beraldoleal:
Since we are getting 'avocado.test' logger, let's use it, otherwise we
are writing to the root logger.


ana (approved):

richtja (approved):
LGTM, thanks

willianrampazzo (approved):
Thanks a lot for this change :)

PR #4827, opened 2021-08-05
[runner] fail early when -p and -m are used

willianrampazzo:
The current logic skips the fail when -p and -m command-line options are
used. Their use together is not supported.


beraldoleal (approved):
LGTM, thank you.

clebergnu (approved):
I can see a number of things wrong around this code, but unrelated to your change which is a fix indeed.  Thanks!

PR #4847, opened 2021-08-13
avocado.utils: limit printed durations to nanosecond precision

skotchandsoda:
We should limit printed elapsed times to nanosecond precision, i.e.
nine figures after the decimal point.

First, long decimal fractions are difficult to eyeball.  This
is annoying to read at a glance:

Command 'kill -19 1429623' finished with 0 after 0.0005124460049035922s

Whereas this is easier to read at a glance:

Command 'kill -19 1429623' finished with 0 after 0.000512446s

Second, the system call behind time.monotonic(), clock_gettime(2), has
a resolution of one nanosecond.  Additional figures are insignificant.

Signed-off-by: Scott Cheloha <cheloha@linux.ibm.com>

### DIFF
--- a/avocado/utils/debug.py
+++ b/avocado/utils/debug.py
@@ -40,7 +40,7 @@ def measure_duration(func):
             duration = time.monotonic() - start
             __MEASURE_DURATION[func] = (__MEASURE_DURATION.get(func, 0) +
                                         duration)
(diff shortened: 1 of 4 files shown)


beraldoleal (approved):
LGTM. Thank you.

ana (approved):
Thanks, this looks good to me.

willianrampazzo (approved):

PR #4862, opened 2021-08-17
[RFC] add support for some AVOCADO_* env variables to nrunner

willianrampazzo:
This adds support to the following environment variables for all nrunner test types:

AVOCADO_VERSION
AVOCADO_TEST_BASEDIR
AVOCADO_TEST_LOGDIR
AVOCADO_TEST_OUTPUTDIR

Although avocado instrumented tests already define these environment variables, the code here adds support for all the other types of tests.

Based on #4751 

### DIFF
--- a/avocado/core/nrunner.py
+++ b/avocado/core/nrunner.py
@@ -356,8 +356,36 @@ class BaseRunner(abc.ABC):
     Base interface for a Runner
     """
 
+    @staticmethod
+    def _get_avocado_version():
+        """Return the Avocado package version, if installed"""
+        version = "unknown.unknown"
+        if PKG_RESOURCES_AVAILABLE:
+            try:
+                version = pkg_resources.get_distribution(
+                    'avocado-framework').version
+            except pkg_resources.DistributionNotFound:
+                pass
+        return version
+
     def __init__(self, runnable):
         self.runnable = runnable
+
+        # set default Avocado env variables
+        #
+        # NOTE: these environment variables are redefined inside an Avocado
+        # instrumented test.
+        #
+        # this is a code duplication from avocado/core/version.py to avoid
+        # nrunner depending on other than python modules
+        os.environ['AVOCADO_VERSION'] = self._get_avocado_version()
+
+        # this is a code duplication from avocado/core/data_dir.py to avoid
+        # nrunner depending on other than python modules
+        os.environ['AVOCADO_TEST_BASEDIR'] = self.runnable.config.get(
+            'datadir.paths.base_dir', '')
+        os.environ['AVOCADO_TEST_LOGDIR'] = self.runnable.config.get(
+            'datadir.paths.logs_dir', '')
 
     @staticmethod
     def prepare_status(status_type, additional_info=None):


avocado/core/nrunner.py
```diff
@@ -359,6 +359,31 @@ class BaseRunner(abc.ABC):
...
+        # this is a code duplication from avocado/core/version.py to avoid
+        # nrunner depending on other than python modules
+        version = "unknown.unknown"
+        if PKG_RESOURCES_AVAILABLE:
+            try:
+                version = pkg_resources.get_distribution(
+                    'avocado-framework').version
+            except pkg_resources.DistributionNotFound:
+                pass
+        os.environ['AVOCADO_VERSION'] = version
```
beraldoleal, line 376: I won't mind if you split this code in a own method `def _get_avocado_version()`.

avocado/core/nrunner.py
```diff
@@ -359,6 +359,31 @@ class BaseRunner(abc.ABC):
...
+
+        # this is a code duplication from avocado/core/data_dir.py to avoid
+        # nrunner depending on other than python modules
+        os.environ['AVOCADO_TEST_BASEDIR'] = self.runnable.config.get(
+            'datadir.paths.base_dir', '')
+        os.environ['AVOCADO_TEST_LOGDIR'] = self.runnable.config.get(
+            'datadir.paths.logs_dir', '')
+        os.environ['AVOCADO_TEST_OUTPUTDIR'] = self.runnable.kwargs.get(
+            'AVOCADO_TEST_OUTPUT_DIR', '')
```
beraldoleal, line 385: Is this last line really needed ? I mean, all kwargs aren't being injected as variable envs?

willianrampazzo: > Is this last line really needed ? I mean, all kwargs aren't being injected as variable envs?

Indeed, this line is not needed.

avocado/core/nrunner.py
willianrampazzo: Ack!

avocado/core/nrunner.py
beraldoleal: What makes me think why we don't inject the others using the same method?

willianrampazzo: My idea with this RFC is to discuss if the code is in the right place and if it makes sense.

beraldoleal: Maybe something like:

```python
def setup_env_variables(self):
  self.runnable.kwargs.update({'AVOCADO_TEST_LOGDIR': self.runnable.config.get('datadir.paths.logs_dir'),
                               'AVOCADO_TEST_BASEDIR':  self.runnable.config.get('datadir.paths.base_dir'))
  # And add here the code of `setup_output_dir()`
```

avocado/core/nrunner.py
```diff
@@ -356,9 +356,37 @@ class BaseRunner(abc.ABC):
...
+        # NOTE: these environment variables are redefined inside an Avocado
+        # instrumented test.
+        #
+        # this is a code duplication from avocado/core/version.py to avoid
+        # nrunner depending on other than python modules
+        os.environ['AVOCADO_VERSION'] = self._get_avocado_version()
+
+        # this is a code duplication from avocado/core/data_dir.py to avoid
+        # nrunner depending on other than python modules
+        os.environ['AVOCADO_TEST_BASEDIR'] = self.runnable.config.get(
```
richtja, line 385: Hi @willianrampazzo, I think in here we should validate here if the directories exist and can be used in the runner environment. Because the environment of runner can be different from environment of avocado core. Or maybe we can limit this just for the `ProcessSpawner` where the environments are the same. What do you think?

willianrampazzo: About the verification, as far as I could check, the directories are always available, but on the local host, where Avocado is executed.

On the other hand, as you mentioned, some may not make sense as they are located in the host and not in the spawner environment. This was something I was expecting to discuss on https://github.com/avocado-framework/avocado/discussions/4751 and is a valid point.

So far, for me, it looks like we are porting something from the runner architecture that does not make much sense on the nrunner architecture, at least when porting it 1 to 1.

richtja: So maybe for now we can create warning in documentation that this can be used only with the `ProcessSpawner`.

willianrampazzo:
Closing in favor of #4923

PR #4866, opened 2021-08-18
Default log file for nrunner based tests

richtja:
This will save every nrunner messages from stdout and stderr to the `debug.log` with the proper tag on the beginning of each line, and it introduces the `debug.log` file as Default log file for nrunner.

Reference: #4804

### DIFF
--- a/avocado/core/messages.py
+++ b/avocado/core/messages.py
@@ -101,10 +101,13 @@ class StartMessageHandler(BaseMessageHandler):
         task_id = TestID.from_identifier(task.identifier)
         base_path = job.test_results_path
         task_path = os.path.join(base_path, task_id.str_filesystem)
+        logfile = os.path.join(task_path, DEFAULT_LOG_FILE)
         os.makedirs(task_path, exist_ok=True)
+        open(logfile, 'w').close()
         metadata = {'job_logdir': job.logdir,
                     'job_unique_id': job.unique_id,
                     'base_path': base_path,
+                    'logfile':  logfile,
                     'task_path': task_path,
                     'time_start': message['time'],
                     'actual_time_start': time.time(),
--- a/avocado/core/messages.py
+++ b/avocado/core/messages.py
@@ -17,6 +17,8 @@ import time
 
 from .test_id import TestID
 
+DEFAULT_LOG_FILE = 'debug.log'
+
 
 class BaseMessageHandler:
     """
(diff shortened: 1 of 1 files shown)


avocado/core/messages.py
```diff
@@ -157,6 +159,51 @@ def handle(self, message, task, job):
...
     """Base interface for resolving running messages."""
 
+    _tag = b''
+
+    def __init__(self):
+        self.line_buffer = b''
+
+    def _get_tagged_line(self, data, encoding=None):
+        """
+        It will add the tag in to each line of data as a prefix. If the data
```
beraldoleal, line 172: PEP257, please.

avocado/core/messages.py
```diff
@@ -281,7 +332,10 @@ class StderrMessageHandler(BaseRunningMessageHandler):
              'time': 18405.55351474}
     """
 
+    _tag = b'[stdout] '
```
beraldoleal, line 338: IIUC, s/stdout/stderr/g since this is the StderrMessageHandler?

avocado/core/messages.py
```diff
@@ -157,6 +159,51 @@ def handle(self, message, task, job):
...
+        The default log file is based on `DEFAULT_LOG_FILE` variable and every
+        line of log will be saved with prefix based on `_tag` variable
+
+        :param message: message from runner
+        :type message: dict
+        :param task: runtime_task which message is related to
+        :type task: :class:`avocado.core.nrunner.Task`
+        """
+        encoding = message.get('encoding', None)
+        data = self._get_tagged_line(message['log'], encoding)
```
beraldoleal, line 206: This method name is indicating that we are going to "get something", but in fact this method is also setting some object property (line 179, 187). I would suggest to refactor this.

beraldoleal, line 206: Also, `log` and `encoding` are both part of `message`. So maybe just pass the `message` to the method.

avocado/core/messages.py
```diff
@@ -157,6 +159,51 @@ def handle(self, message, task, job):
...
+
+        The default log file is based on `DEFAULT_LOG_FILE` variable and every
+        line of log will be saved with prefix based on `_tag` variable
+
+        :param message: message from runner
+        :type message: dict
+        :param task: runtime_task which message is related to
+        :type task: :class:`avocado.core.nrunner.Task`
+        """
+        encoding = message.get('encoding', None)
```
beraldoleal, line 205: None is not needed here.

avocado/core/messages.py
```diff
@@ -157,6 +159,51 @@ def handle(self, message, task, job):
...
+    _tag = b''
+
+    def __init__(self):
+        self.line_buffer = b''
+
+    def _get_tagged_line(self, data, encoding=None):
+        """
+        It will add the tag in to each line of data as a prefix. If the data
+        don't finish with the new line character, the last line is marked
+        as incomplete, and it is saved to the line buffer for next usage.
```
beraldoleal, line 174: This last part that I'm talking about on a following comment. Maybe we could let each method do only one single thing and the split/check for incomplete lines could be done in a separated method.

beraldoleal (changes requested):
Hi @richtja, thanks. I left a few comments for you.

avocado/core/messages.py
richtja: Ok, I will refactor this.

richtja:
@beraldoleal Thanks for your comments, I will prepare new version.

richtja:
Closing in favor of #4876

PR #4869, opened 2021-08-18
core.tapparser: non tap messages should be ignored

beraldoleal:
From the the specs:

"A harness must only read TAP output from standard output and not from standard
error. Lines written to standard output matching /^(not )?ok\b/ must be
interpreted as test lines. All other lines must not be considered test output."

### DIFF
--- a/avocado/core/tapparser.py
+++ b/avocado/core/tapparser.py
@@ -153,8 +153,6 @@ class TapParser:
             if line == '':
                 continue
 
-            yield self.Error('unexpected input at line %d' % (lineno,))
-
         if state == self._YAML:
             yield self.Error('YAML block not terminated (started on line %d)' % (yaml_lineno,))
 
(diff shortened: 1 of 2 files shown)


willianrampazzo (approved):
Easy fix :D

clebergnu (approved):
LGTM, thanks!

PS: It hurts me that the standard is so lax...  :/

eskultety:
> LGTM, thanks!
> 
> PS: It hurts me that the standard is so lax... :/

If it weren't there's a good chance the accompanying strings/outcome descriptions would be quite cryptic, IOW you'd have to refrain from using any mention of the words "skip,error,failure,etc." - for example, one of the strings in TCK says "NO_DOMAIN error was raised" and because of that the test was marked as ERROR. Another example was a test where some of the scenarios were unavailable because of lack of support/tooling and so the string said "#skip <...>" and that made the whole test marked as SKIP, even though it actually ran and provided output.

PR #4883, opened 2021-08-20
variants.yml: remove i386 in ubuntu

ana:
After updating the ubuntu versions at 11649747d3ef5834172bf2fe9c525a1927f3a6a1 the pre-release pipeline fails because ubuntu doesn't include i386, see https://github.com/avocado-framework/avocado/actions/runs/1148234681

### DIFF
--- a/selftests/pre_release/tests/vmimage.py.data/variants.yml
+++ b/selftests/pre_release/tests/vmimage.py.data/variants.yml
@@ -57,6 +57,7 @@ distro: !mux
   ubuntu:
     name: ubuntu
     !filter-out : /run/architectures/arm
+    !filter-out : /run/architectures/i386
     !filter-out : /run/architectures/ppc
     !filter-out : /run/architectures/ppc64
     ppc64le:


richtja (approved):

beraldoleal (approved):

willianrampazzo:
The interesting fact is that http://cloud-images.ubuntu.com/releases/releases/ do not have i386 for quite some time.

ana:
> The interesting fact is that http://cloud-images.ubuntu.com/releases/releases/ do not have i386 for quite some time.

Yes, the latest ubuntu version we were testing was 18.04 that still had i386. The problem was triggered when my commit included 20.10 and 21.04

willianrampazzo:
> > The interesting fact is that http://cloud-images.ubuntu.com/releases/releases/ do not have i386 for quite some time.
> 
> Yes, the latest ubuntu version we were testing was 18.04 that still had i386. The problem was triggered when my commit included 20.10 and 21.04

Ah, okay, I looked only on 20.10 and 21.04.

Issue #4739, opened 2021-06-28
Plugin robot requirements needs updating


ana:
The robot plugin (`avocado-framework-plugin-robot`) has a requirement on` 'robotframework<=3.1.2'`. This version of robotframework is old and needs fixing for Python 3.10 that are in more recent versions. 

This makes the selftests to fail on Python3.10:

```
Traceback (most recent call last):
  File "/home/ana/python310/bin/avocado-runner-robot", line 33, in <module>
    sys.exit(load_entry_point('avocado-framework-plugin-robot', 'console_scripts', 'avocado-runner-robot')())
  File "/home/ana/python310/bin/avocado-runner-robot", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/usr/lib64/python3.10/importlib/metadata/__init__.py", line 162, in load
    module = import_module(match.group('module'))
  File "/usr/lib64/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 992, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/home/ana/avocado/optional_plugins/robot/avocado_robot/__init__.py", line 22, in <module>
    from robot import run
  File "/home/ana/python310/lib/python3.10/site-packages/robotframework-3.1.2-py3.10.egg/robot/__init__.py", line 44, in <module>
    from robot.rebot import rebot, rebot_cli
  File "/home/ana/python310/lib/python3.10/site-packages/robotframework-3.1.2-py3.10.egg/robot/rebot.py", line 40, in <module>
    from robot.conf import RebotSettings
[...]

PR #4890, opened 2021-08-23
Updates for ci.yml and weekly.yml

ana:
CI/`ci.yml`: allow to run workflow manually

CI/`weekly.yml:` Duplicate the pipeline and make two different tests:
    * one identical to the one run in every PR but with development python
    * one without optional_plugin tests, only with "core" avocado.
Export AVOCADO_LOG_DEBUG and AVOCADO_CHECK_LEVEL properly.

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -13,12 +13,13 @@ jobs:
   latest-python:
     name: Linux with Python ${{ matrix.python-version }}
     runs-on: ubuntu-20.04
-
     strategy:
       matrix:
         # see list of available Python versions at https://github.com/actions/python-versions/blob/main/versions-manifest.json
         python-version: [3.10.0-rc.1]
       fail-fast: false
+    outputs:
+      job_status: ${{ job.status }}
 
     steps:
       - run: echo "Job triggered by a ${{ github.event_name }} event on branch is ${{ github.ref }} in repository is ${{ github.repository }}, runner on ${{ runner.os }}"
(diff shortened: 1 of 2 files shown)


ana:
For the record: the core avocado works with python 3.10, however there are some problems with the plugins. One of the problems is reported with https://github.com/avocado-framework/avocado/issues/4739 but it's not the only one.

richtja (approved):
LGTM, thanks

.github/workflows/weekly.yml
```diff
@@ -13,7 +13,6 @@ jobs:
   latest-python:
```
beraldoleal, line 13: Hi @ana I'm not sure if I understood the problems that this is solving.

Could you please explain more on the commit description?

ana: > Hi @ana I'm not sure if I understood the problems that this is solving.
> 
> Could you please explain more on the commit description?

Hi @beraldoleal 
I've updated the commit message, hopefully it's more clear now!

ana:
I forgot to put an example of how the pipeline will look after this change. You can see it here:
https://github.com/ana/avocado/actions/runs/1162553599

.github/workflows/weekly.yml
beraldoleal: Thanks @ana for explaining it.

After reading the description, what about instead of duplicating the pipelines if we set properly the artifacts so we can download log messages to better debug?

ana: > Thanks @ana for explaining it.
> 
> After reading the description, what about instead of duplicating the pipelines if we set properly the artifacts so we can download log messages to better debug?

Mostly 2 reasons:
- I want to know quickly without debugging if this something that requires quick attention (problem in core avocado) or not (problem in a plugin) 
- Looking at the log messages you don't realize immediately they're caused by a plugin, you need to run the test locally.

beraldoleal: About 1: IMO, we should pay attention not only on "core" issues. Due the Avocado architecture, some  plugins are essential. Sometimes, a problem with a specific plugin might need a "quick attention" as well.

About 2: I agree that debugging in Avocado is a area for improvements, but IIUC, error messages are giving us the right hints (see bellow):

```
Failed tests:
check-892-selftests/functional/test_streams.py:StreamsTest.test_custom_stream_and_level FAIL
check-1033-optional_plugins/varianter_yaml_to_mux/tests/test_unit.py:TestMultipleLoaders.test_multiple_loaders FAIL
```

In some cases, yes, it is part of the debug process to configure a local environment so we can reproduce the problem. 
But, if we had the artifacts, we could look at the stdout/stderrors on each of those tests without the need of running locally on a specific environment.

Since we don't have that yet, I did a local test and seems to me that:

892: the problem here is that robot framework is not ready for the python3.10 syntax.
1033: yaml.safe_loader() for some reason is returning a different dict type in some cases (collections.OrderedDict). We are already discussing possible solutions on #4891 

If you think that is valuable to split the runs here, I will not be against it, but seems to me that the logs are showing exactly the what are our problems. And the artifacts could help a lot here. IMO, even more than the split.

ana: > About 1: IMO, we should pay attention not only on "core" issues. Due the Avocado architecture, some plugins are essential. Sometimes, a problem with a specific plugin might need a "quick attention" as well.

ACK

> About 2: I agree that debugging in Avocado is a area for improvements, but IIUC, error messages are giving us the right hints (see bellow):
> 
> ```
> Failed tests:
> check-892-selftests/functional/test_streams.py:StreamsTest.test_custom_stream_and_level FAIL
> check-1033-optional_plugins/varianter_yaml_to_mux/tests/test_unit.py:TestMultipleLoaders.test_multiple_loaders FAIL
> ```
> 
> In some cases, yes, it is part of the debug process to configure a local environment so we can reproduce the problem.
> But, if we had the artifacts, we could look at the stdout/stderrors on each of those tests without the need of running locally on a specific environment.
> 
> Since we don't have that yet, I did a local test and seems to me that:
> 
> 892: the problem here is that robot framework is not ready for the python3.10 syntax.

Yes, this was reported at https://github.com/avocado-framework/avocado/issues/4739
However, when running `python selftests/check.py --disable-plugin-checks robot` like at https://github.com/avocado-framework/avocado/runs/3396962071  The issue is still there, so there is some any other problem (and it's very likely with another plugin)

> 1033: yaml.safe_loader() for some reason is returning a different dict type in some cases (collections.OrderedDict). We are already discussing possible solutions on #4891
> 
> If you think that is valuable to split the runs here, I will not be against it, but seems to me that the logs are showing exactly the what are our problems. And the artifacts could help a lot here. IMO, even more than the split.

I'll add keeping the artifacts. One week is enough?

beraldoleal: > Yes, this was reported at #4739
> However, when running `python selftests/check.py --disable-plugin-checks robot` like at https://github.com/avocado-framework/avocado/runs/3396962071 The issue is still there, so there is some any other problem (and it's very likely with another plugin)

`--disable-plugin-checks robot` is disabling the plugin completely or just not running those tests? Even that we are not checking it, the plugin is still loaded: try to run `avocado run /bin/true` on python3.10. You will the error is still there.

The test is not inside the plugin, the test is `selftests/functional/test_streams.py:StreamsTest.test_custom_stream_and_level` and it is still capturing the problem. So I think that the test worked just fine here capturing something that is not supposed to be there.

> I'll add keeping the artifacts. One week is enough?

Since we are going to have the manual check, I think so.

ana: > > Yes, this was reported at #4739
> > However, when running `python selftests/check.py --disable-plugin-checks robot` like at https://github.com/avocado-framework/avocado/runs/3396962071 The issue is still there, so there is some any other problem (and it's very likely with another plugin)
> 
> Is `--disable-plugin-checks robot` disabling the plugin completely or just not running those tests? Even that we are not checking it, the plugin is still loaded: try to run `avocado run /bin/true` on python3.10. You will see that the error is still there.
> 
> The test is not inside the plugin, the test is `selftests/functional/test_streams.py:StreamsTest.test_custom_stream_and_level` and it is still capturing the problem. So I think that the test worked just fine here capturing something that is not supposed to be there.

No, if you don't install the plugins with `python3 setup.py develop --user --skip-optional-plugins` (the job in this PR does that)  you don't get any issue with `selftests/functional/test_streams.py:StreamsTest.test_custom_stream_and_level`

I didn't make any remark before about the installation without plugins and I should have insisted on this. Please, see the comment about updating this PR below for more information.

ana:
@beraldoleal  I have force-pushed an update.
You can see how the pipeline will look now https://github.com/ana/avocado/actions/runs/1165772219

The first job is identical to the one we run in every PR but using python 3.10 (currently in development), if this job success, the pipeline stops there. And if this job fails, the job-result folder is stored as an artifact and a second job is started.
This second job has 2 differences, it install only avocado without the optional plugins ` python3 setup.py develop --user --skip-optional-plugins` and then run the tests without the plugins:
` python selftests/check.py --disable-plugin-checks golang --disable-plugin-checks html --disable-plugin-checks resultsdb --disable-plugin-checks result_upload --disable-plugin-checks robot --disable-plugin-checks varianter_cit --disable-plugin-checks varianter_pict --disable-plugin-checks varianter_yaml_to_mux`
This long line is needed, because there are some tests that needs the plugin inside `selftests/` and the` --disable-plugin-checks` option will disable it.
If this job fails,  the job-result folder is stored as an artifact.

beraldoleal (approved):

PR #4891, opened 2021-08-24
optional plugins: python3.10 uses collections.OrderedDict

beraldoleal:
Fixes one of the issue with python3.10 on our GH

### DIFF
--- a/optional_plugins/varianter_yaml_to_mux/tests/test_unit.py
+++ b/optional_plugins/varianter_yaml_to_mux/tests/test_unit.py
@@ -1,3 +1,4 @@
+import collections
 import copy
 import itertools
 import os
@@ -441,7 +442,7 @@ class TestMultipleLoaders(unittest.TestCase):
         self.assertEqual(type(treenode.children[0]), mux.MuxTreeNode)
         # equivalent to yaml.load("...", Loader=yaml.SafeLoader)
         plain = yaml.safe_load("foo: bar")
-        self.assertEqual(type(plain), dict)
+        self.assertIn(type(plain), [dict, collections.OrderedDict])
 
 
 class TestInternalFilters(unittest.TestCase):


ana:
I looked at this issue some time ago, I checked exactly the same fix but then when running only these lines in a venv of python3.10 I got this:

```
(python310) [anguerre@anguerre venv]$ python
Python 3.10.0rc1 (default, Aug  3 2021, 00:00:00) [GCC 11.2.1 20210728 (Red Hat 11.2.1-1)] on linux
>>> 
>>> import yaml
>>> plain = yaml.safe_load("foo: bar")
>>> type(plain)
<class 'dict'>

```
So something else is going there.

ana:
Before I forget, as pointed in IRC, I think this is a problem with the plugins requirements. So we'd need to review all the requirements first, update the versions when required and then check if we need changes likes this one.

beraldoleal:
Hi @ana, why do you think that is a plugin requirement problem? Any hint on that direction?

I did some investigation before sending the PR, and I suspect that is something related to the internal python/setuptools implementation in Python3.10 (that explains why we are not getting in a normal console test).

Here is a possible reproducible. Can you please do the following test in your environment?

1. Remove avocado:

```
$ make uninstall && make clean
```

2.  replace your setup.py with a minimal working setup.py for `avocado-runner-python-unittest`:
```python
#!/bin/env python3

from setuptools import setup

if __name__ == '__main__':
    setup(name='avocado-framework',
          version="90.0",
          packages=['avocado'],
          entry_points={
              'console_scripts': [
                  'avocado-runner-python-unittest = avocado.core.nrunner:main',
                  ],
              },
          python_requires='>=3.6',
          install_requires=['setuptools'])
```

With that, avocado binaries are not being installed. Just the  `avocado-runner-python-unittest` entry point.

3. Install avocado:

```
$ python3 setup.py install --force # this is to force recreating the entry-point scripts
``` 

4. Create this basic test in /tmp/foo.py

```
import unittest
import yaml


class TestMultipleLoaders(unittest.TestCase):

    def test_multiple_loaders(self):
        plain = yaml.safe_load("foo: bar")
        self.assertEqual(type(plain), dict)

if __name__ == '__main__':
    unittest.main()
```

5. Execute the test using the entry point and then using directly:

```
$ avocado-runner-python-unittest task-run -i foo -k python-unittest -u /tmp/test.py:TestMultipleLoaders.test_multiple_loaders
$ python3.10 avocado/core/nrunner.py  task-run -i foo -k python-unittest -u /tmp/test.py:TestMultipleLoaders.test_multiple_loaders
```

What are you getting ? Here, when using the entrypoint it will return the error that we are hunting.

ana (approved):
Hi  @beraldoleal  thanks for the detailed steps! 
In my comments from yesterday, I mixed this with PR#4890 and I thought this was for fixing the robot plugin. I agree it's a valid fix for the `varianter_yaml_to_mux` plugin. Thanks.

PR #4912, opened 2021-08-28
nrunner: fix nrunner test_interrupt processes

beraldoleal:
nrunners does not support yet this kind of operation because of its
architecture.


clebergnu (approved):
LGTM, thanks!

PR #4914, opened 2021-08-30
prerelease.yml: update artifacts naming

ana:
Give different names to each job artifacts so they are all kept without overwritting.

Do not save any artifact for the job check-rpm-available-copr, since this job only install the packages and doesn't create any output.

Update path for the artifacts generated by the job avocado-deployment-copr, this job is running the tests inside a container and the path changes.

An example of this pipeline being forced to fail  (running `/usr/bin/false`  as last step)  can be browsed at:
https://github.com/ana/avocado/actions/runs/1182268924
The artifacts are listed in the bottom of that page.

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -29,7 +29,7 @@ jobs:
         if: failure()
         uses: actions/upload-artifact@v2
         with:
-          name: job-results
+          name: job-results-make-check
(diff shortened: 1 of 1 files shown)


willianrampazzo (approved):

beraldoleal (approved):
Thank you!!!

PR #4931, opened 2021-09-03
CI/ci.yml: save more artifacts

ana:
Save the logs if the tests fail.
Keep the Python Eggs as artifacts in case we want to test them.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,13 @@ jobs:
           AVOCADO_LOG_DEBUG: "yes"
           AVOCADO_CHECK_LEVEL: "1"
         run: make check
+      - name: Archive failed tests logs
+        if: failure()
+        uses: actions/upload-artifact@v2
+        with:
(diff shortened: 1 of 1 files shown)


clebergnu (approved):
LGTM, thanks!

beraldoleal (approved):
LGTM, thanks.

But, just out of curiosity, the reason that you are using retention-days = 1 is because of space limitations? Or something else?

willianrampazzo (approved):

ana:
> LGTM, thanks.
> 
> But, just out of curiosity, the reason that you are using retention-days = 1 is because of space limitations? Or something else?

This is run every time we submit or force-push  a PR, so I didn't see the point to keep it longer.

beraldoleal:
> > LGTM, thanks.
> > But, just out of curiosity, the reason that you are using retention-days = 1 is because of space limitations? Or something else?
> 
> This is run every time we submit or force-push a PR, so I didn't see the point to keep it longer.

I got you. Thanks.

PR #4951, opened 2021-09-16
check.py/setup.py: fix --disable-plugin-checks

ana:
Update `--disable-plugin-checks` to allow disabling several plugins with the same option
    
When running from the selftests/check.py, it needs to be used as:
`      python3 selftests/check.py ... --disable-plugin-checks html robot`
From the setup.py test command as:
`      python3 setup.py test ... --disable-plugin-checks "html robot"`

 
Fixes: #4904
References: https://github.com/avocado-framework/avocado/issues/4904

A second commit to force always using the default (n)runner.

### DIFF
--- a/selftests/check.py
+++ b/selftests/check.py
@@ -201,9 +201,6 @@ def parse_args():
     parser.add_argument('--job-api',
                         help='Run job API checks',
                         action='store_true')
-    parser.add_argument('--disable-plugin-checks',
-                        help='Disable checks for a plugin (by directory name)',
-                        action='append', default=[])
     parser.add_argument('--nrunner-interface',
                         help='Run selftests/functional/test_nrunner_interface.py',
                         action='store_true')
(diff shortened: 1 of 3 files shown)


selftests/check.py
```diff
@@ -219,7 +216,9 @@ def parse_args():
     parser.add_argument('--optional-plugins',
                         help='Run optional_plugins/*/tests/',
                         action='store_true')
-
+    parser.add_argument('--disable-plugin-checks',
+                        help='Disable checks for one or more plugins (by directory name)',
+                        nargs='+', action='append', default=[])
```
beraldoleal, line 221: Something that has caught my attention here is the fact that we have two syntaxes depending on the file that we are calling. Besides that, having `--disable-plugin-checks html robot` might be problematic even when using quotes. Positional arguments are parsed as part of this same list.

I understand that this is the default way that argparse works when using nargs. In fact, I wrote this same recommendation at BP001 to use space as separator. However, this might be problematic in some situations (for instance, if we need to add a positional argument at the end of the argument).

So my question is: how hard is to use the same syntax for both? Have you considered that? I.e: 

```
--disable-plugin-checks html,robot
```

ana: @beraldoleal yeah, I hit into some of the the limitations of argparse. 
Personally, I don't think we have two different syntaxes. I considered the syntax  you proposed because you mentioned it in the bug report, that was the first implementation and I changed it because I liked more this one... I don't mind changing it back thought.

beraldoleal: I like this one too, but unfortunately there are those limitations. I would prefer the comma, if you don't as a huge problem.

ana: @beraldoleal I have updated it. Please tell me what you think now.

beraldoleal (approved):
Sounds good to me, thank you.

willianrampazzo (approved):

PR #4955, opened 2021-09-17
[utils] Hosts: quit remote_session when RemoteHost exits

cforno12:
Hosts utils ensures that we quit the session object upon exit

### DIFF
--- a/avocado/utils/network/hosts.py
+++ b/avocado/utils/network/hosts.py
@@ -124,6 +124,15 @@ class RemoteHost(Host):
         self.password = password
         self.remote_session = self._connect()
 
+    def __enter__(self):
+        if not self.remote_session:
+            self._connect()
+        return self
+
+    def __exit__(self, _type, _exc_value, _traceback):
+        if self.remote_session:
+            self.remote_session.quit()
+
     def _connect(self):
         session = Session(host=self.host,
                           port=self.port,


avocado/utils/network/hosts.py
```diff
@@ -124,6 +124,9 @@ def __init__(self, host, username, port=22, key=None, password=None):
         self.password = password
         self.remote_session = self._connect()
 
+    def __exit__(self):
+        self.remote_session.quit()
```
beraldoleal, line 128: This is a nice idea. But since self._connect can raise exceptions, self.remove_session can be empty.

So I would add a basic check here:

```python
if self.remote_sesion:
    self.remote_session.quit()
```

beraldoleal (changes requested):
Hi @cforno12, thanks for that, nice to have.

Just one small comment.

avocado/utils/network/hosts.py
cforno12: Good point. Thanks for the quick review!

beraldoleal (approved):
LGTM, thank you.

beraldoleal (commented):
This LGTM, but by looking now, have you tough about having a __enter__ as well?

cforno12:
> This LGTM, but by looking now, have you tough about having a **enter** as well?

Now that you brought it up we could split `__init__` and `__enter__`, it would look something like this:
``` python
def __init__(self, host, username, port=22, key=None, password=None):
        super().__init__(host)
        self.port = port
        self.username = username
        self.key = key
        self.password = password
        self.remote_session = None

    def __enter__(self):
        self._connect()
        return self
```
similar to what ssh.py is doing. Would you like me to add it to this PR as well?

beraldoleal:
> > This LGTM, but by looking now, have you tough about having a **enter** as well?
> 
> Now that you brought it up we could split `__init__` and `__enter__`, it would look something like this:
> 
> ```python
> def __init__(self, host, username, port=22, key=None, password=None):
>         super().__init__(host)
>         self.port = port
>         self.username = username
>         self.key = key
>         self.password = password
>         self.remote_session = None
> 
>     def __enter__(self):
>         self._connect()
>         return self
> ```
> 
> similar to what ssh.py is doing. Would you like me to add it to this PR as well?

Yes, that was the same direction that I hat. Please, go a head.

avocado/utils/network/hosts.py
```diff
@@ -122,7 +122,15 @@ def __init__(self, host, username, port=22, key=None, password=None):
         self.username = username
         self.key = key
         self.password = password
-        self.remote_session = self._connect()
+        self.remote_session = None
```
beraldoleal, line 125: Before this patch, self.remote_session it was initialized during object initialization. __enter__ is not called when object is being initialized. This change here is not necessary.

avocado/utils/network/hosts.py
```diff
@@ -122,7 +122,15 @@ def __init__(self, host, username, port=22, key=None, password=None):
...
+
+    def __enter__(self):
+        self._connect()
```
beraldoleal, line 128: And here, only tries to connect if not connected.

avocado/utils/network/hosts.py
```diff
@@ -122,7 +122,15 @@ def __init__(self, host, username, port=22, key=None, password=None):
...
+        return self
+
+    def __exit__(self):
```
beraldoleal, line 131: Also, exit signature is missing some arguments.

beraldoleal (changes requested):

avocado/utils/network/hosts.py
cforno12: Oh! TIL. Thank you

beraldoleal (approved):
LGTM.

PraveenPenguin (approved):
LGTM

PR #4956, opened 2021-09-17
[utils] network/interfaces: add VLAN functionality for an interface

cforno12:
Add fuctionality to retrieve a dict of vlans for a given interface, add a vlan
interface given a vlan tag number, delete a vlan interface given a vlan tag
number, and delete all vlans associated with an interface


cforno12:
Re-posting since I noticed that my patch was reverted recently.

clebergnu (approved):

PR #4963, opened 2021-09-21
examples/jobs: add an example for a custom exec-test

willianrampazzo:
This adds an example of a custom exec-test using the Job API. The main objective is to document how args and kwargs are used.


wainersm (approved):
Thanks for working on this example, @willianrampazzo !

If I had it before I would have struggled to understand the API :)

beraldoleal (approved):

beraldoleal (approved):

PR #4964, opened 2021-09-22
Introducing issues templates for bugs and feature requests

beraldoleal:
When creating bug reports inside this repository would be nice to have a
template to follow. This will save developers time when debugging the 
problem, since most of our requests don't have a reproducible today. The 
same for new features requests, since some features are small and don't
need a entire blueprint.


beraldoleal:
This was made by the github web interface that is the reason of the new branch and missing the signedoff. Closing and opening a proper PR.

PR #4991, opened 2021-09-29
Fixes tmpdirs leftovers (v2)

beraldoleal:
Hi, this is a work on top of cleber's PR. And basically is fixing #4392.

@clebergnu I decided to not deprecate core/data_dir.py:get_tmp_dir()
because I saw a few calls for it on avocado-vt side. We need to decide
if the cleanup part will be delegated the same way we do with job
(enter/exit). So this could be another PR/discussion.

### Changes from v1:

- removed one commit content with non-intentional removal;
- fixed typo on commit message
- removing the keep_tmp workarounds

### DIFF
--- a/avocado/core/job.py
+++ b/avocado/core/job.py
@@ -147,6 +147,7 @@ class Job:
         self.logfile = None
         self.tmpdir = None
         self.__keep_tmpdir = True
+        self._base_tmpdir = None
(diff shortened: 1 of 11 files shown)


clebergnu (approved):
LGTM, many thanks!

PR #4993, opened 2021-09-29
utils: fixes root logger

beraldoleal:
As discussed before, we are using a new approach where each module is logging in its own namespace.

### DIFF
--- a/avocado/utils/cpu.py
+++ b/avocado/utils/cpu.py
@@ -35,6 +35,8 @@ VENDORS_MAP = {
     'ibm': (rb"POWER\d", rb"IBM/S390", ),
 }
 
+LOG = logging.getLogger(__name__)
+
 
 class FamilyException(Exception):
     pass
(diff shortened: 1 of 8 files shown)


richtja (changes requested):
Hi @beraldoleal, nice work thanks, the code LGTM I just found another occurrence in [process utils](https://github.com/avocado-framework/avocado/blob/master/avocado/utils/process.py#L297)

richtja:
I am not sure if it's related to this PR because this is just about utils, but during the review I found out that we are manipulating with root logger also in [avocado-instrumented runner](https://github.com/avocado-framework/avocado/blob/master/avocado/core/test.py#L607)

beraldoleal:
> Hi @beraldoleal, nice work thanks, the code LGTM I just found another occurrence in [process utils](https://github.com/avocado-framework/avocado/blob/master/avocado/utils/process.py#L297)

Yes, I left process for a second round because I need to discuss with you a few things there.

beraldoleal:
> I am not sure if it's related to this PR because this is just about utils, but during the review I found out that we are manipulating with root logger also in [avocado-instrumented runner](https://github.com/avocado-framework/avocado/blob/master/avocado/core/test.py#L607)

I don't think is related to this pr, but for sure we need to fix this as well.

beraldoleal:
Thank you @richtja. Closing this and creating a new PR (not a v2 since I'm using a different approach).

Issue #4997, opened 2021-09-30
logging: use specific namespaces for each utility library


beraldoleal:
**Is your feature request related to a problem? Please describe.**

Currently all our libraries are using 'avocado.test' as a logger namespace instead of a specific namespace. Changing this is not trivial, since we need to make sure that test logs are being correctly populated. Also today some users are expecting this as a possible command to see the tests results:

```
$ avocado --show=test run foo.py 
```

**Describe the solution you'd like**

Since we can't be sure that this will always be called from a test we could use a specific namespace for each library for instance:

```python
LOG = logging.getLogger(__name__)
```

This will make debug easier since it will be converted to something like  `avocado.utils.process`. However we need to make sure test's logs are being correctly populated.

PR #4998, opened 2021-09-30
Fixes on logging system

beraldoleal:
We are trying to avoid using the root logger since July. But we are still missing some bits.
This PR is fixing some of those issues.

### DIFF
--- a/avocado/core/job.py
+++ b/avocado/core/job.py
@@ -198,7 +198,7 @@ class Job:
 
     def __start_job_logging(self):
         # Enable test logger
-        fmt = ('%(asctime)s %(module)-16.16s L%(lineno)-.4d %('
+        fmt = ('%(asctime)s %(name)s L%(lineno)-.4d %('
                'levelname)-5.5s| %(message)s')
         test_handler = output.add_log_handler(LOG_JOB,
                                               logging.FileHandler,
--- a/avocado/core/runners/utils/messages.py
+++ b/avocado/core/runners/utils/messages.py
@@ -202,6 +202,7 @@ def start_logging(config, queue):
     log.setLevel(log_level)
     log.propagate = False
     root_logger = logging.getLogger('avocado')
+    root_logger.setLevel(log_level)
     root_logger.addHandler(log_handler)
     output.LOG_UI.addHandler(RunnerLogHandler(queue, 'stdout'))
 
--- a/avocado/core/runners/utils/messages.py
+++ b/avocado/core/runners/utils/messages.py
@@ -193,7 +193,7 @@ def start_logging(config, queue):
     """
     log_level = config.get('job.output.loglevel', logging.DEBUG)
     log_handler = RunnerLogHandler(queue, 'log')
-    fmt = ('%(asctime)s %(module)-16.16s L%(lineno)-.4d %('
+    fmt = ('%(asctime)s %(name)s L%(lineno)-.4d %('
            'levelname)-5.5s| %(message)s')
     formatter = logging.Formatter(fmt=fmt)
     log_handler.setFormatter(formatter)
--- a/avocado/core/test.py
+++ b/avocado/core/test.py
@@ -601,10 +601,11 @@ class Test(unittest.TestCase, TestData):
         self.log.addHandler(self._file_handler)
         self.log.propagate = False
 
-        # add the test log handler to the root logger so that
-        # everything logged while the test is running, for every
-        # logger, also makes its way into the test log file
-        logging.root.addHandler(self._file_handler)
+        # Adding the the test log FileHandler to the Avocado's root logger so
+        # that everything logged while the test is running, for every logger,
+        # also makes its way into the test log file
+        root = logging.getLogger('avocado')
+        root.addHandler(self._file_handler)
 
         stream_fmt = '%(message)s'
         stream_formatter = logging.Formatter(fmt=stream_fmt)
--- a/avocado/utils/crypto.py
+++ b/avocado/utils/crypto.py
@@ -17,6 +17,8 @@ import io
 import logging
 import os
 
+LOG = logging.getLogger('avocado.test')
+
 
 def hash_file(filename, size=None, algorithm="md5"):
     """
@@ -42,8 +44,8 @@ def hash_file(filename, size=None, algorithm="md5"):
     try:
         hash_obj = hashlib.new(algorithm)
     except ValueError as detail:
-        logging.error('Returning "None" due to inability to create hash '
-                      'object: "%s"', detail)
+        LOG.error('Returning "None" due to inability to create hash '
+                  'object: "%s"', detail)
         return None
 
     with open(filename, 'rb') as file_to_hash:
@@ -52,7 +54,7 @@ def hash_file(filename, size=None, algorithm="md5"):
                 chunksize = size
             data = file_to_hash.read(chunksize)
             if len(data) == 0:
-                logging.debug("Nothing left to read but size=%d", size)
+                LOG.debug("Nothing left to read but size=%d", size)
                 break
             hash_obj.update(data)
             size -= len(data)
--- a/avocado/utils/software_manager/backends/rpm.py
+++ b/avocado/utils/software_manager/backends/rpm.py
@@ -6,7 +6,8 @@ from ... import path as utils_path
 from ... import process
 from .base import BaseBackend
 
-log = logging.getLogger('avocado.utils.software_manager')
+
+log = logging.getLogger('avocado.test')
 
 
 class RpmBackend(BaseBackend):
--- a/avocado/utils/software_manager/backends/rpm.py
+++ b/avocado/utils/software_manager/backends/rpm.py
@@ -6,8 +6,7 @@ from ... import path as utils_path
 from ... import process
 from .base import BaseBackend
 
-
-log = logging.getLogger('avocado.test')
+LOG = logging.getLogger('avocado.test')
 
 
 class RpmBackend(BaseBackend):
@@ -77,7 +76,7 @@ class RpmBackend(BaseBackend):
         :param software_components: log in a format suitable for the
                                     SoftwareComponent schema
         """
-        log.debug("Listing all system packages (may take a while)")
+        LOG.debug("Listing all system packages (may take a while)")
 
         if software_components:
             cmd_format = "rpm -qa --qf '%s' | sort"
@@ -127,7 +126,7 @@ class RpmBackend(BaseBackend):
         :rtype: bool
         """
         if not os.path.isfile(file_path):
-            log.warning('Please provide proper rpm path')
+            LOG.warning('Please provide proper rpm path')
             return False
 
         nodeps = "--nodeps " if no_dependencies else ""
@@ -138,7 +137,7 @@ class RpmBackend(BaseBackend):
             process.system(cmd)
             return True
         except process.CmdError as details:
-            log.error(details)
+            LOG.error(details)
             return False
 
     @staticmethod
@@ -150,7 +149,7 @@ class RpmBackend(BaseBackend):
         :returns: whether the verification was successful
         :rtype: bool
         """
-        logging.info("Verifying package information.")
+        LOG.info("Verifying package information.")
         cmd = "rpm -V " + package_name
         result = process.run(cmd, ignore_status=True)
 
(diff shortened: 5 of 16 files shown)


avocado/core/runners/utils/messages.py
```diff
@@ -202,6 +202,7 @@ def start_logging(config, queue):
     log.setLevel(log_level)
     log.propagate = False
     root_logger = logging.getLogger('avocado')
+    root_logger.setLevel(log_level)
```
pevogam, line 205: Hmmm nice catch

avocado/core/test.py
```diff
@@ -601,10 +601,11 @@ def _start_logging(self):
...
 
-        # add the test log handler to the root logger so that
-        # everything logged while the test is running, for every
-        # logger, also makes its way into the test log file
-        logging.root.addHandler(self._file_handler)
+        # Adding the the test log FileHandler to the Avocado's root logger so
+        # that everything logged while the test is running, for every logger,
+        # also makes its way into the test log file
+        root = logging.getLogger('avocado')
+        root.addHandler(self._file_handler)
```
pevogam, line 608: Do you think this would affect VT tests too? If so I can perhaps do some testing here even tomorrow.

avocado/utils/crypto.py
```diff
@@ -17,6 +17,8 @@
 import logging
 import os
 
+LOG = logging.getLogger('avocado.test')
```
pevogam, line 20: Is there a preferred way to access this logger? For instance, how does this type of accessing it compare to
```
from avocado.core.output import LOG_JOB
```
that some utilities have done before?

pevogam (commented):
Hi @beraldoleal, I have a few minor questions regarding these changes (perhaps just stuff I don't understand) and cand do some testing with these changes tomorrow in case you are interested.

avocado/core/test.py
beraldoleal: Probably it will. Avocado-vt is using either the root logger or `__name__` (which is starting with `avocado-vt` instead `avocado.`). I was planning to use `"avocado" + __name__` as the logger name on all modules there to fix that.

If you could check if you are missing any content inside the log files with this change it would be great.

avocado/utils/crypto.py
beraldoleal: Probably we should use the global one `LOG_JOB` when we are sure that  'avocado.test' is our namespace. However I decided to keep as it is because I think we should change this in the future for `__name__` as described here #4997.

avocado/core/test.py
richtja: For more context, the VT tests should be affected by this change only on the legacy runner, because the nrunner doesn't use this method.

pevogam: Thanks for narrowing this down @richtja, as we are already trying to use the nrunner and the serial runner is planned already as legacy I will do more of the testing once we have more changes in place and the ones concerning the nrunner.

richtja (commented):
Hi @beraldoleal, this LGTM, there are just problems with selftests.

beraldoleal:
@richtja fixed the tests, lets see if GH gets happy.

richtja (approved):
LGTM, thanks

pevogam (approved):

pevogam (changes requested):

avocado/core/job.py
```diff
@@ -198,7 +198,7 @@ def __exit__(self, _exc_type, _exc_value, _traceback):
 
     def __start_job_logging(self):
         # Enable test logger
-        fmt = ('%(asctime)s %(module)-16.16s L%(lineno)-.4d %('
+        fmt = ('%(asctime)s %(name)-25.25s L%(lineno)-.4d %('
```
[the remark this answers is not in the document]
beraldoleal: We can have multiple modules in different packages using the same name, so having only the module name is not so helpful.

Also, multiple modules are using the 'avocado.test' namespace and we are using that information for figuring out what should be printed and where.

However, I agree with you that cutting this is not good. I was also bothered thinking that a huge name could be trimmed, and you just gave me an example.

IMO sounds strange to have a fixed length size inside the log file and I'm assuming that is to get a "nice" output if we would like to see only the "message". But this could be done using a cut by the "|" (pipe) instead of a cut by an exact column.

So a see two alternatives: 1) increase the length (which will be wasted on a small namespace); 2)  remove the namespace fixed length, showing the entire namespace. This will make the message to not be aligned as it is today, but is the way logs are and I don't see much as a problem.

Let me know what you think, but also keep in mind that I would like to provide soon a better way to users configure the formatters as well. (Post LTS)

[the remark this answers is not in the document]
pevogam: I am not sure of the options still, because long namespaces will then shift every log line further to the right (in a fix manner or not), occupying large screen real estate only printing identical initial information like `avocado.test` (assuming most of the logging happens from there which is mostly the case for all practical uses). Also considering plugin developers are down the supply chain, they would inevitably introduce new name spaces if they want to be able to debug and these name spaces will inevitably be subspaces of `avocado.test` or similar, leading to long names for everybody.

But allowing for full namespace length seems like the lesser evil and the only solution that could satisfy both points of view. I guess future formatting functionality could help even though there still isn't any optimal choice for the people that use it.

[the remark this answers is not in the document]
pevogam: Generic log filtering by `avocado.test` keywords also won't be very useful since we will end up with 95% if not all lines (rendering the filtering useless) but a good point for the `name` is that we could all then do log-postprocessing on top and there is at least a marginal improvement of filterability here.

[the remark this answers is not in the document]
beraldoleal: I will consider the comments here as acceptance of the proposal. Thanks.

avocado/core/test.py
```diff
@@ -601,10 +601,11 @@ def _start_logging(self):
...
         self.log.propagate = False
 
-        # add the test log handler to the root logger so that
-        # everything logged while the test is running, for every
-        # logger, also makes its way into the test log file
-        logging.root.addHandler(self._file_handler)
+        # Adding the the test log FileHandler to the Avocado's root logger so
+        # that everything logged while the test is running, for every logger,
+        # also makes its way into the test log file
+        root = logging.getLogger('avocado')
```
clebergnu, line 607: Nitpick: it would be much clearer if the new code did not use the `root` variable for the logger containing the `avocado` namespace.  This is about removing the use of the root logger but then mentioning `root` dilutes that point.

avocado/utils/software_manager/backends/rpm.py
```diff
@@ -6,7 +6,8 @@
 from ... import process
 from .base import BaseBackend
 
-log = logging.getLogger('avocado.utils.software_manager')
+
+log = logging.getLogger('avocado.test')
```
clebergnu: For consistency.: s/log/LOG/

beraldoleal: Probably this is a cache issue, it was already fixed.

avocado/utils/software_manager/backends/rpm.py
```diff
@@ -6,8 +6,7 @@
 from ... import process
 from .base import BaseBackend
 
-
-log = logging.getLogger('avocado.test')
+LOG = logging.getLogger('avocado.test')
```
clebergnu, line 9: Changes look good, but I don't understand why this is separate from the previous commit.

avocado/core/job.py
```diff
@@ -198,7 +198,7 @@ def __exit__(self, _exc_type, _exc_value, _traceback):
 
     def __start_job_logging(self):
         # Enable test logger
-        fmt = ('%(asctime)s %(module)-16.16s L%(lineno)-.4d %('
+        fmt = ('%(asctime)s %(name)s L%(lineno)-.4d %('
```
clebergnu, line 201: AFAICT we're losing the aligment of the "headers" (prefixes) of the log entries here, right?

avocado/utils/software_manager/backends/rpm.py
beraldoleal: The first one I changed only the namespace, to keep consistent with others. The second one I fixed the usage of 'logging' , since it was bug it self (some parts where using LOG and others logging). I could have replaced s/log/LOG/ on the first yes, But I would still prefer to keep both commits.

selftests/functional/test_output.py
```diff
@@ -254,40 +254,50 @@ def _check_output(path, exps, name):
...
         test.save()
         result = process.run("%s run --job-results-dir %s --disable-sysinfo "
-                             "--test-runner=runner "
-                             "--json - -- %s" % (AVOCADO, self.tmpdir.name, test))
+                             "--json - -- %s" % (AVOCADO,
+                                                 self.tmpdir.name,
+                                                 test))
         res = json.loads(result.stdout_text)
-        joblog = res["debuglog"]
+        joblog = res["tests"][0]["logfile"]
```
clebergnu, line 261: This deserves a change in variable name too, otherwise it'd be misleading.

selftests/functional/test_output.py
```diff
@@ -254,40 +254,50 @@ def _check_output(path, exps, name):
...
         exps = [b"[stdout] top_print", b"[stdout] top_stdout",
-                b"[stderr] top_stderr", b"[stdout] top_process",
-                b"[stdout] init_print", b"[stdout] init_stdout",
-                b"[stderr] init_stderr", b"[stdout] init_process",
```
clebergnu, line 264: I'd recommend that, instead of making the `process.run()` output silent with regards to *this* test expectations, we keep checking it, but as a more explicit check.  Example:

```diff
diff --git a/selftests/functional/test_output.py b/selftests/functional/test_output.py
index 9f01268ee..aa24ff5f4 100644
--- a/selftests/functional/test_output.py
+++ b/selftests/functional/test_output.py
@@ -70,7 +70,8 @@ class OutputTest(Test):
         print("init_print")
         sys.stdout.write("init_stdout\\n")
         sys.stderr.write("init_stderr\\n")
-        process.run("/bin/echo init_process")
+        self.log.info("[process.run() stdout] %s",
+                      process.run("/bin/echo init_process").stdout_text)
 
     def test(self):
         print("test_print")
@@ -261,7 +262,7 @@ class OutputTest(TestCaseTmpDir):
         exps = [b"[stdout] top_print", b"[stdout] top_stdout",
                 b"[stderr] top_stderr", b"[stdout] top_process",
                 b"[stdout] init_print", b"[stdout] init_stdout",
-                b"[stderr] init_stderr", b"[stdout] init_process",
+                b"[stderr] init_stderr", b"[process.run() stdout] init_process",
                 b"[stdout] test_print", b"[stdout] test_stdout",
                 b"[stderr] test_stderr", b"[stdout] test_process"]
         _check_output(joblog, exps, "job.log")
```

This would work with both runners, and would provide a way forward for a more explicit API.

avocado/core/runners/utils/messages.py
```diff
@@ -202,6 +202,7 @@ def start_logging(config, queue):
     log.setLevel(log_level)
```
clebergnu, line 202: I understand being grateful :smile:  but I guess this commit message was intended for a GitHub comment instead of a commit message... ?

avocado/core/job.py
beraldoleal: Yes, we are. Unless I'm missing something, I would prefer to have full namespace here instead of trying to set a fixed length (that will make us waste some columns in case of a short namespace).

I understand the usability of having those fixed, but although logs are made for humans I don't thing we should see those as UI. Usually we use ':' or '|' to separate this and make the parsers life easier. If we are providing a tool or a command to show those logs without the headers, then we could easily split by '|' here.

avocado/utils/software_manager/backends/rpm.py
clebergnu: OK, fair enough.

avocado/core/job.py
clebergnu: OK, given that we're being more precise about where a log entry comes from, I think it's a fair trade-off.

clebergnu (changes requested):
Overall, looks good. I believe I spotted a couple of small things that need changing.

selftests/functional/test_output.py
beraldoleal: IIUC, are you also recommending do the same with the method `def test()` and `def __del__()`? I mean: all process.run()' s output, should be inside a self.log?

clebergnu: > IIUC, are you also recommending do the same with the method `def test()` and `def __del__()`? I mean: all process.run()' s output, should be inside a self.log?

Correct, that's the proposal.

pevogam (approved):
I have redone the testing and I guess the problem I reported is on the Avocado VT side, so this looks good now on my side.

selftests/functional/test_output.py
beraldoleal: @clebergnu I'm just reading properly this test, I don't think we should expect that log output is going to be part of stdout (as the test name suggests). `self.log()` is not redirecting to stdout. This should be in a different method. So I will make some changes.

beraldoleal:
Thank you all. Closing in favor of a v2.

PR #5002, opened 2021-10-04
[utils] network/interface: check if interface is a bonding device

cforno12:
Added a method to interface.py to check if the interface is a bonding device. It
checks to see if /proc/net/bonding/<interface_name> is a valid file. Returns
True if it is, returns False otherwise.

### DIFF
--- a/avocado/utils/network/interfaces.py
+++ b/avocado/utils/network/interfaces.py
@@ -477,6 +477,22 @@ class NetworkInterface:
             log.debug(msg)
             return False
 
+    def is_bond(self):
+        """Check if interface is a bonding device.
+
+        This method checks if the interface is a bonding device or not.
+
+        rtype: bool
+        """
+        cmd = 'cat /proc/net/bonding/{}'.format(self.name)
+        try:
+            run_command(cmd, self.host)
+            return True
+        except Exception as ex:
+            msg = "{} is not a bond device. {}".format(self.name, ex)
+            log.debug(msg)
+            return False
+
     def remove_cfg_file(self):
         """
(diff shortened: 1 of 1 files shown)


avocado/utils/network/interfaces.py
```diff
@@ -477,6 +477,22 @@ def is_available(self):
...
+    def is_bond(self):
+        """Check if interface is a bonding device.
+
+        This method checks if the interface is a bonding device or not.
+
+        rtype: bool
+        """
+        cmd = 'cat /proc/net/bonding/{}'.format(self.name)
+        try:
+            run_command(cmd, self.host)
```
beraldoleal, line 489: You are not concerned about the content here, because of that I would recommend to check if the file exist and is a file using the python api for that, also use as property since is a question. Something like this:

```python
@property
def is_bond(self):
    return os.path.isfile('/proc/net/bonding/{}'.format(self.name))
```

beraldoleal (changes requested):

avocado/utils/network/interfaces.py
cforno12: I agree, we are not concerned with the content, but I would like to check if the interface is a bond interface on the remote side as well, hence `run_command`.

beraldoleal: Humm I see. In this case I will make my final comments.

beraldoleal (approved):

beraldoleal:
Approved, but this should be part of #93. Since we are in stabilization period, only fixing critical bugs.

Issue #5029, opened 2021-10-12
Use Debian images from cloud.debian.org


ana:
I was working in changing the class `DebianImageProvider`  to use directly `current-$VERSION`. However, this is now marked as old and doesn't include images from the lastest debian 11 (bullseye) .  We should migrate avocado to fetch the images from https://cdimage.debian.org/images/cloud/

The **generic** images there should run in any environment using cloud-init, for e.g. OpenStack.

PR #5032, opened 2021-10-13
utils/vmimage: download directly latest Debian image

ana:
Download directly the latest image for a given Debian version.
This is done by downloading the the image pointed by the
`current-$VERSION` symlink.

Update `variants.yml ` accordingly.

---
Important note: **this is for 92.0 if we decide it's a good idea to include it.**  Eventually (if) we do a new release, `variants.yml `  will need to be updated with the Debian versions. With this change, it'll continue to work unless https://cdimage.debian.org/cdimage/openstack/ is retired.

If we don't accept this PR for 92.0,  I'll close it because for post-92.0 I'm sending a new PR migrating to http://cloud.debian.org (https://github.com/avocado-framework/avocado/issues/5029 )

### DIFF
--- a/avocado/utils/vmimage.py
+++ b/avocado/utils/vmimage.py
@@ -304,8 +304,7 @@ class DebianImageProvider(ImageProviderBase):
 
     name = 'Debian'
 
-    def __init__(self, version='[0-9]+.[0-9]+.[0-9]+.*', build=None,
-                 arch=DEFAULT_ARCH):
+    def __init__(self, version='[0-9]+', build=None, arch=DEFAULT_ARCH):
         # Debian uses 'amd64' instead of 'x86_64'
         if arch == 'x86_64':
             arch = 'amd64'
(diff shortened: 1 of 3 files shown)


ana:
> 1. what happens to existing code that points to a valid/current version (say I wrote a test that asks for `9.13.28`).  Have you checked if the API will continue to take this and give the user the current version?  The `version='[0-9]+` regex makes me believe that it will require the user to change tests.

No,  `9.13.28` won't work.  I expect the user to be already used to change the tests often given the debian openstack images are changing every 2-3 weeks. So this would be, hopefully, the last change and the user should get a meaningful message about this. 

> 2. With the "loss" of extra version information, and the use of _files_ named such as `debian-9-openstack-amd64.qcow2`, won't those files be updated while the user expects (and has on cache, and knows of its hash) a previous version?

From my POV, the user always will want to use the latest version of the *image* (and not of Debian because all the images belong to the same version of Debian).

> And a minor concern:
> 
> 1. The way "by location" _asset_ caching currently works is by hashing the base location (usually the directory) of where the files are hosted.  This change will invalidate the currently cached files, even if they are the same _files_ because of the location change.  This is unavoidable because of the current asset implementation, but we should keep it in mind.
>
> I'll give this further testing (other than just thinking about the implications) and I'll let you know of the outcome.

Keep me posted. As said in the initial message of this commit, I'm fine if this kind of change is not suitable for the release. And it's better to focus in how we want things to work with the move to cloud.debian.org ( https://github.com/avocado-framework/avocado/issues/5029)

ana:
Closing this, we won't include it.

PR #5044, opened 2021-10-19
Debian: fix packaging errors

amikan:
After checking avocado deb package with lintian tool several issues occurred:
```
$ lintian -I ../avocado_91.0_amd64.changes --no-tag-display-limit
E: avocado source: source-is-missing optional_plugins/html/avocado_result_html/templates/bootstrap.min.js
E: avocado source: source-is-missing optional_plugins/html/avocado_result_html/templates/datatables.min.js
E: avocado source: source-is-missing optional_plugins/html/avocado_result_html/templates/jquery.min.js
W: avocado source: unused-build-dependency-on-cdbs
W: avocado source: package-needs-versioned-debhelper-build-depends 10
W: avocado source: changelog-should-mention-nmu
W: avocado source: source-nmu-has-incorrect-version-number 91.0
W: avocado source: ancient-standards-version 3.8.4 (released 2010-01-27) (current is 4.3.0)
I: avocado source: testsuite-autopkgtest-missing
W: avocado source: debian-watch-file-in-native-package
E: avocado: description-starts-with-package-name
I: avocado: description-synopsis-might-not-be-phrased-properly "Avocado is a framework for fully automated testing."
E: avocado: extended-description-is-empty
I: avocado: package-contains-documentation-outside-usr-share-doc usr/lib/python3/dist-packages/avocado/etc/avocado/scripts/job/post.d/README
I: avocado: package-contains-documentation-outside-usr-share-doc usr/lib/python3/dist-packages/avocado/etc/avocado/scripts/job/pre.d/README
W: avocado: binary-without-manpage usr/bin/avocado
W: avocado: binary-without-manpage usr/bin/avocado-runner
W: avocado: binary-without-manpage usr/bin/avocado-runner-avocado-instrumented
W: avocado: binary-without-manpage usr/bin/avocado-runner-dry-run
W: avocado: binary-without-manpage usr/bin/avocado-runner-exec-test
W: avocado: binary-without-manpage usr/bin/avocado-runner-noop
W: avocado: binary-without-manpage usr/bin/avocado-runner-python-unittest
W: avocado: binary-without-manpage usr/bin/avocado-runner-requirement-asset
W: avocado: binary-without-manpage usr/bin/avocado-runner-requirement-package
W: avocado: binary-without-manpage usr/bin/avocado-runner-sysinfo
W: avocado: binary-without-manpage usr/bin/avocado-runner-tap
W: avocado: binary-without-manpage usr/bin/avocado-software-manager
```
[...]

PR #5048, opened 2021-10-20
HTML plugin: update JS template engines

amikan:
Minified JS and CSS files cannot be considered as real source files.
That's why we should also add original Bootstrap, JQuery and DataTables sources.
Because it is not easy to find original versions that were used to create minified ones, it's better to just bump to the latest.

I also need an assist of someone using HTML plugin in real test cases to proof this change is correct.

This patchset is required to pass lintian testing in Debian.


amikan:
We need to have full version as 'source', and we need to have minified version for installing to target (for faster loading and caching).
The second option is to have only full version and use some tool like 'minify' to create minified version during the build.

richtja:
Closing this based on the [discussion](https://github.com/avocado-framework/avocado/pull/5044#issuecomment-1905772227) in #5044

PR #5097, opened 2021-11-05
remove nrunner references when using command line

beraldoleal:
There is no need for those, since nrunner is the default one.


willianrampazzo (approved):

clebergnu (approved):
Ah, I missed those on e59a2e78183748c30f19ae742cbf0afafedd487b. LGTM, Thanks!

PR #5107, opened 2021-11-09
ci.yml: Add experimental checks + Makefile.gh [v2]

ana:
Re-implementation of https://github.com/avocado-framework/avocado/pull/4918

Add a couple of experimental checks:
   * codespell to complement pycheck's spellchecker
   * bandit to find common security issues in Python code
    
Both will output the problems found without any action.
    
The actions are run intentionally from a new makefile `Makefile.gh` instead of using pre-made GitHub Actions.
This allows reproducing the checks locally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -188,3 +188,26 @@ jobs:
         path: /home/runner/work/avocado/avocado/dist/
         retention-days: 1
     - run: echo "🥑 This job's status is ${{ job.status }}."
+
+  experimental-checks:
+    name: Experimental checks
+    runs-on: ubuntu-20.04
+    steps:
+      - run: echo "Job triggered by a ${{ github.event_name }} event on branch is ${{ github.ref }} in repository is ${{ github.repository }}, runner on ${{ runner.os }}"
+      - name: Check out repository code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Run Codespell Check
+        run: make -f Makefile.gh codespell
+        continue-on-error: True
+      - name: Run bandit check
+        run: make -f Makefile.gh bandit
+        continue-on-error: True
+      - name: Save bandit output as artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: bandit-results
+          path: /home/runner/work/avocado/avocado/bandit-output.txt
+          retention-days: 1
+      - run: echo "🥑 This job's status is ${{ job.status }}."
--- a/Makefile.gh
+++ b/Makefile.gh
@@ -1,0 +1,14 @@
+# This Makefile contains targets used by GitHub Actions
+
+all:
+	@echo
+	@echo "This file contains targets used by GitHub Actions."
+	@echo
+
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -188,3 +188,26 @@ jobs:
...
+    name: Experimental checks
+    runs-on: ubuntu-20.04
+    steps:
+      - run: echo "Job triggered by a ${{ github.event_name }} event on branch is ${{ github.ref }} in repository is ${{ github.repository }}, runner on ${{ runner.os }}"
+      - name: Check out repository code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Run Codespell Check
+        run: make -f Makefile.gh codespell
```
beraldoleal, line 202: Since the makefile is inside the same directory, what about avoiding the .gh extension so we don' t need the -f?

Are you planning to have multiple Makefiles inside this directory?

Makefile.gh
```diff
@@ -0,0 +1,14 @@
+# This Makefile contains targets used by GitHub Actions
+
+all:
+	@echo
+	@echo "This file contains targets used by GitHub Actions."
+	@echo
+
+codespell:
+	pip install codespell
```
beraldoleal, line 9: Have you considered adding codespell and bandit to our `requirements-dev.txt`?

beraldoleal (commented):
Thank you @ana. LAGTM, just a few comments for you.

.github/workflows/ci.yml
ana: `Makefile.gh` is in the root directory of avocado, we can't remove the .gh extension because we already have a Makefile there.
 I'm planning to have multiple Makefile in the root directory. A casual contributor will only need to care for the main `Makefile` and we'll have the extra Makefiles for development purposes. I considered adding them in a subdirectory, but it was add extra complexity in the rules (they need to be run from the root directory) for little gain.

beraldoleal: My mistake I tough this was inside this directory. Lets go this way, but I think having multiple Makefiles inside the root might be confusing.

Makefile.gh
beraldoleal: I got it. The only "minor problem" I see with this approach is this _might_ be installing using a different method of what the developer choose depending on the user/env installing it. It is not respecting `PYTHON_DEVELOP_ARGS` the same way other targets in our main Makefile are.

So if I do the following commands:

```
$ make requirements-plugins
$ make develop
$ make -f Makefile.gh codespell
```

The first  two commands are using the `--user` option if available while the last is not, this _might_ result in different requirements being installed in different places. Maybe not a huge problem, but some might get confused. I'm also not saying that we need to fix this here in this PR, just saying out of loud so we can keep that in mind.

I will create another issue for tracking this.

beraldoleal: I know. But still relevant the location of the requirements.

beraldoleal (approved):

clebergnu (approved):
LGTM, thanks!

Issue #4935, opened 2021-09-03
Avocado's next runner hanging indefinitely at some tests


pevogam:
This is a report of an issue that I have experienced for some time but didn't share until I removed LXC container spawners and other functionality that could have been the reason for it. Right now I am using simple process spawners (therefore cannot run in parallel but could rule out causes) and still experience this so it is time to report it. This issue will also gradually be updated with more information I obtain from some diagnostics later on.

The issue is unfortunately quite fatal for test runs, namely a test run could hang indefinitely at a given (VT) test without any timeouts or other failure taking place to be able to at least go on and finish the test runs with fewer casualties. Simple stracing of the child process:
```
 206621 pts/4    Sl+    3:19              \_ /usr/bin/python3 /usr/bin/avocado-runner-avocado-vt task-run -i 411-mis1-normal.nongui.backup[...] -j 4d18092ffffcad098f48e6d76044d856d3fb8b5e -k avocado-vt -u normal.nongui.backup[...] -c {} name=normal.nongui.backup[...] shortname=normal.nongui.backup[...] [other-params] -s 127.0.0.1:8888
```
reveals a problem with a lock:
```
# strace -p 206621
strace: Process 206621 attached
futex(0x7fec6d7b5000, FUTEX_WAIT_BITSET|FUTEX_CLOCK_REALTIME, 0, NULL, FUTEX_BITSET_MATCH_ANY
```
Some gdb-based debugging shows we are stuck at a logging message (disclaimer: extracted from another occurrence where I stored it):
```
(gdb) py-list
  90        def _make_methods(self):
  91            self.acquire = self._semlock.acquire
  92            self.release = self._semlock.release
  93    
  94        def __enter__(self):
 >95            return self._semlock.__enter__()
  96    
  97        def __exit__(self, *args):
  98            return self._semlock.__exit__(*args)
  99    
 100        def __getstate__(self):
(gdb) py-bt
Traceback (most recent call first):
  File "/usr/lib64/python3.9/multiprocessing/synchronize.py", line 95, in __enter__
    return self._semlock.__enter__()
  File "/usr/lib64/python3.9/multiprocessing/queues.py", line 377, in put
    with self._wlock:
  File "/usr/lib/python3.9/site-packages/avocado/core/runners/utils/messages.py", line 158, in emit
    self.queue.put(self.message.get(msg))
  File "/usr/lib64/python3.9/logging/__init__.py", line 952, in handle
    self.emit(record)
[...]

PR #5112, opened 2021-11-09
s/SimpleQueue/Queue/ just for instrumented runner

beraldoleal:
SimpleQueue it is a simplified Queue type, very close to a locked Pipe.
Because of the way it is implemented, signals and GC can cause deadlocks.
Replacing it with a Queue it will return a process shared queue
implemented using a pipe and a few locks/semaphores. This is not
replacing all SimpleQueues by Queues, instead I'm reaplacing only here
because of the issue #4935. Both API (for the cases used here) are the
same, so even with a more complete API there is no significant change
here.

Fixes ##4935.

### DIFF
--- a/avocado/core/runners/avocado_instrumented.py
+++ b/avocado/core/runners/avocado_instrumented.py
@@ -93,7 +93,7 @@ class AvocadoInstrumentedTestRunner(nrunner.BaseRunner):
     def run(self):
         yield messages.StartedMessage.get()
         try:
-            queue = multiprocessing.SimpleQueue()
+            queue = multiprocessing.Queue()
             process = multiprocessing.Process(target=self._run_avocado,
                                               args=(self.runnable, queue))
 


beraldoleal:
FIY: I have 2 tests failing locally:


So far,  I couldn't find a relationship with this change. I will let the CI run to see if is something with my env.

```
check-1021-optional_plugins/varianter_yaml_to_mux/tests/test_functional.py:MultiplexTests.test_run_mplex_passtest FAIL
check-1028-optional_plugins/varianter_yaml_to_mux/tests/test_functional.py:MultiplexTests.test_mux_inject FAIL
```

beraldoleal:
CI errors are different from local errors. Although I cannot reproduce the errors individually, I believe it is related to this PR. (I still don't know how)

pevogam:
Hi @beraldoleal sorry I didn't yet find the time to trace the progress on this. Is this PR closed because it is not working or was there some other reason?

PR #5119, opened 2021-11-11
Moving asset warn message to info

PraveenPenguin:
Moving warn to info as while running misc test repo in every
run where we depend on using asset library it flash warning
and some time that is misleading on test front as we used test
warn tag for atual Test warn

### DIFF
--- a/avocado/utils/asset.py
+++ b/avocado/utils/asset.py
@@ -317,8 +317,7 @@ class Asset:
         If asset_hash is None then will consider a valid asset.
         """
         if asset_hash is None:
-            LOG.warning("No hash provided. Cannot check the asset file"
-                        " integrity.")
+            LOG.info("No hash provided. Cannot check the asset file integrity.")
             return True
 
         hash_path = cls._get_hash_file(asset_path)


PraveenPenguin:
@willianrampazzo as we are hitting this every case so will do  this change , please have review and let me your throught

avocado/utils/asset.py
```diff
@@ -317,8 +317,7 @@ def _has_valid_hash(cls, asset_path, asset_hash=None):
         If asset_hash is None then will consider a valid asset.
         """
         if asset_hash is None:
-            LOG.warning("No hash provided. Cannot check the asset file"
-                        " integrity.")
+            LOG.info("No hash provided. Cannot check the asset file integrity.")
```
beraldoleal, line 320: Actually, the right way to avoid mixing Avocado log messages with test's logs, would be to change the namespace of `LOG` here to `__name__` , instead of using `avocado.test`. This goes on the same line has the recent changes on avocado-vt side. There is an open issue for that: #4997.

I just created a PR (#5121) with just this small change, please, could you try that one and let us know if it will solve your problem?

beraldoleal (changes requested):

avocado/utils/asset.py
PraveenPenguin: @beraldoleal thanks nice approach there , closing this one

PR #5121, opened 2021-11-11
utils.asset: use __name__ as logger namespace

beraldoleal:
This is part of bigger effort to better organize our log messages. Also
a proper alternative to #5119. More to come here, but I just would like
to test if this will fix the issue reported there.


willianrampazzo (approved):

beraldoleal:
Looks like packit has some issues right now.

PraveenPenguin (approved):
wow , good approach, working for me   , LGTM as well

Issue #4990, opened 2021-09-28
[HTML plugin] params not showing up on hint box


willianrampazzo:
**Describe the bug**
When running tests with `-p`, the params are not added to the hint box on the HTML output.

**Steps to reproduce**
`avocado run -p sleep_length=2 examples/tests/sleeptest.py`

**Expected behavior**
When opening the HTML output and the mouse over the test name, it should display the `sleep_length` param.

**Current behavior**
Right now, the hint box display `No params`.

**System information (please complete the following information):**
 - Avocado version: 91.0
 - Avocado installation method: dev upstream clone

PR #5123, opened 2021-11-11
Convert test parameters to variants in tasks

richtja:
After the 050bc33 the suite will raise an error when the variants and test parameters are used together. With this in mind, there is no difference between one variant and test parameters from the task perspective. Let's add test parameters to test tasks as a variant. In the future, we can discuss if the `variant` is the right variable name in tasks for holding data about parameters.

This change will fix the issue with representation of parameters in result plugins.

Reference: #4990


clebergnu (approved):
LGTM, thanks!

PR #4961, opened 2021-09-21
 Add github action for the release test plan [v2]

ana:
Add a new GitHub Action that will help with the release.
This Action will:
- update all the VERSION files and the spec file
- update and build the new version at readthedocs.org
- build and update the packages (tar.gz and wheel) at pypi.org
- build and publish for ach python version the egg packages  in GitHub Fix:#4692

For making some of this actions, it's required to set up four tokens in github:
 `RELEASE_TOKEN` to allow update the repository avocado-framework/avocado
 `RTD_TOKEN` to update readthedocs
 `PYPI_USER` and `PYPI_PASSWD` to allow twine update pypi.org

Update the release testplan to the use the new Action.

References:
https://github.com/avocado-framework/avocado/issues/3456
https://github.com/avocado-framework/avocado/issues/4692

Signed-off-by: Ana Guerrero Lopez <anguerre@redhat.com>

PR #5127, opened 2021-11-12
Add github action for the release test plan [v3] 

ana:
This is a v3 of  https://github.com/avocado-framework/avocado/pull/4961
It uses a makefile for the actions (when it makes sense) to allow reproducing things locally. 
I have to say I find the action is hard to read now.

----
Add a new GitHub Action that will help with the release.
This Action will:
- update all the VERSION files and the spec file
- update and build the new version at readthedocs.org
- build and update the packages (tar.gz and wheel) at pypi.org
- build and publish for ach python version the egg packages  in GitHub Fix:#4692

For making some of this actions, it's required to set up four tokens in github:
 `RELEASE_TOKEN` to allow update the repository avocado-framework/avocado
 `RTD_TOKEN` to update readthedocs
 `PYPI_USER` and `PYPI_PASSWD` to allow twine update pypi.org

Update the release testplan to the use the new Action.

References:
https://github.com/avocado-framework/avocado/issues/3456
https://github.com/avocado-framework/avocado/issues/4692

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,89 @@
+# must do before using this action:
+# - set up RELEASE_TOKEN in secrets of the repository avocado-framework/avocado
+# - set up RTD_TOKEN in secrets to update readthedocs
+# - set up two tokens for twine: PYPI_USER and PYPI_PASSWD
+
+name: Release tests
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Release version'
+        required: true
+        default: '0.0'
+      devel_name:
+        description: 'Developer Name'
+        required: true
+        default: 'Avocado Developer'
+      devel_mail:
+        description: 'Developer mail'
+        required: true
+        default: 'avocado@redhat.com'
+      rtd_project:
+        description: 'readthedocs project name'
+        required: true
+        default: 'avocado-framework'
+
+jobs:
+
+  release:
+    name: Release pipeline
+    runs-on: ubuntu-latest
+    container:
+      image: fedora:34
+
+    steps:
+      - name: install required packages
+        run:  dnf -y install rpmdevtools git python3-pip make
+      - uses: actions/checkout@v2
+        with:
+          persist-credentials: false
+          fetch-depth: 0
+      - name: Update VERSION files and python-avocado.spec
+        run: |
+          export VERSION=${{ github.event.inputs.version }}
+          export DEVEL_NAME=${{ github.event.inputs.devel_name }}
+          export DEVEL_MAIL=${{ github.event.inputs.devel_mail }}
+          make -f Makefile.gh propagate-version
+          make -f Makefile.gh release-update-spec
+      - name: Commit files and tag
+        run: |
+          export VERSION=${{ github.event.inputs.version }}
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,119 @@
...
+        run:  dnf -y install rpmdevtools git python3-pip make
+      - uses: actions/checkout@v2
+        with:
+          persist-credentials: false
+          fetch-depth: 0
+      - name: Update VERSION files and python-avocado.spec
+        run: |
+          export VERSION=${{ github.event.inputs.version }}
+          export DEVEL_NAME=${{ github.event.inputs.devel_name }}
+          export DEVEL_MAIL=${{ github.event.inputs.devel_mail }}
```
beraldoleal, line 46: Hi @ana I'm trying to understand why those exports are needed here (here and in other parts of this file).

I did some tests using the `env` section and exporting directly the variables. It seems to be working, including for multiple steps:


```yaml
jobs:

  release:
    name: Release pipeline
    runs-on: ubuntu-latest
    container:
      image: fedora:34
    env:
      DEVEL_NAME: ${{ github.event.inputs.devel_name }}
      DEVEL_MAIL: ${{ github.event.inputs.devel_mail }}

    steps:
      - name: install required packages
        run:  dnf -y install rpmdevtools git python3-pip make
      - uses: actions/checkout@v2
        with:
          persist-credentials: false
          fetch-depth: 0
      - name: Update VERSION files and python-avocado.spec
        run: |
          set | grep DEVEL
          make test
      - name: Commit files and tag 
        run: |
          set
          make test
```

This example is here: https://github.com/beraldoleal/foobartest/

What I'm missing?

.github/workflows/release.yml
```diff
@@ -0,0 +1,119 @@
...
+          name: wheel
+          path: /__w/avocado-test/avocado-test/PYPI_UPLOAD/
+          retention-days: 3
+      - name: Upload to pypi
+        env:
+          TWINE_USERNAME: '${{ secrets.PYPI_USER }}'
+          TWINE_PASSWORD: '${{ secrets.PYPI_PASSWD }}'
+        run: |
+          export TWINE_USERNAME=$TWINE_USERNAME
+          export TWINE_PASSWORD=$TWINE_PASSWORD
```
beraldoleal, line 88: Same here: with my previous examples it was not needed the export. Probably I'm missing something.

beraldoleal (changes requested):
Hi @ana, thank you for this. 

LAGTM, however I left one small comment that might help with the readability aspect that you mentioned. Maybe I'm just missing something.

beraldoleal (approved):
@clebergnu, because of the @ana TZ and the release deadline, feel free to merge this, since my comments are not critical.

@ana take your time to reply my comments, no need to care about timing. We can see this next week.

clebergnu (approved):
LGTM, thanks!

.github/workflows/release.yml
ana: @beraldoleal I had the feeling that could be changed in the way you mentioned, but I didn't have time to test properly, I focused in checking the secret tokens weren't showed in the github action logs.
I found another problem with the variables so I'll include this change as well.

PR #5130, opened 2021-11-15
release.yml: updates for setting up variables and move readthedocs step

ana:
This is a follow up of https://github.com/avocado-framework/avocado/pull/5127 with some updates after testing.

`release.yml`: use jobs.env for setting up variables
    
This makes the pipeline easier to read.

Rename TOKEN to TOKEN_RTD to make it clearer this token is for readthedocs.org
    
A couple of minor fixes:
    - use python 3.10.0 because github action don't recognise 3.10
    - use `${{github.workspace}}` variable instead of hardcode a path

`release.yml`: move readthedocs step to run later
    
readthedocs needs a couple of seconds to refresh it copy of the git repository. If we try to update readthedocs immediately after tagging the new release, it'll fail.
    

This pipeline can be seen in action at https://github.com/ana/avocado-test/runs/4211607135?check_suite_focus=true
After a commit adding "fake" release notes for 93.0 
Even if the steps are all in **green**, "upload to pypi" is made to fail on purpose.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,6 +31,13 @@ jobs:
     runs-on: ubuntu-latest
     container:
       image: fedora:34
+    env:
+      VERSION: ${{ github.event.inputs.version }}
+      DEVEL_NAME: ${{ github.event.inputs.devel_name }}
+      DEVEL_MAIL: ${{ github.event.inputs.devel_mail }}
+      PYTHON: /usr/bin/python3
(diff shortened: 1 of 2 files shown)


beraldoleal (approved):

richtja (approved):

willianrampazzo (approved):

clebergnu (approved):
LGTM, thanks!

Issue #5099, opened 2021-11-08
Allow utils.process.run log into a custom logger


beraldoleal:
**Is your feature request related to a problem? Please describe.**
Currently utils.process.run() is draining stdout|stderr output to the `LOG` global variable. Which will get combined with any `avocado.test` namespace. While most of the this execution came from a test execution, since this is a library, there is no guarantee this is a test. Ideally the logger here should be `__name__` and users could be able to customize the logger here.

**Describe the solution you'd like**
```python
import logging
from avocado.utils import process

LOG = logging.getLogger(__name__)

process.run("/path/to/command", log=LOG)
```
By default logs will be redirected to the module logger, unless the caller specify an alternative logger.

**Describe alternatives you've considered**
none.

**Additional information**
Please, keep in mind this should affect only stdout/stderr of the process.run() outputs. Others loggers inside this module, should keep using the default logger, since it has Avocado internal messages.

PR #5131, opened 2021-11-15
Custom logger for `utils.process.run`

richtja:
This PR adds a new attribute `log` to the `utils.process.run`
method. This will help users to set their own loggers for running
processes with `utils.process`.

Reference: #5099

### DIFF
--- a/avocado/utils/process.py
+++ b/avocado/utils/process.py
@@ -35,9 +35,6 @@ from . import astring, path
 from .wait import wait_for
 
 LOG = logging.getLogger('avocado.test')
-stdout_log = logging.getLogger('avocado.test.stdout')
-stderr_log = logging.getLogger('avocado.test.stderr')
-output_log = logging.getLogger('avocado.test.output')
 
 #: The active wrapper utility script.
 CURRENT_WRAPPER = None
@@ -533,7 +530,7 @@ class SubProcess:
 
     def __init__(self, cmd, verbose=True, allow_output_check=None,
                  shell=False, env=None, sudo=False,
-                 ignore_bg_processes=False, encoding=None):
+                 ignore_bg_processes=False, encoding=None, log=None):
         """
         Creates the subprocess object, stdout/err, reader threads and locks.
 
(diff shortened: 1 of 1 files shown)


willianrampazzo (approved):
Nice!

avocado/utils/process.py
```diff
@@ -35,9 +35,6 @@
 from .wait import wait_for
 
 LOG = logging.getLogger('avocado.test')
```
beraldoleal, line 37: IMO, the default namespace should be `__name__`.

avocado/utils/process.py
```diff
@@ -581,6 +578,10 @@ def __init__(self, cmd, verbose=True, allow_output_check=None,
                          of the command result stdout and stderr, by default
                          :data:`avocado.utils.astring.ENCODING`
         :type encoding: str
+        :param log: User's custom logger, which will be logging the subprocess
+                    outputs. When this variable is not used, the
+                    `avocado.utils.process` logger will be used.
```
beraldoleal, line 583: So it will be coherent with this comment.

avocado/utils/process.py
```diff
@@ -533,7 +530,7 @@ class SubProcess:
 
     def __init__(self, cmd, verbose=True, allow_output_check=None,
                  shell=False, env=None, sudo=False,
-                 ignore_bg_processes=False, encoding=None):
+                 ignore_bg_processes=False, encoding=None, log=None):
```
beraldoleal, line 533: nitpick: maybe rename log to "logger".

beraldoleal (changes requested):
Hi @richtja, thanks for this. Almost good to me, I left one comment for you.

avocado/utils/process.py
richtja: Yes, I wanted to do that, but I got a lot of problems with functional output tests, which I haven't fixed yet, so I decided to do this change in the separated PR.

avocado/utils/process.py
richtja: I can do this change.

avocado/utils/process.py
beraldoleal: I find strange to have this separated specially because of the docstrings above. We are changing the API, so IMO we could do this with the less impact as possible to the users. Adding an argument now and then changing the default value latter, sounds strange to me.

richtja:
@beraldoleal, @willianrampazzo thanks for your comments, I created new version #5156 with the fixes.

PR #4960, opened 2021-09-21
Update makefile target, remove Makefile.include

ana:
* remove requirements-plugins target
It's not used, none of the plugins under optional_plugins provides a `requirements.txt` file or a `Makefile` with a requirements target.

* remove rpm build targets
Remove all the related rpm build targets. They are not longer used since all the continuous packaging build is done by the packit service.
Move the two remaining targets: `pip` and `install` to `Makefile` and remove `Makefile.include`

* remove requirements-plugins target
It's not used, none of the plugins under optional_plugins provides a `requirements.txt` file or a `Makefile` with a requirements target.
    
Reference: https://github.com/avocado-framework/avocado/issues/4372

PR #5135, opened 2021-11-16
Reorg makefiles, add makefile with packaging rules (v2)

ana:
This is a v2 of https://github.com/avocado-framework/avocado/pull/4960  
Changes: we have decided to keep the packaging targets in a different Makefile

---

Reorganize makefiles to create `Makefile.pkg` with all the variables and targets related to rpm packaging.

Remove completely variables RELEASE_COMMIT and RELEASE_SHORT_COMMIT, since they are not used.

Remove requirements-plugins target, it's not used. None of the plugins under `optional_plugins` provides a `requirements.txt` file or a `Makefile` with a requirements target.


Makefile
```diff
@@ -31,27 +22,17 @@ all:
...
-	@echo
-	@echo "Release related targets:"
-	@echo "source-release:  Create source package for the latest tagged release"
-	@echo "srpm-release:    Generate a source RPM package (.srpm) for the latest tagged release"
-	@echo "rpm-release:        Generate binary RPMs for the latest tagged release"
-	@echo
 
-include Makefile.include
+pip:
+	$(PYTHON) -m pip --version || $(PYTHON) -m ensurepip $(PYTHON_DEVELOP_ARGS) || $(PYTHON) -c "import os; import sys; import urllib; f = urllib.urlretrieve('https://bootstrap.pypa.io/get-pip.py')[0]; os.system('%s %s' % (sys.executable, f))"
```
beraldoleal, line 35: I understand that you are basically moving things from a different place here, but since this target has no help string, It would be much appreciated if we had a proper help message for this target.

Makefile.pkg
```diff
@@ -0,0 +1,66 @@
...
+	@echo "source-release:  Create source package for the latest tagged release"
+	@echo "srpm-release:    Generate a source RPM package (.srpm) for the latest tagged release"
+	@echo "rpm-release:        Generate binary RPMs for the latest tagged release"
+	@echo
+	@echo "variables:         Show the value of variables as defined in this Makefile or"
+	@echo "                   given as input to make"
+	@echo
+
+clean:
+	rm -rf SOURCES BUILD
```
beraldoleal, line 30: When using the default Makefile, calling `make clean` it will also remove `SOURCES` and `BUILD` directories. So, to avoid any kind of contamination I'm wondering if it wouldn't be safer to call `make clean` here. What do you think?

Also, would be nice to have a help message for clean as well, specially if there is a difference between this and the main one.

Makefile.pkg
```diff
@@ -0,0 +1,66 @@
...
+
+source: clean
+	if test ! -d SOURCES; then mkdir SOURCES; fi
```
beraldoleal, line 33: Just out of curiosity, in what cases this if is needed ? 

This target it will only be skipped if the previous return != 0 (in case of rm permission denied). However when the directory does not exist IMO it is safe to ignore and assume that the directory it was removed (that is the reason rm returns 0). 

So, maybe this if is not needed and you can create the directory anyway. Also mkdir has the `-p` that will ignore if it is already there.

beraldoleal (commented):
HI @ana thanks for this. LGTM, just two small comments for you, before accepting.

Makefile
ana: I'm not sure that would be a good idea. This is an auxiliary target and we don't want people using it directly. It's just a help for people who don't have pip in their systems.

Makefile.pkg
ana: First, This makefile shouldn't depend on the main one at all. It's a self-contained one for packaging  stuff.
Second, I was in fact planning to modify the clean command in setuptools to not longer remove SOURCES and BUILD, since it doesn't make sense anymore to remove it there. 
Ok for the help message.

Makefile
beraldoleal: IMO all targets could have a help message. Specially this case, so a disclaimer saying "Use this target only if you don't have pip installed".

Makefile.pkg
beraldoleal: Yes, makes sense. Thanks.

ana:
@beraldoleal I have forced-pushed an update adding the help message for `pip (Makefile)`  and  `clean (Makefile.pkg)`

beraldoleal (approved):
LGTM, thanks.

willianrampazzo (approved):

Issue #5083, opened 2021-11-01
Avocado/Spawners should be smart enough when setting/saving to the output dir


willianrampazzo:
**Is your feature request related to a problem? Please describe.**
On the legacy runner, the content saved to the output dir was available in the `job-results` directory. With the nrunner architecture, we decided to leave the content in a temporary directory, which may be located in a different environment than where Avocado was started. This decision was made because the transfer of huge files could be a bottleneck to the execution of the tests. The current implementation allows the users to _get_ the content using the `avocado jobs get-output-files`.

See https://github.com/avocado-framework/avocado/pull/5059#discussion_r737291135.

**Describe the solution you'd like**
Avocado and the spawners should be smart enough to _know_ they have access to the `job-results` folder and save the test content to it.

For the `process` spawner, as it is running in the same environment where Avocado was started, there is no need to create another temporary folder to store the test files. The output dir can be inside the `job-results` dir.

For the `podman` spawner, mounting the `job-results/<test>` dir is a possibility to make it accessible to the tests.

**Describe alternatives you've considered**
The alternative is implemented today and uses a manual step to _get_ the test files saved to the output dir.

**Additional information**

richtja:
Resolved in  #5179

PR #5154, opened 2021-11-23
Spawners output_dir

richtja:
Right now, the test output directory is managed by runners, and it is separated from the `job/test-results` folder. Most of the runners are using the temporary directory for test output, and this information might be lost. Only possible way how to store this information to the avocado results is use the avocado messages or use the `avocado jobs get-output-files` script. For some spawners like `ProcessSpawner` this is an unnecessarily complicated process, because we have to save the same data twice even the tests are run in the same environment as avocado.

This PR will make spawners smart enough to set the output directory for runners, so for the avocado will be easier to manage those files. For example, the `ProcessSpawner` will set output dir to the `job/test-reulst` so there is no need for creating the temporary directory and copping to the `job/test-results`

Reference: #5083

### DIFF
--- a/avocado/core/nrunner.py
+++ b/avocado/core/nrunner.py
@@ -122,6 +122,7 @@ class Runnable:
         self.tags = kwargs.pop('tags', None)
         self.requirements = kwargs.pop('requirements', None)
         self.variant = kwargs.pop('variant', None)
+        self.output_dir = kwargs.pop('output_dir', None)
         self.kwargs = kwargs
 
     def __repr__(self):
@@ -191,6 +192,9 @@ class Runnable:
         if self.variant is not None:
             args.append('variant=json:%s' % json.dumps(self.variant))
 
+        if self.output_dir is not None:
+            args.append('output_dir=json:%s' % json.dumps(self.output_dir))
+
         for key, val in self.kwargs.items():
             if not isinstance(val, str) or isinstance(val, int):
                 val = "json:%s" % json.dumps(val)
@@ -218,6 +222,8 @@ class Runnable:
             kwargs['tags'] = self.get_serializable_tags()
         if self.variant is not None:
             kwargs['variant'] = self.variant
+        if self.output_dir is not None:
+            kwargs['output_dir'] = self.output_dir
         if kwargs:
             recipe['kwargs'] = kwargs
         return recipe
@@ -928,10 +934,9 @@ class Task:
             runners_registry = RUNNERS_REGISTRY_STANDALONE_EXECUTABLE
         return self.runnable.pick_runner_command(runners_registry)
 
-    def setup_output_dir(self):
-        self.output_dir = tempfile.mkdtemp(prefix='.avocado-task-')
-        env_var = {'AVOCADO_TEST_OUTPUTDIR': self.output_dir}
-        self.runnable.kwargs.update(env_var)
+    def setup_output_dir(self, output_dir):
+        self.output_dir = output_dir or tempfile.mkdtemp(prefix='.avocado-task-')
+        self.runnable.output_dir = self.output_dir
 
     @classmethod
     def from_recipe(cls, task_path, known_runners):
@@ -977,7 +982,6 @@ class Task:
         return args
 
     def run(self):
-        self.setup_output_dir()
         runner_klass = self.runnable.pick_runner_class(self.known_runners)
         runner = runner_klass(self.runnable)
         for status in runner.run():
--- a/avocado/core/nrunner.py
+++ b/avocado/core/nrunner.py
@@ -550,11 +550,9 @@ class ExecTestRunner(BaseRunner):
         return avocado_test_env_variables
 
     def run(self):
-        env = None
+        env = dict(os.environ)
         if self.runnable.kwargs:
-            current = dict(os.environ)
-            current.update(self.runnable.kwargs)
-            env = current
+            env.update(self.runnable.kwargs)
 
         # set default Avocado environment variables if running on a valid Task
         if self.runnable.uri is not None:
--- a/avocado/core/task/statemachine.py
+++ b/avocado/core/task/statemachine.py
@@ -247,6 +247,7 @@ class Worker:
             await asyncio.sleep(0.1)
             return
 
+        await self._spawner.create_task_output_dir(runtime_task)
         LOG.debug('Task "%s": about to be spawned with "%s"',
                   runtime_task.task.identifier, self._spawner)
         start_ok = await self._spawner.spawn_task(runtime_task)
--- a/avocado/core/task/statemachine.py
+++ b/avocado/core/task/statemachine.py
@@ -283,6 +283,10 @@ class Worker:
             except asyncio.TimeoutError:
                 runtime_task.status = 'FINISHED W/ TIMEOUT'
         result_stats = self._state_machine._status_repo.result_stats
+
+        if runtime_task.task.runnable.config.get('test.output.enabled', True):
+            await self._spawner.save_task_output_dir(runtime_task)
+
         if self._failfast and 'fail' in result_stats:
             await self._state_machine.abort("FAILFAST is enabled")
             raise TestFailFast("Interrupting job (failfast).")
--- a/avocado/core/test.py
+++ b/avocado/core/test.py
@@ -276,12 +276,15 @@ class Test(unittest.TestCase, TestData):
             self.__base_logdir_tmp = tempfile.TemporaryDirectory(prefix=prefix)
             self.__base_logdir = self.__base_logdir_tmp.name
 
-        self.__base_logdir = os.path.join(self.__base_logdir, 'test-results')
-        logdir = os.path.join(self.__base_logdir, self.name.str_filesystem)
-        if os.path.exists(logdir):
-            raise exceptions.TestSetupFail("Log dir already exists, this "
-                                           "should never happen: %s"
-                                           % logdir)
+        if self._config.get("run.test_runner", 'nrunner') == 'nrunner':
+            logdir = self.__base_logdir
+        else:
+            self.__base_logdir = os.path.join(self.__base_logdir, 'test-results')
+            logdir = os.path.join(self.__base_logdir, self.name.str_filesystem)
+            if os.path.exists(logdir):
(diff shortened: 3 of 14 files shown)


avocado/core/nrunner.py
```diff
@@ -928,10 +934,9 @@ def are_requirements_available(self, runners_registry=None):
             runners_registry = RUNNERS_REGISTRY_STANDALONE_EXECUTABLE
         return self.runnable.pick_runner_command(runners_registry)
 
-    def setup_output_dir(self):
-        self.output_dir = tempfile.mkdtemp(prefix='.avocado-task-')
-        env_var = {'AVOCADO_TEST_OUTPUTDIR': self.output_dir}
```
willianrampazzo, line 933: This breaks https://avocado-framework.readthedocs.io/en/latest/guides/writer/chapters/writing.html#environment-variables-for-tests for exec-tests.

avocado/core/task/statemachine.py
```diff
@@ -247,6 +247,7 @@ async def start(self):
             await asyncio.sleep(0.1)
             return
 
+        await self._spawner.create_task_output_dir(runtime_task)
```
willianrampazzo, line 250: Personally, I think the call to create the output-dir should not be the responsibility of the state machine.

beraldoleal, line 250: I'm not sure if this is the right place for this call. IIUC, the statemachine would be responsible only for the 'state machine', having this logic inside this file would open a precedence to more code living here. I'm not sure, but would be nice to double check.

@clebergnu ?

avocado/plugins/spawners/podman.py
```diff
@@ -176,6 +184,16 @@ async def spawn_task(self, runtime_task):
...
 
+    async def create_task_output_dir(self, runtime_task):
+        task = runtime_task.task
+        output_dir_path = os.path.join(self.job_output_dir,
+                                       task.identifier.str_filesystem)
+        output_podman_path = os.path.join('/tmp', 'avocado-results')
+
+        os.makedirs(output_dir_path, exist_ok=True)
+        runtime_task.task.setup_output_dir(output_podman_path)
+        await self._create_container_for_task(runtime_task, output_dir_path)
```
beraldoleal, line 195: While is ok to call other methods inside a bigger method like "spawn_task", I find odd having this inside this method.

Based on this method name, I was not expecting to create dir *plus* creating a container as well. Maybe this call should be outside of this method. This could return the output_dir_path, so that the upper caller could be responsible to passing this to the _create_cotainer_for_task. What do you think?

avocado/plugins/spawners/podman.py
```diff
@@ -176,6 +184,16 @@ async def spawn_task(self, runtime_task):
...
+
     async def wait_task(self, runtime_task):
         while True:
             if not self.is_task_alive(runtime_task):
```
beraldoleal, line 202: BTW, I'm doing some changes inside this file as part of the podman + eggs effort. So I guess since I'm late with that, I will have to handle the conflict at this point. :(

avocado/core/test.py
```diff
@@ -276,12 +276,15 @@ def record_and_warn(*args, **kwargs):
             self.__base_logdir_tmp = tempfile.TemporaryDirectory(prefix=prefix)
             self.__base_logdir = self.__base_logdir_tmp.name
 
-        self.__base_logdir = os.path.join(self.__base_logdir, 'test-results')
-        logdir = os.path.join(self.__base_logdir, self.name.str_filesystem)
-        if os.path.exists(logdir):
-            raise exceptions.TestSetupFail("Log dir already exists, this "
-                                           "should never happen: %s"
-                                           % logdir)
+        if self._config.get("run.test_runner", 'nrunner') == 'nrunner':
```
beraldoleal, line 279: You are setting a default hardcoded here. Have you hit any issues trying to get the "registration time" default value?

avocado/core/test.py
```diff
@@ -276,12 +276,15 @@ def record_and_warn(*args, **kwargs):
...
+            logdir = self.__base_logdir
+        else:
+            self.__base_logdir = os.path.join(self.__base_logdir, 'test-results')
+            logdir = os.path.join(self.__base_logdir, self.name.str_filesystem)
+            if os.path.exists(logdir):
+                raise exceptions.TestSetupFail("Log dir already exists, this "
+                                               "should never happen: %s"
+                                               % logdir)
```
beraldoleal, line 287: Also, since legacy runner is deprecated and not the default one, I think maybe we don't need this logic anymore.

avocado/core/test.py
```diff
@@ -740,7 +743,8 @@ def _run_avocado(self):
             process.OUTPUT_CHECK_RECORD_MODE = 'combined'
 
         testMethod = getattr(self, self._testMethodName)
-        self._start_logging()
+        if self._config.get("run.test_runner", 'nrunner') != 'nrunner':
+            self._start_logging()
```
beraldoleal, line 747: Same here.

selftests/unit/test_test.py
```diff
@@ -51,8 +51,10 @@ def tearDown(self):
     def test_ugly_name(self):
         def run(name, path_name):
             """ Initialize test and check the dirs were created """
+            config = {"run.test_runner": 'runner'}
```
beraldoleal, line 54: Same here, maybe we need to adapt those tests to use the default runner (nrunner).

avocado/core/nrunner.py
```diff
@@ -550,11 +550,9 @@ def _get_env_variables(self):
         return avocado_test_env_variables
 
     def run(self):
-        env = None
+        env = dict(os.environ)
```
beraldoleal, line 553: Could you please describe better the problem you are solving in the commit message?

avocado/plugins/spawners/common.py
```diff
@@ -0,0 +1,30 @@
...
+        if parser is None:
+            return
+
+        parser = parser.add_argument_group('spawner specific options')
+        help_msg = ("This will disable the fetching output files from test "
+                    "environment. This might be useful when you are running the "
+                    "tests in different environment than avocado, like external "
+                    "machine, and fetching might be time-consuming.")
+        settings.register_option(section='test.output',
+                                 key='enabled',
```
beraldoleal, line 22: IMO `test.output.enabled = True|False` is very generic, it is not clear that is about the "fetching" aspect. What about calling `test.output.fetch = True|False` matching the long_arg option?

beraldoleal (changes requested):
Hi @richtja nice work here. I left a few comments/suggestions for you.

avocado/core/nrunner.py
```diff
@@ -122,6 +122,7 @@ def __init__(self, kind, uri, *args, config=None, **kwargs):
         self.tags = kwargs.pop('tags', None)
         self.requirements = kwargs.pop('requirements', None)
         self.variant = kwargs.pop('variant', None)
+        self.output_dir = kwargs.pop('output_dir', None)
```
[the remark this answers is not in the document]
richtja: I understand your concern abut this, but I don't think that usage of config is a good solution here. The configuration is set by user, and it is something which shouldn't by changed by avocado itself during runtime. We have the same problem with `tags`, `requirments` and `variants` so maybe we should come up with different solution for sending those data to runners.

avocado/core/nrunner.py
richtja: Thanks, I will investigate it.

avocado/core/nrunner.py
```diff
@@ -928,10 +934,9 @@ def are_requirements_available(self, runners_registry=None):
             runners_registry = RUNNERS_REGISTRY_STANDALONE_EXECUTABLE
         return self.runnable.pick_runner_command(runners_registry)
 
-    def setup_output_dir(self):
-        self.output_dir = tempfile.mkdtemp(prefix='.avocado-task-')
-        env_var = {'AVOCADO_TEST_OUTPUTDIR': self.output_dir}
-        self.runnable.kwargs.update(env_var)
+    def setup_output_dir(self, output_dir):
+        self.output_dir = output_dir or tempfile.mkdtemp(prefix='.avocado-task-')
+        self.runnable.output_dir = self.output_dir
```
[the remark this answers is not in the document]
richtja: I put it to the Runnable because it is needed on runner side and I leave it in Task because it was part of changes for the  https://avocado-framework.readthedocs.io/en/latest/guides/writer/chapters/writing.html#environment-variables-for-tests. But as @willianrampazzo pointed out, this PR breaks this feature, so I will have to make some changes here.

avocado/plugins/spawners/podman.py
richtja: sure I will change that.

avocado/core/test.py
richtja: Because the `test.py` has a lot of legacy code, I decided to leave it here for now. And we can remove it when we will be removing every legacy code from `test.py`

selftests/unit/test_test.py
richtja: IIUTC those tests are for the legacy features and when we will completely remove the legacy code then those tests will be removed too.

richtja:
@willianrampazzo @beraldoleal thanks for your comments I will work on new version.

avocado/core/nrunner.py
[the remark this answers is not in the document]
beraldoleal: My problem with this approach is that we teach developers and users saying that `args` and `kwargs` are handled different by runners and usually. In the case of exectest, `kwargs` are environment variables passed to the tests. However here, we have to keep a "reserved words" list where those "environment variables"  can't be used because we are using internally, and we are just increasing this list.
  
About the configuration aspect: users can configure things based on `avocado.conf` or using the config dict (when using the Job API). The entire idea behind the config dict was that we could concentrate settings in a single place, to avoid confusion and here we are still using a different method to configure things. IMO this could be solved using the same dict, by using different namespaces: `runnable.*` or `runners.*` for instance. And we could even state that some namespaces would be read only, because they are used internally.

Also, due our architecture, third-party plugins wouldn't need to change core method's signature to support new features, they could rely on some config's namespace.

Btw, this is not a blocker, but I think we should think about it to avoid future issues.

richtja:
Closing in favor of v2 #5179

PR #5155, opened 2021-11-23
log: use namespace instead of module

beraldoleal:
For end-users, module name + line number is not an useful information.
Also, recent changes are using custom namespaces to better organize the
loggers. This patch it was postponed for post-lts and now is time to use
this new format.

Related to #5099.

### DIFF
--- a/avocado/core/job.py
+++ b/avocado/core/job.py
@@ -198,8 +198,7 @@ class Job:
 
     def __start_job_logging(self):
         # Enable test logger
-        fmt = ('%(asctime)s %(module)-16.16s L%(lineno)-.4d %('
-               'levelname)-5.5s| %(message)s')
+        fmt = ('%(asctime)s %(name)s %(levelname)-5.5s| %(message)s')
         test_handler = output.add_log_handler(LOG_JOB,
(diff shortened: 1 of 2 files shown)


richtja (approved):
LGTM, thanks

pevogam:
Hi @beraldoleal is there a way for us to use our own logging format here? For all sorts of debug we always used to find the log line information very useful and after the recent update to Avocado 98.0 we noticed it is gone. The line has saved us quite many hours of debugging for all the years so I wonder if you provide an alternative to this commit that allows custom formatting for those of us that don't quite agree about it being useless?

beraldoleal:
Hi @pevogam, sorry for that. I can see how that line was useful for developers during debugging.

We decided to remove it because it was too much details for users to see during their log messages.

The original plan it was to introduce the ability to customize this via `logging.config`. I will discuss this with the team, so we can introduce this as soon as possible.

pevogam:
Thanks for looking into it @beraldoleal. If there is any way in which I can assist you let me know. For now on our side we will simply add a patch that restores the logged line number.

PR #5164, opened 2021-11-29
check.py: missing initialization for optional-plugins

ana:
resolver.references must be initialized for optional-plugins otherwise they can't be run.

This fixes a problem when running: `python selftests/check.py --select optional-plugins`


clebergnu (approved):
LGTM, thanks!

richtja (approved):

PR #5167, opened 2021-11-30
.travis.yml: save some build credits

ana:
Do not run static checks in travis, they're run in github actions
and this is enough since the statics checks are independant of the arch.

Not using 'make check' has also the advantage of not running
the targets clean and uninstall. This should save some build time
as well.

Do not set AVOCADO_PARALLEL_LINT_JOBS, it was used by the static
jobs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,6 @@ arch:
 python:
     - "3.9"
 
-env:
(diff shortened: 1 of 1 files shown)


richtja (approved):

clebergnu (approved):
LGTM!

PS: I'm still puzzled by the `skipIf` not being effective, but this is a different issue.

PR #5173, opened 2021-12-03
optional_plugins/html: changes to implement PEP420

ana:
Rename __init__.py to avocado_result_html.py and update entry_points accordingly.
Use find_namespace_packages instead find_packages to declare explicitly namespace packages provided by this plugin/package.

Reading first https://packaging.python.org/guides/packaging-namespace-packages/#native-namespace-packages should help to review this.


beraldoleal (approved):
LGTM, thanks.

clebergnu (approved):
LGTM, thanks!

PS: I think we should think about a different naming standard for simple plugins with a single file, like `plugin.py`, but that is a different discussion.

clebergnu (changes requested):
Actually, there a number of CI failures.  @ana have you looked into them?

ana:
> Actually, there a number of CI failures. @ana have you looked into them?

Yes, I am doing tests on https://github.com/avocado-framework/avocado/pull/5174 and then I jumped into other updates with CI that should help here, like https://github.com/avocado-framework/avocado/pull/5183

ana:
> 
> PS: I think we should think about a different naming standard for simple plugins with a single file, like `plugin.py`, but that is a different discussion.

That's still doable. What you have in mind? `html.py` ?(I considered using this one) Exactly the plugin name `avocado-framework-plugin-result-html.py`? shorter name `plugin-result-html.py`?

ana:
After merging https://github.com/avocado-framework/avocado/pull/5186 and a small change, this should work. I'm closing this PR and sending a new one including other plugins.

PR #5176, opened 2021-12-07
core.job: remove legacy root logger

beraldoleal:
Currently we are using 'avocado' namespace and not touching on the
'root' logger. So this is not needed anymore.


richtja (approved):
LGTM, thanks

Issue #5166, opened 2021-11-30
Avocado still modifying the root logger


pevogam:
Earlier today @samiraguiar pointed out to me that Avocado is still modifying the root logger here:

https://github.com/avocado-framework/avocado/blob/2f52fd3531b6677c4d564e7c897f5e2558918482/avocado/core/output.py#L392

Is this supposed to really happen? I thought the point of staying away from the root logger was to leave it to the user to decide about the logging level and other properties when it comes to this logger in particular.

It could be that I am missing something important here and this is more appropriate in discussions but we can convert it if you think this is more appropriate.

willianrampazzo:
Hi, @pevogam, this line was added six years ago. I think it is a remaining from the changes that @beraldoleal and @richtja are doing to avoid using the root logger; thanks for pointing it out!

pevogam:
Hi @willianrampazzo, I see this sounds pretty interesting. The reasoning behind the question is that this modification resets some of our application-level logging down to DEBUG level and we have a workaround, however I am not sure if this is a desired side effect considering Avocado's transition to logging namespaces.

beraldoleal:
@pevogam thanks for reporting. This is not expected and probably a left over.

Thanks @richtja for self assigned on this and please let me know if you need any help.

pevogam:
Thanks all and thanks to @samiraguiar for making me aware of this.

Issue #5178, opened 2021-12-07
Improvements on our log system (big picture)


beraldoleal:
_Tl;dr: This is an attempt to bond all logging issues related to this epic issue. Here we describe our problem and some room for improvements regarding our logging system._

During the last Avocado releases, its architecture has evolved considerably, however, even with some particular improvements, our logging mechanism is still in need of some care.

So far, we have adopted two relevant conventions: 1. We are avoiding the use of the root logger, saving everything inside the ‘avocado.*’ namespace; and 2. We are reserving the JOB_LOG ('avocado.test') only for logs related to the execution of the tests, in this way, each module starts logging in its namespace, using `__name__` as logger name.

However, besides some minor changes on our current effort, we still have some room for improvement:

 1. _Logs outside the scope of a job_: With the new architecture, there is a need to log messages outside the scope of a job (messages that do not fit in job.log). For that, we need to define our handlers. Related issues:

    -  #5177
    - #5120
    - #4856

 2. _Configuration_: Currently we do the configuration in code with several routines scattered in some places. In the current scenario, maybe this is no longer necessary and we should try to eliminate as much as possible this "logic in code" and use more the `logging.config` module from python itself. Here we still need to define some command line "helpers" that will be useful during the execution of jobs without the need to change those files. Related issues:
 
     - #4856
     - #4121

 3. _Documentation_: We lack proper documentation here. It should be crystal clear to our users and contributors, what our handlers are and what namespaces they are acting on. Related issues:

beraldoleal:
@pevogam @clebergnu @richtja toughs?

pevogam:
@beraldoleal This sounds great and thanks for summarizing all logging-related issues and adaptations we need.

My only concern so far that we might break a lot of the already fixed logging issues so it would be great if we also have a way to test for missing logging messages and such. This wasn't trivial to detect so far and it took us multiple weeks of discoveries to identify these issues and now that most of them are fixes it would be great if we test for them (hopefully automatically) with more changes to come.

PR #5180, opened 2021-12-07
core.output: remove legacy root logger

richtja:
Currently we are using 'avocado' namespace and not touching on the
'root' logger. Unfortunately, we still had some occurrences of the root
logger in `core.output`, this is fixing that.

Reference: #5166

### DIFF
--- a/avocado/core/output.py
+++ b/avocado/core/output.py
@@ -383,13 +383,13 @@ def early_start():
         add_log_handler(LOG_UI.getChild("debug"), logging.StreamHandler,
                         sys.stdout, logging.DEBUG)
     if os.environ.get('AVOCADO_LOG_EARLY'):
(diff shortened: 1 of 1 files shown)


beraldoleal (approved):
Hi @richtja thanks for this. I'm curious to understand why we do need the custom InMemoryHandler, but I guess for now this makes sense.

richtja:
Hi @beraldoleal we need them for the early stage, because the Avocado is wrapping the `sys.stdout` and `sys.stderr` to the `avocado.core.output.StdOutput` class for easier configuration of those streams. Because of this functionality, we need `InMemoryHandler` to record logs in the early stage when the logging handlers are not set properly by `avocado.core.output.reconfigure()`.  I didn't want to modify this process in here, because we will be doing the full refactoring of avocado logging system in  #5178

clebergnu (approved):
LGTM, thanks!

PR #5183, opened 2021-12-08
.cirrus.yml: Update Fedora versions (remove 33, add 35)

ana:
Remove Fedora 33 that reached EOL on 2021-11-30 and add Fedora 35.

Update fedora_selftests_task to use Fedora 35 prebuilt container image, add the dockerfile for this image.

Minor update on `selftests/functional/test_lv_utils.py`  add check for binary modprobe. Skip the test if modprobe is not present in the system and show an useful error message.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -5,7 +5,7 @@ redhat_version_task:
        - python3 -m avocado --version
     container:
         matrix:
-          - image: fedora:33
+          - image: fedora:35
           - image: fedora:34
           - image: registry.access.redhat.com/ubi8/ubi
 
@@ -22,7 +22,7 @@ fedora_develop_install_uninstall_task:
        - test `python3 -m avocado plugins | grep ^html | wc -l` -eq "0"
        - test `python3 -m avocado plugins | grep ^robot | wc -l` -eq "0"
     container:
-      image: fedora:34
+      image: fedora:35
 
 redhat_egg_task:
     egg_script:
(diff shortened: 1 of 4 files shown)


selftests/functional/test_lv_utils.py
```diff
@@ -114,6 +114,8 @@ class DiskSpace(unittest.TestCase):
                      'macOS does not support scsi_debug module')
     @unittest.skipIf(not process.can_sudo(), "This test requires root or "
                      "passwordless sudo configured.")
+    @unittest.skipIf(process.system("which modprobe", ignore_status=True),
+                     "kmod not installed (command modprobe is missing)")
```
beraldoleal, line 118: This is fine. However we have now, 4 definitions of `def missing_binary()` and now this new method. So maybe would be the time to merge them into `selftests/utils.py`. Just a suggestion, not a blocker.

.cirrus.yml
```diff
@@ -74,4 +74,4 @@ fedora_selftests_task:
        - PATH=$HOME/.local/bin:$PATH LANG=en_US.UTF-8 AVOCADO_CHECK_LEVEL=0 python3 selftests/check.py  --skip static-checks
     container:
         matrix:
-          - image: quay.io/avocado-framework/avocado-ci-fedora-33
+          - image: quay.io/avocado-framework/avocado-ci-fedora-35
```
beraldoleal, line 77: Are we still in need of `./contrib/containers/ci/selftests/fedora-33.docker` file? If not, perhaps we could drop those files together with this commit.

Also I wouldn't mind if you squash this commit and the previous one, since they are related.

selftests/functional/test_lv_utils.py
ana: I'll change it. and it's not new, it's also at https://github.com/avocado-framework/avocado/blob/master/selftests/functional/test_lv_utils.py#L25
We'll need a consolidation of all this once we move the tests to avocado.utils

ana:
@beraldoleal updated

clebergnu (approved):
LGTM, thanks!

beraldoleal (approved):
LGTM, thanks.

PR #5184, opened 2021-12-08
utils: avoid using root logger

### LINKED ISSUE #4997: logging: use specific namespaces for each utility library

beraldoleal:
Those are remaining calls to root logger and some small fixes with the
namespaces. Fixes #4997.

### DIFF
--- a/avocado/utils/debug.py
+++ b/avocado/utils/debug.py
@@ -20,7 +20,7 @@ import os
 import time
 
 # Use this for debug logging
-LOGGER = logging.getLogger("avocado.app.debug")
+LOGGER = logging.getLogger(__name__)
 
 # measure_duration global storage
 __MEASURE_DURATION = {}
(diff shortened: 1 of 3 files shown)


richtja (approved):

clebergnu (approved):
LGTM, thanks!

PR #5188, opened 2021-12-10
pre-release checks failed tests summary

richtja:
Let's add the output of the failed tests to the pre-release checks for
easier understanding, which tests have failed.

### DIFF
--- a/selftests/pre_release/jobs/pre_release.py
+++ b/selftests/pre_release/jobs/pre_release.py
@@ -28,6 +28,7 @@ vmimage = {
 
 if __name__ == '__main__':
     os.chdir(ROOT_DIR)
-    with Job.from_config({}, [cirrus_ci, parallel_1, vmimage]) as j:
+    config = {'job.output.testlogs.statuses': ['FAIL', 'ERROR', 'INTERRUPT']}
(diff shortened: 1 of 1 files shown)


ana (approved):

selftests/pre_release/jobs/pre_release.py
```diff
@@ -28,6 +28,8 @@
 
 if __name__ == '__main__':
     os.chdir(ROOT_DIR)
-    with Job.from_config({}, [cirrus_ci, parallel_1, vmimage]) as j:
+    config = {'job.output.testlogs.statuses': ['FAIL', 'ERROR', 'INTERRUPT'],
+              'job.output.testlogs.logfiles': ['debug.log']}
```
beraldoleal, line 32: Currently `job.output.testlogs.logfiles`  has as default `['debug.log']`. So, I'm just curious if we really need this second option.

beraldoleal (changes requested):

selftests/pre_release/jobs/pre_release.py
richtja: yes, you are right, I just copied it from `selftest/check.py` so I think we can remove it there too. Thanks

beraldoleal (approved):
LGTM, thanks @richtja.

Issue #5208, opened 2021-12-20
GitHub actions: static-checks and linux checks give false positives


ana:
Some GitHub Actions are giving false positives.  You can see this at this PR https://github.com/avocado-framework/avocado/pull/5207
I have made changes to make the static-checks and linux checks to fail but they're shown as passing. If you expand the log at the tests of this PR you can see they failed.

ana:
@beraldoleal  @clebergnu  @richtja  I'm afraid this is a blocker to merge PR post-release :(

PR #5210, opened 2021-12-21
setup.py test exit code fix

richtja:
The `setup.py test` was always returning exit code 0, because the exit
code from `check.py` wasn't respected. This will fix that, by exiting
`setup.py test` with code from `check.py`.

Reference: #5208


ana (approved):
Great, thanks!

PR #5212, opened 2021-12-21
Update plugins to support PEP420

ana:
This updates all the plugins to support PEP 420. 
I sent a PR to update the html plugin a couple of weeks ago (https://github.com/avocado-framework/avocado/pull/5173). This changeset contains that initial PR and a workaround for jinja2, used by the html plugin.


optional_plugins/robot/tests/test_robot.py
```diff
@@ -1,7 +1,7 @@
 import os
 import unittest
 
-import avocado_robot
+import avocado_robot.robot
```
beraldoleal, line 4: Maybe I'm just thinking about future changes, so I'm not sure if this should be part of this patch series or not (probably not). In any case, I'm wondering how the plugins' namespace will be in the future? Have you considered this?

Are all plugins going to have `avocado_*` on the package name? Or they could be part of `avocado.plugins.*`?  It feels to me that the second option it would be more explicit that we are talking about a plugin.

beraldoleal (commented):
Hi @ana, thanks for this. LGTM, just one small question for you.

optional_plugins/robot/tests/test_robot.py
ana: @beraldoleal Right now, I'm keeping the namespace we are already using:  'avocado_$PLUGIN_NAME`. The plugins are included in the code of avocado but we're still installing them as a separate thing.

A change of namespace such as moving to `avocado.plugins.*` would need a discussion and it'll be hard not to tangle it  with the open discussion about what to do with `optional_plugins.`

beraldoleal (approved):

richtja (approved):
LGTM, thanks

PR #5222, opened 2022-01-04
pylint: do not override R1725, update GHA/setup.yml and some scripts cleanup

ana:
This PR has 3 separate changes of things that has been bothering me:

* pylint: do not override R1725 (super-with-arguments)
  Remove this old override and instead remove these arguments because they're the default arguments anyway.

* selftests: remove old excludes
Quick scrips cleanup: remove files that don't need to be excluded anymore from tests

* `setup.yml`: install requirements and skip static-checks
  The job 'developer virtualenv installation' was missing the psutils to run the checks. This has been failing for some time but it was hidden because issue  https://github.com/avocado-framework/avocado/issues/5208
    
Also, skip running the static checks.


richtja (approved):
LGTM, thanks

.github/workflows/setup.yml
```diff
@@ -221,10 +221,11 @@ jobs:
         run: |
          python3 -m venv env
          source env/bin/activate
+         pip install psutil==5.8.0
```
beraldoleal, line 224: This requirement is already defined at `requirements-dev.txt` and if we need to change this is in the future we will chave to update twice.  This file has also other requirements need for selftests.

What about calling `make develop`? Or if installing `make` it is not an option, maybe calling from pip with: `pip install -r requirements-dev.txt`?

beraldoleal (changes requested):
LAGTM, just one small comment.

.github/workflows/setup.yml
ana: Hi @beraldoleal

> This requirement is already defined at `requirements-dev.txt` and if we need to change this is in the future we will chave to update twice. This file has also other requirements need for selftests.

This was deliberate I wanted to test with the minimum requeriments possible. This test doesn't run the static checks, so the only requeriment is `psutils`. I agree the fixed version  isn't ideal so maybe the alternative is just to leave it without version? psutils doesn't change that often.

beraldoleal: Right, you are skipping `static-checks`, but we are running functional tests and unit tests here as well. So, if `psutils` is the only missing dependency, I can think about the following possibilities:

1. Other dependencies are already installed by the system;
3. Some might not be used anymore (i.e: maybe `elementpath`);
3. In some cases, if some dependencies are not installed, then some tests are being skipped (for instance `netifaces` is defined at `requirements-dev.txt` and used at `selftests/functional/plugin/test_vmimage.py`).

So, we are running `python setup.py test`, but not testing everything.

If we would like to have a minimal requirement (and allow some tests to skip), so maybe we could use better the extra_requires feature of pip? and having a default minimal requirement file with an extra called "full" that will install everything.

(I'm not concerned only about the version here. But having the requirements around multiple files it is odd for me.)

ana: @beraldoleal ok, thanks for expanding your concerns. I'll replace the step with `pip install -r requirements-dev.txt`

beraldoleal (approved):
LGTM, thanks.

Issue #5217, opened 2021-12-24
test status should be already set and available for tearDown()


bfaccini:
**Is your feature request related to a problem? Please describe.**
We need to be able to run very early actions (like dumping some state/context informations) upon any test failure/error, but presently the test status is not available (None) during tearDown().

**Describe the solution you'd like**
The test status should be already set and available for tearDown() stage.

**Describe alternatives you've considered**
Overload fail()/error() methods of Avocado base Test class.

bfaccini:
thank you for your help and having fixing it so quickly!
Just a last question, do you have any idea about which Avocado version will start containing this fix ?

PR #3295, opened 2019-09-09
Enable pylints w0402 and w1113 [v2]

richtja:
Changes from v1 (#3281):
- Fixed problem with arguments parsing in `software_manager.py`

PR #5224, opened 2022-01-07
 test status available for tearDown() 

richtja:
This PR will split the `test._run_avocado` into the two methods. The
`_run_test` and `_tearDown`. Thanks to this change, we can store test
status before execution of the tearDown and make it available to the
users.

Reference: #5217


avocado/core/test.py
```diff
@@ -859,7 +859,8 @@ def run_avocado(self):
             self._tag_end()
             self._report()
             self.log.info("")
-            self._stop_logging()
+            if self._config.get("run.test_runner") != 'nrunner':
+                self._stop_logging()
```
beraldoleal: IMO, we could drop this method. We should not keep support to others runners at this point and we don't need to introduce new code just to support the legacy runner.

richtja: yes, I agree with you, we should drop the `_start_logging` and `_stop_logging`. I just didn't want to do it before avocado-vt [#3295](https://github.com/avocado-framework/avocado-vt/pull/3295), because last time we have compatibility issues after dropping legacy code. So I would like to wait until the compatibility between avocado and avocado-vt will be resolved.

richtja:
Hi @bfaccini can you please review this PR, it is solving your problem with test statuses during tearDown() from  #5217.

ana (approved):

clebergnu (approved):
LGTM, thanks!

beraldoleal (approved):
LGTM

bfaccini:
> Hi @bfaccini can you please review this PR, it is solving your problem with test statuses during tearDown() from #5217.

Sorry for being late, and anyway I would not have been a good reviewer as I am not enough familiar with Avocado internals at this time. 
But I would like to thank you for your help and having fixing it so quickly!
Just a last question, do you have any idea about which Avocado version will start containing this fix ?

richtja:
> > Hi @bfaccini can you please review this PR, it is solving your problem with test statuses during tearDown() from #5217.
> 
> Sorry for being late, and anyway I would not have been a good reviewer as I am not enough familiar with Avocado internals at this time. But I would like to thank you for your help and having fixing it so quickly! Just a last question, do you have any idea about which Avocado version will start containing this fix ?

This fix will be available in avocado 95, if everything goes well it should be released on Feb 7.

PR #5237, opened 2022-01-21
Remove two __init__.py files, changes for PEP420

ana:
Changes for PEP420:

Remove  `avocado/core/safeloader/__init__.py` and update headers to work without the defined  `__all__`
Fix assert on `selftests/unit/test_loader.py` to compare against the test path and not against the name.
    
Remove current `avocado/core/requirements/cache/__init__.py` and update headers to work directly with the sqlite based backend, this is the only available implementation anyway.
    

ana:
This PR has previous requirements, I will resubmit later.

PR #5238, opened 2022-01-21
setup.py: changes to implement PEP420

ana:
Use find_namespace_packages instead find_packages to declare explicitly namespace packages provided by this plugin/package.

Add a fallback for systems still using setuptools < 40.
    
Remove all empty `__init__` files. They're not longer necessary now we're using native namespace packages.

Add a  pylint override for W0406 (import-self). The previous change has triggered this warning.
    
References: https://github.com/avocado-framework/avocado/issues/5165


ana:
Experiment is over.

Issue #5116, opened 2021-11-10
test.id and test.name are not right internally


beraldoleal:
**Describe the bug**
Currently we are exposing to our users the right definition of a Test ID and Test Name. However we are not saving the right value to the right property and even test.name is empty today.

**Steps to reproduce**
```python
In [1]: from avocado.core.job import Job 
   ...:  
   ...: job_config = {'resolver.references': ['/bin/true']} 
   ...:  
   ...: with Job.from_config(job_config=job_config) as job: 
   ...:     job.run() 
   ...:     print(job.result.tests[0]['name']) 
   ...:                                                                                                                                                                                                                
JOB ID     : 3cca77a1fce374d40c9d14c82e662f1cb5fc9279
JOB LOG    : /home/local/avocado/job-results/job-2021-11-10T17.30-3cca77a/job.log
 (1-1/1) /bin/true: STARTED
 (1-1/1) /bin/true: PASS (0.01 s)
RESULTS    : PASS 1 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB HTML   : /home/local/avocado/job-results/job-2021-11-10T17.30-3cca77a/results.html
JOB TIME   : 1.02 s
1-1-/bin/true
```

**Expected behavior**
According to our doc, `name` should be only `/bin/true`. Also `jsonresult.py:45` is saving the name as 'id'.

tests.append({'id': str(test.get('name', UNKNOWN)),

**Current behavior**
We are showing the id instead of the name. 

**System information (please complete the following information):**
 - OS: `lsb_release -a`: fedora 33
 - Avocado version: `avocado -v`: master
 - Avocado installation method: pip, rpm, github ?: develop mode

PR #5242, opened 2022-01-26
fixes test.id vs test.name on jsonresult and xunit plugins

beraldoleal:
According to our documentation, 'name' should be only the 'name' portion
of the test, without the prefix identifier. Besides being consistent
with out documentation, this will help third-party tools and future
plugins to detect flaky tests when the same test is executed in a
different suite. Related to #5116.
    

clebergnu (approved):
Checked the code and outcome, and it LGTM.  Thanks!

ana (approved):

PR #5247, opened 2022-02-01
Functional test for requirements resolver

richtja:
The b42f35c7aed488c191f16cb57b2bfc20de410788 is solving the problems of
requirements tests from  #4620, so we can enable them. The only issue is
those are functional tests which can't be run in parallel because the
multiple executions of avocado with requirement runner is  breaking
avocado-software-manager. This commit solves this problem by running
them in serial. It splits the functional selftests into serial and
parallel and runs them in separated suite.

Reference: #4620

### DIFF
--- a/selftests/check.py
+++ b/selftests/check.py
@@ -578,10 +578,20 @@ def create_suites(args):  # pylint: disable=W0621
         suites.append(TestSuite.from_config(config_check_jobs, "jobs"))
 
     if args.dict_tests['functional']:
-        config_check_functional = copy.copy(config_check)
-        config_check_functional['resolver.references'] = ['selftests/functional/']
-        suites.append(TestSuite.from_config(config_check_functional,
-                                            "functional"))
+        functional_path = os.path.join('selftests', 'functional')
+        references = glob.glob(os.path.join(functional_path, 'test*.py'))
+        references.extend([os.path.join(functional_path, 'utils'),
+                           os.path.join(functional_path, 'plugin')])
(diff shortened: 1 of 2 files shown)


selftests/check.py
```diff
@@ -575,8 +575,18 @@ def create_suites(args):  # pylint: disable=W0621
...
+        references = glob.glob(os.path.join(functional_path, 'test*.py'))
+        references.extend([os.path.join(functional_path, 'utils'),
+                           os.path.join(functional_path, 'plugin')])
+        config_check['resolver.references'] = references
+        suites.append(TestSuite.from_config(config_check,
+                                            "functional-parallel"))
+
+        config_check['resolver.references'] = ['selftests/functional/serial/']
+        config_check['nrunner.max_parallel_tasks'] = 1
+        suites.append(TestSuite.from_config(config_check, "functional-serial"))
```
beraldoleal, line 588: IIUC, here you are using the same dict for two different suites. Due to the dict nature, one suite could update the dict and modify on the second suite. So, just to be a little bit safer, maybe a `copy` or `deep copy` here?

richtja: Yes, I can do that. I was reusing the approach from other types of tests, but I can update them as well.

richtja:
Hi @beraldoleal I did update with force-push.

beraldoleal (approved):
Thanks @richtja , LGTM.

ana (approved):

Issue #5200, opened 2021-12-17
Print a list of failed tests at the end of test summary


eskultety:
**Is your feature request related to a problem? Please describe.**
Consider the following execution
```
$ avocado run /bin/false /bin/true /bin/false
JOB ID      : latest
JOB LOG     : /home/gitlab-runner/avocado/job-results/job-2021-12-17T16.52-9b59568/job.log
SPAWNER     : process

Test ID      End Time            Run Time Status
2-/bin/true  1970/01/01 00:00:49 0.010845 PASS
3-/bin/false 1970/01/01 00:00:49 0.010837 FAIL
1-/bin/false 1970/01/01 00:00:49 0.011155 FAIL
RESULTS     : PASS 1 | ERROR 0 | FAIL 2 | SKIP 0 |WARN 0 | INTERRUPT 0 | CANCEL 0
```
From an output like ^this it is very simple to tell which tests passed and which failed, but do it for hundreds of tests and then one has to take scrollback and look for a needle in the haystack to find which tests have failed.

**Describe the solution you'd like**
 Since failed tests is all everyone cares about when looking at the results, it would be both very convenient and fast to just look at a summary which might look like this:
```
...
RESULTS     : PASS 1 | ERROR 0 | FAIL 2 | SKIP 0 |WARN 0 | INTERRUPT 0 | CANCEL 0
Failed tests:
   3-/bin/false
   1-/bin/false
```

**Describe alternatives you've considered**

1.  Having the following the config file:
```
[job.output.testlogs]
statuses = ['FAIL', 'ERROR']
```
Avocado will tell me what the logfile for a given failed tests contained which can be very verbose as the developer is primarily interested in the list of failed tests so that they can go and investigate on their own not mentioning the fact the the log files may not contain all the necessary data for an analysis.

2. grepping jobs
```
$ avocado jobs show | grep FAIL
```
This could be a somewhat satisfactory workaround, but you have to run 2 commands (run, jobs) which will produce 2 summaries with the latter being slightly confusing (naturally because of using grep):
```
$ avocado run /bin/false /bin/true /bin/false
...
RESULTS    : PASS 1 | ERROR 0 | FAIL 2 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0

$ avocado jobs show | grep FAIL
1-/bin/false 1970/01/01 00:13:55 0.010567 FAIL
3-/bin/false 1970/01/01 00:13:55 0.010627 FAIL
[...]

PR #5249, opened 2022-02-01
display failed tests at the end of a run

beraldoleal:
This will improve the readability when running multiple tests inside a
CI job without the need to run a second command or grep to find failed
tests. This is related to: #5200

### DIFF
--- a/avocado/plugins/testlogs.py
+++ b/avocado/plugins/testlogs.py
@@ -30,6 +30,16 @@ class TestLogsUIInit(Init):
                                  default=['debug.log'],
                                  help_msg=help_msg)
 
+        help_msg = ("Status that will trigger the output of a summary "
+                    "after the job ends. This is useful to list failed "
+                    "tests for instances at the end of a job run. "
+                    "Valid statuses: %s" % ", ".join(STATUSES))
+        settings.register_option(section='job.output.testlogs',
+                                 key='summary_statuses',
+                                 key_type=list,
+                                 default=['FAIL', 'ERROR'],
+                                 help_msg=help_msg)
+
 
 class TestLogsUI(JobPre, JobPost):
 
@@ -40,21 +50,25 @@ class TestLogsUI(JobPre, JobPost):
 
     def post(self, job):
         statuses = job.config.get('job.output.testlogs.statuses')
-        if not statuses:
-            return
+        summary = job.config.get('job.output.testlogs.summary_statuses')
 
         if job.config.get('stdout_claimed_by') is not None:
             return
 
         try:
-            with open(os.path.join(job.logdir, 'results.json'), encoding='utf-8') as json_file:
+            with open(os.path.join(job.logdir, 'results.json'),
+                      encoding='utf-8') as json_file:
                 results = json.load(json_file)
         except FileNotFoundError:
             return
 
         logfiles = job.config.get('job.output.testlogs.logfiles')
+        summary_tests = []
         for test in results['tests']:
-            if test['status'] not in statuses:
+            if test['status'] in summary:
+                line = f"{test['name']}: {test['status']}"
+                summary_tests.append(line)
+            if not statuses or test['status'] not in statuses:
                 continue
             for logfile in logfiles:
                 path = os.path.join(test['logdir'], logfile)
@@ -66,6 +80,12 @@ class TestLogsUI(JobPre, JobPost):
                 except (FileNotFoundError, PermissionError) as error:
                     LOG_UI.error('Failure to access log file "%s": %s',
                                  path, error)
+
+        if summary_tests:
+            LOG_UI.info("")
+            LOG_UI.info("Test summary:")
+            for test in sorted(summary_tests):
+                LOG_UI.error(test)
 
 
 class TestLogging(ResultEvents):
--- a/avocado/plugins/testlogs.py
+++ b/avocado/plugins/testlogs.py
@@ -43,6 +43,9 @@ class TestLogsUI(JobPre, JobPost):
         if not statuses:
             return
 
+        if job.config.get('stdout_claimed_by') is not None:
+            return
+
         try:
             with open(os.path.join(job.logdir, 'results.json'), encoding='utf-8') as json_file:
                 results = json.load(json_file)


richtja (changes requested):
Hi @beraldoleal thanks for this, it is working fine, and it meets the requirements from #5200, but I think we should resolve the connection between variable `failed_statuses` and config `job.output.testlogs.statuses`. 

Actual behavior:

- When I have `job.output.testlogs.statuses` empty I don't get any Failed tests summary. 

- When I have `job.output.testlogs.statuses = ['FAIL', 'ERROR']` it is working fine, I will get logs and also summary. 

- When I set `job.output.testlogs.statuses = ['SKIP']` it starting to be strange, I will get logs of `SKIP` tests and summary of  'FAIL' and 'ERROR' tests

IMO this is inconsistent and might be confusing for users. 

So I think we have two options:
1. create new config variable for failed test summary. Like `job.output.testlogs.statuses-failed`
2. Use `job.output.testlogs.statuses` as configuration for test summary, so with `job.output.testlogs.statuses = ['SKIP']` I will get summary of `SKIP` tests and with `job.output.testlogs.statuses = ['FAIL']` I will get summary of `FAIL` tests.

beraldoleal:
Hi @richtja I agree with you. I missed a return on line 45. I will try to come up with something on those lines. Thanks.

beraldoleal:
@richtja added a new option with a force-push. Let me know what you think.

richtja (approved):
@beraldoleal thanks for the changes, LGTM. I think we can use this solution in our selftests instead of [actual summary](https://github.com/avocado-framework/avocado/blob/master/selftests/check.py#L598), but that could be done in a different PR.

ana (approved):
Nice!

avocado/plugins/testlogs.py
```diff
@@ -63,6 +76,13 @@ def post(self, job):
                 except (FileNotFoundError, PermissionError) as error:
                     LOG_UI.error('Failure to access log file "%s": %s',
                                  path, error)
+        if summary_tests:
+            LOG_UI.info("")
+            LOG_UI.info(" Summary Output:")
```
eskultety, line 81: Possibly a better name could be "Test summary:"

avocado/plugins/testlogs.py
```diff
@@ -63,6 +76,13 @@ def post(self, job):
...
+            LOG_UI.info(" (Defined by 'job.output.testlogs.summary_statuses')")
```
eskultety, line 82: This feels like it's replacing the documentation, so IMO the message should be dropped and documentation should be added instead which I don't see in this patch.

eskultety (commented):
I like the output :). One question though, what's the reasoning for not making this the default summary behaviour and needs to be configured through the config first?

I'm also missing the information on which variable to tweak in the commit message, so I had to look at the code to figure out what should be done to test this

beraldoleal:
> I like the output :). One question though, what's the reasoning for not making this the default summary behaviour and needs to be configured through the config first?

I thought about this being default, and I think we can set "FAIL" and "ERROR" as default values. However I would like to remove this from being hard coded and keep it as a configuration option.

> I'm also missing the information on which variable to tweak in the commit message, so I had to look at the code to figure out what should be done to test this

I will fix on v2. thanks.

avocado/plugins/testlogs.py
beraldoleal: Documentation is auto generated and it will be available on `avocado config reference` and also in our rtd page. But I'm ok removing the message.

beraldoleal:
@eskultety fixed with a force - push. Thanks for the comments.

clebergnu (changes requested):
Hi Beraldo,

Even though the execution is not guaranteed to happen in the exact sequence given as test references to the resolver, I think it makes sense to present the test summary in the order given to the resolver (which means it will be numerically ordered).

Also, I don't get the reason for the `Test Summary` line to be indented.  IMO, it works like the other info (blue) lines, and should be at column 1 (instead of 2).

Oh, and CI is failing.

Other than that, LGTM.

avocado/plugins/testlogs.py
```diff
@@ -63,6 +76,12 @@ def post(self, job):
                 except (FileNotFoundError, PermissionError) as error:
                     LOG_UI.error('Failure to access log file "%s": %s',
                                  path, error)
+        if summary_tests:
+            LOG_UI.info("")
+            LOG_UI.info(" Test summary:")
+            for test in summary_tests:
```
ana, line 82: As you can see from the CI results, these lines are ending up in the TAP, json, xml files.

ana (changes requested):

avocado/plugins/testlogs.py
beraldoleal: Actually the files are right. Only the stdout is wrong. I'm fixing, thanks.

eskultety:
> > I like the output :). One question though, what's the reasoning for not making this the default summary behaviour and needs to be configured through the config first?
> 
> I thought about this being default, and I think we can set "FAIL" and "ERROR" as default values. However I would like to remove this from being hard coded and keep it as a configuration option.

What does it mean for the end-user? I'm asking because with this latest PR state, this is what I'm getting:
```
$ make develop
$ avocado run /bin/false /bin/true /bin/false
JOB ID     : c348d39b440091dcf2eb0220b4c8114e019dc950
JOB LOG    : /home/gitlab-runner/avocado/job-results/job-2022-02-16T17.27-c348d39/job.log
 (1/3) /bin/false: STARTED
 (2/3) /bin/true: STARTED
 (3/3) /bin/false: STARTED
 (1/3) /bin/false: FAIL (0.01 s)
 (2/3) /bin/true: PASS (0.01 s)
 (3/3) /bin/false: FAIL (0.01 s)
RESULTS    : PASS 1 | ERROR 0 | FAIL 2 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB HTML   : /home/gitlab-runner/avocado/job-results/job-2022-02-16T17.27-c348d39/results.html
JOB TIME   : 1.14 s
```
not even with
```
$ cat /tmp/avocado.conf
[job.output.testlogs]
summary_statuses=['FAIL','ERROR']
$ avocado --config /tmp/avocado.conf run /bin/false /bin/true /bin/false
JOB ID     : 75809b1a1f4f81d9046113a0832561e769bcaa32                                                     
JOB LOG    : /home/gitlab-runner/avocado/job-results/job-2022-02-16T17.30-75809b1/job.log
 (1/3) /bin/false: STARTED                                                                                
 (2/3) /bin/true: STARTED               
 (3/3) /bin/false: STARTED 
 (1/3) /bin/false: FAIL (0.01 s)                                                                          
 (2/3) /bin/true: PASS (0.01 s)                                                                           
 (3/3) /bin/false: FAIL (0.01 s)                                                                          
RESULTS    : PASS 1 | ERROR 0 | FAIL 2 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB HTML   : /home/gitlab-runner/avocado/job-results/job-2022-02-16T17.30-75809b1/results.html
JOB TIME   : 1.11 s    
```

What am I missing?

beraldoleal:
@eskultety, you are not missing anything. It means that I forgot to update the last commit :)

ana:
Works now:

```
$  avocado run /bin/false /bin/true /bin/false
JOB ID     : 7fd5d4b7ad57df1b3e38515d8735a6a1b66b26e2
JOB LOG    : /home/ana/avocado/job-results/job-2022-02-21T12.13-7fd5d4b/job.log
 (1/3) /bin/false: STARTED
 (2/3) /bin/true: STARTED
 (3/3) /bin/false: STARTED
 (1/3) /bin/false: FAIL (0.01 s)
 (2/3) /bin/true: PASS (0.01 s)
 (3/3) /bin/false: FAIL (0.01 s)
RESULTS    : PASS 1 | ERROR 0 | FAIL 2 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB HTML   : /home/ana/avocado/job-results/job-2022-02-21T12.13-7fd5d4b/results.html
JOB TIME   : 0.88 s

Test summary:
/bin/false: FAIL
/bin/false: FAIL
```

ana (approved):

eskultety:
I can confirm it now works with no configuration, thank you. FWIW I approve of the change.

beraldoleal:
@clebergnu ?

eskultety:
ping @clebergnu

clebergnu (approved):
LGTM, thanks!

PR #5251, opened 2022-02-01
core.suite: remove legacy runner leftovers

beraldoleal:
This is a preparation for a more drastical removal here.
Related to #5246

### DIFF
--- a/avocado/core/suite.py
+++ b/avocado/core/suite.py
@@ -25,7 +25,7 @@ from avocado.core.parser import HintParser
 from avocado.core.resolver import ReferenceResolutionResult, resolve
 from avocado.core.settings import settings
 from avocado.core.tags import filter_test_tags, filter_test_tags_runnable
-from avocado.core.test import DryRunTest, Test
+from avocado.core.test import DryRunTest
 from avocado.core.tree import TreeNode
 from avocado.core.varianter import Varianter, is_empty_variant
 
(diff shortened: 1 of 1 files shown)


ana (approved):

avocado/core/suite.py
```diff
@@ -277,9 +259,7 @@ def size(self):
     def stats(self):
         """Return a statistics dict with the current tests."""
         runner_name = self.config.get('run.test_runner')
-        if runner_name == 'runner':
-            return self._get_stats_from_runner()
-        elif runner_name == 'nrunner':
+        if runner_name == 'nrunner':
             return self._get_stats_from_nrunner()
```
richtja, line 263: Do we still need `if runner_name == 'nrunner'` here? Also, I think we can rename ` _get_stats_from_nrunner` to just `_get_stats` or just move the stats computing here. What do you think?

avocado/core/suite.py
```diff
@@ -298,9 +278,7 @@ def status(self):
     def tags_stats(self):
         """Return a statistics dict with the current tests tags."""
         runner_name = self.config.get('run.test_runner')
-        if runner_name == 'runner':
-            return self._get_tags_stats_from_runner()
-        elif runner_name == 'nrunner':
+        if runner_name == 'nrunner':
             return self._get_tags_stats_from_nrunner()
```
richtja, line 282: Same here

richtja (commented):
Hi @beraldoleal thanks for this effort, the changes LGTM. I just add comments that maybe we can remove more of it.

avocado/core/suite.py
beraldoleal: the if is there because of the same reason it was before. For some reason sometimes this value is not set. I can do a proper refactoring here after we remove the runner, right now I would like to just remove old references.

richtja: Ok, I am fine with leaving it like it is. can you please create an issue about this problem?

richtja (approved):

PR #5252, opened 2022-02-01
utils/test_network: fixed test on ipv6 disabled

beraldoleal:
Currently, this test is failing when IPV6 is disabled on the host
machine. Since we currently lack a proper skipIf decorator here, I'm
using the already defined catch/logic to ignore when IPv6 is not
supported.

### DIFF
--- a/selftests/unit/utils/test_network.py
+++ b/selftests/unit/utils/test_network.py
@@ -83,7 +83,7 @@ class FreePort(unittest.TestCase):
                             good.append("%s, %s, %s" % (family, protocol,
                                                         addr))
                     except Exception as exc:
-                        if getattr(exc, 'errno', None) in (-2, 2, 22, 94):
+                        if getattr(exc, 'errno', None) in (-2, 2, 22, 94, 99):
                             skip.append("%s, %s, %s: Not supported: %s"
                                         % (family, protocol, addr, exc))
                         else:


ana:
Hi @beraldoleal 
I tested without IPV6 trying to understand this change and I couldn't reproduce the original problem.

beraldoleal:
> Hi @beraldoleal I tested without IPV6 trying to understand this change and I couldn't reproduce the original problem.

Have you tried to disable ipv6 on your machine, rebooted and the test is still passing?

ana:
> > Hi @beraldoleal I tested without IPV6 trying to understand this change and I couldn't reproduce the original problem.
> 
> Have you tried to disable ipv6 on your machine, rebooted (or force applied) and the test is still passing?

Yes, I disabled ipv6 in `/etc/sysctl.conf`, rebooted without ipv6 and test passed. I even went to remove fully networking from my VM and the tests still passed.

beraldoleal:
> > > Hi @beraldoleal I tested without IPV6 trying to understand this change and I couldn't reproduce the original problem.
> > 
> > 
> > Have you tried to disable ipv6 on your machine, rebooted (or force applied) and the test is still passing?
> 
> Yes, I disabled ipv6 in `/etc/sysctl.conf`, rebooted without ipv6 and test passed. I even went to remove fully networking from my VM and the tests still passed.

Are you using the host OS or a container? Using container, bind() call will not fail when ipv6 is disabled. But when using a base OS it is failing during bind():

```
$ ip a | grep inet6
$ ipython3
In [1]: import socket                                                                                                  

In [2]: sock = socket.socket(socket.AF_INET6, socket.SOCK_STREAM)                                                      

In [3]: sock.bind(('localhost', 2121))                                                                                 
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
<ipython-input-3-9c024125f5f7> in <module>
----> 1 sock.bind(('localhost', 2121))

OSError: [Errno 99] Cannot assign requested address
```

Probably this test should be refactored but since it is already catching some "exception" on "failed to bind", I'm just adding the new code.

clebergnu (commented):
Hi @beraldoleal,

There are a number of system configuration that may match the "IPv6 disabled" criteria.  I tried it with the "ipv6.disabled=1" kernel argument, and the behavior I found was a bit different from what you found.

For instance, the error code returned when trying to create the socket is 97:

```
>>> import socket
>>> sock = socket.socket(socket.AF_INET6, socket.SOCK_STREAM)   
  File "/usr/lib64/python3.9/socket.py", line 232, in __init__
    _socket.socket.__init__(self, family, type, proto, fileno)
OSError: [Errno 97] Address family not supported by protocol
```

Also, this breaks the test in a very different way.  The test takes a few minutes to run, because `is_port_free()` always fails, and thus, `find_free_ports()` keeps asking about the free status for every single port in the default range.  Then, `port` gets assigned `None`, and finally it errors with:

```
  File "/home/cleber/src/avocado/avocado/selftests/unit/utils/test_network.py", line 63, in test_is_port_free
    self.assertTrue(ports.is_port_free(port, "localhost"))
  File "/home/cleber/src/avocado/avocado/avocado/utils/network/ports.py", line 54, in is_port_free
    sock.bind(("", port))
TypeError: an integer is required (got type NoneType)
```

So, maybe just one new error code doesn't do a lot to improve this test.  I'm always open to incremental improvements, but maybe this case deserves a more comprehensive overhaul to avoid a bit of churning.

WDYT?

beraldoleal:
@clebergnu indeed.

Ana's report and yours shows that this is not a easy fix for now because of the different results depending on the system/container and how this is disabled. In fact, maybe the library itself it will need some changes. I will close this and create an issue, since this is a very specific case and it is not breaking our CI, for now.

Issue #5254, opened 2022-02-04
Check bandit to replace codeclimate


ana:
Lots of the checkes made by codeclimate are from bandit https://docs.codeclimate.com/docs/bandit
We currently run bandit under the experimenal checks, it'd would be interesting to check the configuration options and enable the useful ones.

PR #5256, opened 2022-02-04
Add bandit to the static-checks

ana:
Bandit is a tool designed to find common security issues in Python code. This addresses some of the funcionationalities we had with codeclimate.

References: https://github.com/avocado-framework/avocado/issues/5254

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -96,3 +96,24 @@ jobs:
           name: job-results-without-plugins
           path: /home/runner/avocado/job-results/
       - run: echo "🥑 This job's status is ${{ job.status }}."
+
+
+  full-bandit-checks:
+    name: Bandit checks
+    runs-on: ubuntu-20.04
+    steps:
+      - run: echo "Job triggered by a ${{ github.event_name }} event on branch is ${{ github.ref }} in repository is ${{ github.repository }}, runner on ${{ runner.os }}"
+      - name: Check out repository code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Run bandit check
+        run: make -f Makefile.gh bandit
+        continue-on-error: True
+      - name: Save bandit output as artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: bandit-results
+          path: /home/runner/work/avocado/avocado/bandit-output.txt
+          retention-days: 1
+      - run: echo "🥑 This job's status is ${{ job.status }}."
--- a/examples/tests/sleeptenmin.py
+++ b/examples/tests/sleeptenmin.py
@@ -1,4 +1,4 @@
-import os
+import subprocess
 import time
 
 from avocado import Test
@@ -29,4 +29,4 @@ class SleepTenMin(Test):
             if method == 'builtin':
                 time.sleep(length)
             elif method == 'shell':
-                os.system(f"sleep {length}")
+                subprocess.run(["sleep", length], check=True)
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -12,6 +12,7 @@ coverage==5.5
 # To run make check
 aexpect==1.6.1
 psutil==5.8.0
+bandit==1.7.2; python_version > '3.6'
 
 # pycdlib is an optional requirement in production
 # but is necessary for selftests
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -218,15 +218,6 @@ jobs:
       - name: Run Codespell Check
         run: make -f Makefile.gh codespell
         continue-on-error: True
-      - name: Run bandit check
-        run: make -f Makefile.gh bandit
-        continue-on-error: True
-      - name: Save bandit output as artifacts
-        uses: actions/upload-artifact@v2
-        with:
-          name: bandit-results
-          path: /home/runner/work/avocado/avocado/bandit-output.txt
-          retention-days: 1
       - run: echo "🥑 This job's status is ${{ job.status }}."
 
 
--- a/avocado/core/output.py
+++ b/avocado/core/output.py
@@ -19,6 +19,7 @@ import errno
 import logging
 import os
 import re
+import subprocess
 import sys
 import traceback
 
@@ -593,7 +594,7 @@ class Paginator:
             except utils_path.CmdNotFoundError as details:
                 raise RuntimeError("Unable to enable pagination: %s" % details)
 
-        self.pipe = os.popen(paginator, 'w')
+        self.pipe = subprocess.Popen(paginator, stdin=subprocess.PIPE)
 
     def __del__(self):
         self.close()
@@ -601,7 +602,7 @@ class Paginator:
     def close(self):
         if self.pipe:
             try:
-                self.pipe.close()
+                self.pipe.stdin.close()
             except OSError:
                 pass
 
--- a/avocado/utils/process.py
+++ b/avocado/utils/process.py
@@ -635,7 +635,7 @@ class SubProcess:
             self._popen = subprocess.Popen(cmd,
                                            stdout=subprocess.PIPE,
                                            stderr=subprocess.PIPE,
-                                           shell=self.shell,
+                                           shell=self.shell,  # nosec
                                            env=self.env)
         except OSError as details:
             details.strerror += " (%s)" % self.cmd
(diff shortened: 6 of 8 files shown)


examples/tests/sleeptenmin.py
```diff
@@ -29,4 +29,4 @@ def test(self):
             if method == 'builtin':
                 time.sleep(length)
             elif method == 'shell':
-                subprocess.run(["sleep", length])
+                subprocess.run(["sleep", length], check=True)
```
clebergnu, line 32: This seems to not be related to this commit.  Also, it did not influence the bandits results at all.  Maybe did you mean `shell=True` as an standalone change because the method name is "shell"?

clebergnu (changes requested):
Overall looks good, but I did not understand and thus agree with the `check=True` change.

examples/tests/sleeptenmin.py
ana: This was a late fix detected by pylint (W1510) that went to the wront commit, I moved it now to the proper one.
Answering your question, I really meant `check=True`, it's required to make clear the error behaviour. This way it'll raise an exception when the return of the command is non-zero.

ana:
I have force-pushed an update. I move one change to the good commit and updated  https://github.com/avocado-framework/avocado/pull/5256/commits/976ec577063a5e91dc10fcb401f2519d99ca92a5 to fix a problem detected by Travis.

richtja (approved):
LGTM

beraldoleal (changes requested):
LGTM, however I left one small request.

.github/workflows/weekly.yml
```diff
@@ -96,3 +96,24 @@ jobs:
...
+    name: Bandit checks
+    runs-on: ubuntu-20.04
+    steps:
+      - run: echo "Job triggered by a ${{ github.event_name }} event on branch is ${{ github.ref }} in repository is ${{ github.repository }}, runner on ${{ runner.os }}"
+      - name: Check out repository code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Run bandit check
+        run: make -f Makefile.gh bandit
```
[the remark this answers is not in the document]
ana: @beraldoleal it's doable but I find that too complicated for little benefit. Personally, I'd rather add the command line directly in the weekly github action.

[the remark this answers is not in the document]
beraldoleal: >  it's doable but I find that too complicated for little benefit. Personally, I'd rather add the command line directly in the weekly github action.

I have the hardship to find this "too difficult."  This was what I had in mind:

```bash
if [ -z "${AVOCADO_BANDIT_FULL}" ]; then
    echo "Running bandit in full mode"
    bandit -o bandit-output.txt -r --skip B101,B105,B311,B404,B603 .
else
    echo "Running bandit in high-severity mode"
    bandit -r -lll .
fi
```

Something like this would concentrate all bandit logic in a single file and inside "selftests" directory since this is a "selftest". This will make it easier for new contributors to understand the "bandit modes" we are running. 

The way it is in this PR, we have a shell script called "selftests/check-bandit.sh" giving the impression that all bandit checks related are there. If someone is reading only the `check-bandit.sh` or `Makefile.gh`, they will not be aware of all different "modes" we have.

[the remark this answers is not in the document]
beraldoleal: Regardless of having a Makefile or not, I'm still standing with my argument of concentrating this all inside "selftests/check-bandit.sh".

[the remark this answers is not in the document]
clebergnu: So far we have:

* all files that match `selftests/*.sh` being run on every `make check`
* all CI actions that are *not* part of `make check` being put into a `Makefile`

I'm also bothered by having different commands at different places, but this PR is coherent with the logic described above.

examples/tests/sleeptenmin.py
clebergnu: OK, a quick note about the why of the `check=True` change on the commit message would be helpful.  Also, there's a typo in the commit message (deprecrated).

clebergnu (commented):
Hi @ana,

I just spotted a typo and suggest adding a note to the commit about the reason of adding `check=True`.  Other than that, it LGTM.

.github/workflows/weekly.yml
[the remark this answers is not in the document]
beraldoleal: Sorry for insisting on this, but I'm still not convinced that the logic described is enough.

As suggested, moving the actual bandit execution inside `selftests/bandit-check.sh` would keep the logic above and also solve the part that bothers us. IMO having only those two bullets as "accepted logic" is risky because it leaves room for dispersal executions in different files, making code maintenance and reading difficult.

Currently we are "starting loosing" this: right now, in order to know what is being skipped or how bandit is being executed, I have to look at two files (Makefile.gh and selftests/check-bandint.sh). And even those are executed in different contexts, they are both "selftests".

Also, `requirements-dev.txt` has a specific bandit version while `Makefile.gh` is installing just "bandit", leaving room for inconsistent bandit versions being installed.

requirements-dev.txt
```diff
@@ -12,6 +12,7 @@ coverage==5.5
 # To run make check
 aexpect==1.6.1
 psutil==5.8.0
+bandit==1.7.2; python_version > '3.6'
```
beraldoleal, line 15: So, this means that when in python3.6 `selftests/bandit-check.sh` it will fail?

beraldoleal (changes requested):

requirements-dev.txt
ana: This means bandit will be only installed in Python 3.7 and superior.
What change are you requesting here?

.github/workflows/weekly.yml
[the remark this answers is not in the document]
ana: bandit in makefile.gh could go away because once we have this checking in every PR the weekly check should go away.

Beraldo, I'm sorry you're not convinced, but eventually you need to evaluate if blocking something that can help the quality of our code is worthwhile so much discussion. I'm pretty sure all this part of the code (checks) is going to evolve in the next year.

requirements-dev.txt
beraldoleal: At this point I'm not requesting any changes, @ana. I'm just trying to understand why it is ok to fail this when running on python3.6:

A `make check` and individual run of `selftests/bandit-check.sh` it will fail because bandit it will not be installed.

ana: > A `make check` and individual run of `selftests/bandit-check.sh` it will fail because bandit it will not be installed.

Yeah, at this stage I hope no one is using Python 3.6 for developing.

clebergnu: Ooops, I did not catch that, but you're right.

The way I see it, we either add another special condition (I don't like special conditions), or have Python 3.6 removed before this is applied.

BTW, John Snow (not tagging him to avoid noise) was considering pushing a strategy for having Python 3.7+ stacks installed in the Docker images that build QEMU on platforms still with Python 3.6.

The other possibility is to declare Avocado 3.7+ already, and protect the `make check-avocado` execution on QEMU against lesser versions.

.github/workflows/weekly.yml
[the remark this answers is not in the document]
beraldoleal: (I know the PR is closed, but just in case of a future v2)

> bandit in makefile.gh could go away because once we have this checking in every PR the weekly check should go away.

I'm reviewing PRs as it is. And whenever possible, I like to avoid thinking about future changes because I have learned the hard way that technical debits are easily forgotten, especially those "easy and for internal use only".

> Beraldo, I'm sorry you're not convinced, but eventually you need to evaluate if blocking something that can help the quality of our code is worthwhile so much discussion. I'm pretty sure all this part of the code (checks) is going to evolve in the next year.

Part of the submission process is not only convincing reviewers that your idea is good but also that the location, names, and design decisions are also good for the project. It is a slow process, and I understand that sometimes it is frustrating for who is submitting, but I like to think it is healthier in the end. I'm not saying that we can't change our minds and refactor later; of course, we can (I "unblocked" many PRs, after a good argument). I'm just saying that if we keep going in "this mode," we will quickly increase coupling and lose cohesion inside the project. We could avoid that.

I'm not buying the argument that a feature should be accepted without a proper discussion about where it should be placed just because the feature itself improves code quality.

PR #5258, opened 2022-02-07
pre-release-test: fixes to pass

beraldoleal:
Fedora copr builds are failing because HTML plugin is not enabled.
Because of that, html outputs are not generated. I could not reproduce
this locally yet, and couldn't find any other hint on debug.log. There
is a mention to "package init file 'avocado_result_html/__init__.py' not
found (or not a regular file)" and this *might be* the root cause.

Probably this is is not the best solution, but I would like to check if
copr build will pass until we find the best one.

I'm also, doing a small fix on our pre-release plan.


ana:
@beraldoleal Let's see if I understand the problem: the html is not properly generated when running it from the fedora packages? if so, which fedora version?

beraldoleal:
> @beraldoleal Let's see if I understand the problem: the html is not properly generated when running it from the fedora packages? if so, which fedora version?

copr is trying to build this on centos-stream-8 and it is failing, logs are here:

https://download.copr.fedorainfracloud.org/results/@avocado/avocado-latest/centos-stream-8-aarch64/03294756-python-avocado/builder-live.log.gz

ana:
> > @beraldoleal Let's see if I understand the problem: the html is not properly generated when running it from the fedora packages? if so, which fedora version?
> 
> copr is trying to build this on centos-stream-8 and it is failing, logs are here:
> 
> https://download.copr.fedorainfracloud.org/results/@avocado/avocado-latest/centos-stream-8-aarch64/03294756-python-avocado/builder-live.log.gz

Ok, in that case replace the first commit with:

`git revert 2e426bd81395444b6bbfb771f3af995776e87cfc`

beraldoleal:
> 2e426bd81395444b6bbfb771f3af995776e87cfc

Yes, makes sense. I was just trying to make sure this makes sense. Thanks.

ana (approved):

richtja (approved):
LGTM

PR #5260, opened 2022-02-07
bandit: fixes empty output file

beraldoleal:
Currently when running make -f Makefile.gh bandit it is not producing
any data inside output file because of the default format. Bandit
suggests using -f txt for that.

### DIFF
--- a/Makefile.gh
+++ b/Makefile.gh
@@ -23,7 +23,7 @@ codespell:
 
 bandit:
 	pip install bandit
-	bandit -o bandit-output.txt -r --skip B101,B105,B311,B404,B603 .
+	bandit -f txt -o bandit-output.txt -r --skip B101,B105,B311,B404,B603 .
 
 
 propagate-version:


ana:
There is not empty file, maybe a misleading error message. If you check in the artifacts, for example at https://github.com/avocado-framework/avocado/actions/runs/1806849881 the file is there.

ana:
> make -f Makefile.gh bandit

yes, because you get the output in your screen.

beraldoleal:
> > There is something strange here. After a `make -f Makefile.gh bandit` I get:
> > ```
> > $ cat bandit-output.txt 
> > $ wc -l bandit-output.txt 
> > 0 bandit-output.txt
> > ```
> 
> yes, because you get the output in your screen.

What is a very odd combination (-f vs -o) in my option, bandit--. But I guess I can leave with that, since this is the bandit way of doing things. Thanks for the clarification.

ana:
> What is a very odd combination (-f vs -o) in my option, bandit--. But I guess I can leave with that, since this is the bandit way of doing things. Thanks for the clarification.

There are a few things like this. I see it as a quick replacement for codeclimate for at least pointing the big issues. We'll have to think/explore what we can  do for security.

beraldoleal:
> > What is a very odd combination (-f vs -o) in my option, bandit--. But I guess I can leave with that, since this is the bandit way of doing things. Thanks for the clarification.
> 
> There are a few things like this. I see it as a quick replacement for codeclimate for at least pointing the big issues. We'll have to think/explore what we can do for security.

IIUC, bandit covers only the security aspect, right? If that is the case, I don't see it as a replacement of codeclimate that was looking into multiple code quality metrics.

ana:
> > > What is a very odd combination (-f vs -o) in my option, bandit--. But I guess I can leave with that, since this is the bandit way of doing things. Thanks for the clarification.
> > 
> > 
> > There are a few things like this. I see it as a quick replacement for codeclimate for at least pointing the big issues. We'll have to think/explore what we can do for security.
> 
> IIUC, bandit covers only the security aspect, right? If that is the case, I don't see it as a replacement of codeclimate that was looking into multiple code quality metrics.

No, it's not a full replacement for codeclimate, that's why I said "for at least pointing the big (security) issues"

Issue #5219, opened 2022-01-03
Switch to Python f-strings


ana:
Most of the strings are formatted with `.format()` or `'%'`, it would be great to use f-strings for string formatting when possible.

Pylint can help detecting the strings to update. Starting on version 2.11.0 it included a check `c0209: formatting a regular string which could be a f-string (consider-using-f-string)`.

[This link](https://miguendes.me/73-examples-to-help-you-master-pythons-f-strings) provides a good guide on how to update the strings.

ozer550:
Hi there, I changed possible .format strings to f-strings but there were more than 100 instances of it, is there a way through which we can automate this process for %.

ozer550:
Also, c0209 check was disabled is there any specific reason for it?

richtja:
> Hi there, I changed possible .format strings to f-strings but there were more than 100 instances of it, is there a way through which we can automate this process for %.

Hi @ozer550 unfortunately I think there is no way how to automatize this process, and we need to go through the all instances. :confused:

> Also, c0209 check was disabled is there any specific reason for it?

This check is disabled,  because we are still using formatting a regular string and this will make our CI failing. When this issue will be solved, we can enable that check.

al5929:
Hello! I can try to tackle this challenge.
I've followed the contribution procedures and got pylint working. I'll update the strings in accordance with C0209

richtja:
Hi @al5929 and welcome to the avocado community, we will appreciate your contribution. Right now, there is PR #5267 from @ana about f-strings in `contrib` and `examples` directories. Also, there is draft #5266 from @ozer550 who is working on this issue right now, and I am not sure how far he is right now. So please coordinate your work with @ozer550. If you need any help, don't hesitate to ask, we will do our best to help you.

Because this is a huge refactoring, I can imagine splitting the work by `core`, `utils` and `plugins` packages, for easier coordination between contributors. @ozer550 what do you think?

ana:
Hi @al5929 
As @richtja  mentioned, we're waiting to hear back from ozer550. This change can be made almost automatically , see https://github.com/avocado-framework/avocado/pull/5267 so there is little challenge. 
If you're interested to help and also learn, I can give you a more interesting task.

al5929:
@richtja @ana 
Okay, I understand. I'm still pretty new to the whole contribution process, but I can definitely look into other issues.

ozer550:
[...]

PR #5274, opened 2022-03-08
selftests: switch to Python f-strings

ana:
Switching `selftests/ `  to see everything goes fine before I start looking into` avocado/`

References: https://github.com/avocado-framework/avocado/issues/5219

### DIFF
--- a/selftests/functional/plugin/spawners/test_process.py
+++ b/selftests/functional/plugin/spawners/test_process.py
@@ -20,12 +20,11 @@ class ProcessSpawnerTest(TestCaseTmpDir):
         test = script.Script(os.path.join(self.tmpdir.name, "logdir_test.py"),
                              TEST_LOGDIR)
         test.save()
-        result = process.run("%s run --job-results-dir %s --disable-sysinfo "
-                             "--json - -- %s" % (AVOCADO, self.tmpdir.name,
-                                                 test))
+        result = process.run(f"{AVOCADO} run --job-results-dir {self.tmpdir.name}"
+                             f"--disable-sysinfo --json - -- {test}")
         res = json.loads(result.stdout_text)
         logfile = res["tests"][0]["logfile"]
         testdir = res["tests"][0]["logdir"]
         with open(logfile, 'r', encoding='utf-8') as debug_file:
-            expected = "logdir is: %s" % testdir
+            expected = f"logdir is: {testdir}"
             self.assertIn(expected, debug_file.read())
--- a/selftests/functional/plugin/spawners/test_process.py
+++ b/selftests/functional/plugin/spawners/test_process.py
@@ -20,7 +20,8 @@ class ProcessSpawnerTest(TestCaseTmpDir):
         test = script.Script(os.path.join(self.tmpdir.name, "logdir_test.py"),
                              TEST_LOGDIR)
         test.save()
-        result = process.run(f"{AVOCADO} run --job-results-dir {self.tmpdir.name}"
+        result = process.run(f"{AVOCADO} run "
+                             f"--job-results-dir {self.tmpdir.name}"
                              f"--disable-sysinfo --json - -- {test}")
         res = json.loads(result.stdout_text)
         logfile = res["tests"][0]["logfile"]
--- a/selftests/functional/plugin/test_assets.py
+++ b/selftests/functional/plugin/test_assets.py
@@ -75,14 +75,12 @@ class AssetsFetchSuccess(TestCaseTmpDir):
         test_file.write(test_content.encode())
         test_file.close()
 
-        expected_output = "Fetching assets from %s.\n" \
-            "  File hello-2.9.tar.gz fetched or already on cache.\n" \
-            % test_file.name
+        expected_output = (f"Fetching assets from {test_file.name}.\n"
+                           f"  File hello-2.9.tar.gz fetched or already on cache.\n")
         expected_rc = exit_codes.AVOCADO_ALL_OK
 
-        cmd_line = "%s --config %s assets fetch %s " % (AVOCADO,
-                                                        self.config_file.name,
-                                                        test_file.name)
+        cmd_line = (f"{AVOCADO} --config {self.config_file.name} "
+                    f"assets fetch {test_file.name} ")
         result = process.run(cmd_line)
         os.remove(test_file.name)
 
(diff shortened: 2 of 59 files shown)


beraldoleal (changes requested):
Hi @ana, those changes LGTM, and are welcome, thanks.

I just have one thing that bothers me: the lines length. ;( I found a few cases this change is replacing with a line longer than 79 columns, not recommend by PEP8.

Although this isn't in our contributor guide I feel like we've been following this pattern for quite some time now, and I confess it bothers me a little if we loosen this limitation.

richtja:
> Hi @ana, those changes LGTM, and are welcome, thanks.
> 
> I just have one thing that bothers me: the lines length. ;( I found a few cases this change is replacing with a line longer than 79 columns, not recommend by PEP8.
> 
> Although this isn't in our contributor guide I feel like we've been following this pattern for quite some time now, and I confess it bothers me a little if we remove this limitation.

@beraldoleal I agree with you, and I think this should be part of our contributor guide. Also, it looks like we are missing the whole [code style guide](https://avocado-framework.readthedocs.io/en/latest/guides/contributor/chapters/styleguides.html#code-style-guide).

ana:
> Hi @ana, those changes LGTM, and are welcome, thanks.
> 
> I just have one thing that bothers me: the lines length. ;( I found a few cases this change is replacing with a line longer than 79 columns, not recommend by PEP8.
> 
> Although this isn't in our contributor guide I feel like we've been following this pattern for quite some time now, and I confess it bothers me a little if we remove this limitation.

Hi @beraldoleal
Thanks for looking at this. If you want to enforce the max line length, the best is you enable the corresponding pylint check C0301  ( https://github.com/avocado-framework/avocado/blob/master/.pylintrc#L84 ) and update the value at https://github.com/avocado-framework/avocado/blob/master/.pylintrc#L385
Like enabling this pylint check, it'll require some work.

While I think enabling that pylint check is a good idea, I don't think that change should be coupled with this one or block it.

beraldoleal:
> > Hi @ana, those changes LGTM, and are welcome, thanks.
> > I just have one thing that bothers me: the lines length. ;( I found a few cases this change is replacing with a line longer than 79 columns, not recommend by PEP8.
> > Although this isn't in our contributor guide I feel like we've been following this pattern for quite some time now, and I confess it bothers me a little if we remove this limitation.
> 
> Hi @beraldoleal Thanks for looking at this. If you want to enforce the max line length, the best is you enable the corresponding pylint check C0301 ( https://github.com/avocado-framework/avocado/blob/master/.pylintrc#L84 ) and update the value at https://github.com/avocado-framework/avocado/blob/master/.pylintrc#L385 Like enabling this pylint check, it'll require some work.
> 
> While I think enabling that pylint check is a good idea, I don't think that change should be coupled with this one or block it.

I might be wrong, but the way I see this is: C0301 checks are disabled now because we would need some work in order to enable it (as you mention) but not because we are in favor of having lines > 79 (I saw multiple reviews where we asked to reduce the line number to keep this convention). The same way we are disabling W0703 and many others there. We should avoid them whenever possible to make the job of reducing this list easier instead of harder when someone decides to do it.

If accepting f-strings over 79 columns is a common-sense here, I will be glad to remove the blocker flag. However, because of the fuzzy convention we have now (as noticed by @richtja) I think this should be voted/discussed among the core devs first.

@clebergnu or any other dev, comments?

richtja (commented):
Except the @beraldoleal  concerns and conflicts with 62d8ab41f9118e191645bc69f38c7c21b0aa42e6, this LGTM.

ana:
@beraldoleal I just checked while filing issue https://github.com/avocado-framework/avocado/issues/5285 and I haven't introduce any new longer line that wasn't already there with this PR.

ana:
@richtja Fixed the conflict, thanks for pointing it out!

selftests/functional/plugin/spawners/test_process.py
```diff
@@ -20,12 +20,11 @@ def test_logdir_path(self):
         test = script.Script(os.path.join(self.tmpdir.name, "logdir_test.py"),
                              TEST_LOGDIR)
         test.save()
-        result = process.run("%s run --job-results-dir %s --disable-sysinfo "
-                             "--json - -- %s" % (AVOCADO, self.tmpdir.name,
-                                                 test))
+        result = process.run(f"{AVOCADO} run --job-results-dir {self.tmpdir.name}"
```
beraldoleal, line 23: When I made my comment, I was talking about this one: 77 vs 82.

selftests/functional/plugin/test_assets.py
```diff
@@ -75,14 +75,12 @@ def test_asset_fetch_success(self):
         test_file.write(test_content.encode())
         test_file.close()
 
-        expected_output = "Fetching assets from %s.\n" \
-            "  File hello-2.9.tar.gz fetched or already on cache.\n" \
-            % test_file.name
+        expected_output = (f"Fetching assets from {test_file.name}.\n"
+                           f"  File hello-2.9.tar.gz fetched or already on cache.\n")
```
beraldoleal, line 79: This one, 85.

selftests/functional/plugin/test_assets.py
```diff
@@ -237,12 +219,10 @@ def test_asset_fetch_unsupported_class(self):
...
         test_file.close()
 
-        expected_stdout = "Fetching assets from %s.\n" % test_file.name
+        expected_stdout = f"Fetching assets from {test_file.name}.\n"
         expected_rc = exit_codes.AVOCADO_ALL_OK
 
-        cmd_line = "%s --config %s assets fetch %s " % (AVOCADO,
-                                                        self.config_file.name,
-                                                        test_file.name)
+        cmd_line = f"{AVOCADO} --config {self.config_file.name} assets fetch {test_file.name} "
```
beraldoleal, line 225: This one, 95.

beraldoleal:
@ana I found a few, that I comment here. But probably there is more.

beraldoleal (changes requested):

selftests/functional/plugin/spawners/test_process.py
ana: sigh, I was mistaken by the fact that we run pylint *twice*  with different options in `selftests/lint.sh `

ana:
@beraldoleal @richtja I have fixed the long lines introduce by my commit and probably some extra ones. Could you please take a look?

richtja (approved):
LGTM, thanks

beraldoleal (approved):
Thanks @ana for the changes.

PR #5276, opened 2022-03-09
Fix is_port_free() on IPv6-less systems

huth:
The is_port_free() function currently iterates through all combinations
of address families and protocols, and does not return immediately once
a free port has been found. So it only works on systems where a free
AF_INET6 + SOCK_DGRAM port is available, thus it's failing completely on
systems where IPv6 is not available.
Fix it by returning immediately once a free port has been found.

### DIFF
--- a/avocado/utils/network/ports.py
+++ b/avocado/utils/network/ports.py
@@ -60,7 +60,7 @@ def is_port_free(port, address):
                     if localhost:
                         return False
                 sock.close()
-        return True
+                return True
     finally:
         if sock is not None:
             sock.close()


ana:
Thanks for the PR @huth 
Please, take a second look because the testsuite didn't like your change.

huth:
> Thanks for the PR @huth Please, take a second look because the testsuite didn't like your change.

Ok, please never mind this pull request ... seems like both, the ports.py code and the unit tests are just FUBAR, they assume that TCP and UDP share the same port space, which is just completely wrong. Looks like all the related code to find_free_port() and is_port_free() needs a major rewrite, so this certainly can't be cured with a one-liner like this.

PR #5288, opened 2022-03-17
Selftests summary of failed tests removal

richtja:
After the 1bc8f7866967125e4a91dbca71dc4198532ab4e7 the failed tests
summary is not needed in selftest anymore, and it also creates
duplicates in output. Let's remove it.


beraldoleal (approved):
Makes sense. Thanks.

PR #5294, opened 2022-03-23
utils.network.ports refactoring

### LINKED ISSUE #5273: utils: Improve how we get a free port (without brute force)
Currently, `utils.find_free_port()` is brute forcing for a free port. According to @huth about QEMU tests: _"..this takes > 70 seconds for each and every tests from tests/avocado/boot_linux.py to get canceled."_

Implement an improved version of this method.

Maybe something like:

```python
with socket() as s: 
    s.bind(('',0)) 
    return s.getsockname()[1]
```  

**Additional information**
https://www.mail-archive.com/qemu-devel@nongnu.org/msg871809.html

beraldoleal:
Currently we are assuming that UDP and TCP ports shares the the same domain,
which is not true. The same is valid for IPv4 vs IPv6. When asking for
free ports, makes sense to be specific about what kind of port we are
looking for. Because of that, going over all the possible domains to get
a free port is odd. Closes #5273

### DIFF
--- a/avocado/utils/network/ports.py
+++ b/avocado/utils/network/ports.py
@@ -18,6 +18,7 @@ Module with network related utility functions
 
 import random
 import socket
+import warnings
 
 from avocado.utils.data_structures import Borg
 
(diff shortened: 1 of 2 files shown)


clebergnu (commented):
@beraldoleal the changes look good so far. I'll be doing a bit more testing, and looking into if/how we can add a bit coverage there, so please bear with me a little longer.

PR #5301, opened 2022-03-29
Legacy runner warning

richtja:
Because we started removing legacy code and legacy tests. We can't
guarantee the stability of legacy runner on the master branch, but the
`--test-runner=runner`  option wasn't removed yet and the users still
can use legacy runner. Let's add a warning message that using legacy
runner on the newest avocado might be problematic and soon will be
removed.


beraldoleal (approved):
LGTM, thanks.

Issue #5305, opened 2022-04-01
'core.show' option, not working properly with Job API


richtja:
**Describe the bug**
The 'core.show' option is not respected in Job config when it is used in Job API. This problem only appears when  'core.show' option is only in suite config, not in Job config. Reported by @marioaag

**Steps to reproduce**
test file:
```
import logging

from avocado import Test


class PassTest(Test):


    def test(self):
        inject_number = 1
        progress_log = logging.getLogger("avocado.test.progress")
        progress_log.info(f'{self.name}: Running Inject {inject_number}')
```
Job API file:
```
#!/usr/bin/env python3

import sys

from avocado.core.job import Job
from avocado.core.suite import TestSuite

job_config = {'core.show': ['app']}

config = {'resolver.references': ['/tmp/test.py:PassTest.test'], 'core.show': ['app', 'avocado.test.progress']}

suite = TestSuite.from_config(config)
with Job(job_config, [suite]) as j:
    sys.exit(j.run())
```

**Expected behavior**
```
$ python job_api.py 
JOB ID     : cb8e56e45ba00bc03af921866a460ea5f6287068
JOB LOG    : /home/jarichte/avocado/job-results/job-2022-04-01T09.48-cb8e56e/job.log
 (14216f05-b9b8-4fa4-a78e-e4260e2d51ea-1/1) /tmp/test.py:PassTest.test: STARTED
14216f05-b9b8-4fa4-a78e-e4260e2d51ea-1-/tmp/test.py:PassTest.test: avocado.test.progress: 1-/tmp/test.py:PassTest.test: Running Inject 1
 (14216f05-b9b8-4fa4-a78e-e4260e2d51ea-1/1) /tmp/test.py:PassTest.test: PASS (0.01 s)
RESULTS    : PASS 1 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB HTML   : /home/jarichte/avocado/job-results/job-2022-04-01T09.48-cb8e56e/results.html
JOB TIME   : 1.77 s
```

**Current behavior**
```
$ python job_api.py 
JOB ID     : 2dbb4f470ff463b8540d46dcd571cfb9b1fe1cc8
JOB LOG    : /home/jarichte/avocado/job-results/job-2022-04-01T09.50-2dbb4f4/job.log
[...]

PR #5306, opened 2022-04-01
job config rewrites the suite config in job API

richtja:
When you have the same config option in suite config and the job config,
the job config will be used, but it should be the suite option because
it is lower level configuration. This will fix that.

Reference: #5305


ana (approved):

beraldoleal:
Hi @richtja, please do not merge this yet. I'm revisiting the issue to double-check the approach.

beraldoleal (approved):
Yah, this makes sense. Thanks @richtja.

Issue #5347, opened 2022-04-22
Test parameters not being used in Runnable


marioaag:
**Describe the bug**
Parameters provided via run.test_parameters at the Runnable level are not used

**Steps to reproduce**
Generate a Job specifying parameter at the Runnable level, them will not be available at test level (avocado instrumented)

**Expected behavior**
The parameters should be available at test level via self.params.get()

**Current behavior**
The parameters are not available

**System information (please complete the following information):**
 - OS: 
    NAME="Red Hat Enterprise Linux"
    VERSION="8.5 (Ootpa)"
    ID="rhel"
    ID_LIKE="fedora"
    VERSION_ID="8.5"
    PLATFORM_ID="platform:el8"
 - Avocado version: Avocado 96.0
 - Avocado installation method: pip

**Additional information**
This is something that use to work in Avocado 92.0 was reported in the pass in #4569

richtja:
IMO this is not a bug, because the `config` shouldn't be change on the Runnable level. Maybe we should add some warning to inform the user when the runnable config is different to the suite config. Avocado works only with two levels of configuration, the Job config and the Suite config. The runnable should have the same config as suite and after the #5326 it should have only configuration which is necessary for runner.

marioaag:
Hi @clebergnu, thanks for reaching out!
Yes that is in resume what I'm interested to be hable to do but at API level. I have a bunch of different test methods and each one of them requires different parameters, in some cases is the same test method executed with different parameters, in some way it is similar to what variant should be doing but in my case is really useful to be able to control the parameters provided to each one of the test methods, I don't want to run a variant of one test method that doesn't even use that param.

richtja:
Hi @marioaag can you please check if #5601 helps with your problem, thanks. With #5601 you should be able to use Job API to set parameter into runnable, see [sleepjob_parameters example](https://github.com/avocado-framework/avocado/blob/839251dc892d3d369df3a5ed09552532c49b900e/examples/jobs/sleepjob_parameters.py). I hope that it what you have been looking for.

marioaag:
Thanks so much for the update!!!
Looks like the concept is there! Any estimate of when it is going to get released to give it a try?

richtja:
Hi @marioaag, we plan to have the next release on Mar 6th, 2023. But you can try the latest changes now, with:
``` 
pip install https://github.com/avocado-framework/avocado/archive/refs/heads/master.zip
```

richtja:
[...]

PR #5348, opened 2022-04-22
Adding run.test_params to test params

marioaag:
Adding logic to use the parameters defined in run.test_params as part of the Runnable params.

Bug related: #5347

Signed-off-by: Mario Alberto Alvarado Galindo <Mario.Alberto.Galindo@ibm.com>


richtja (changes requested):
Hi @marioaag, thank you for your work, but I think this will need another approach. We removed this code in 8e768567e8dcc053b8f691bdc05862b2ea60a05e when we united the test parameters and variants. IMO if we have some problem with  `run.test_params` it should be solved in avocado core not in runner itself, because the parameters are also used by other runners like `exec-runner`. If you will need any help, don't hesitate to ask.

richtja:
Hi @marioaag gentle ping here. Do you have any status to this? thanks.

marioaag:
The issue planned to be fix here was already solved in the new releases.

PR #5354, opened 2022-04-26
Revert "Function checks for packet loss during ping execution"

beraldoleal:
This method is working properly with LocalHost and RemoteHost. Please
check my comments on PR #5349.

This reverts commit 88673e242571ece6456b706ce3f2c8962e720983.


clebergnu (commented):
I agree with the revert on the grounds of the comments from @beraldoleal [here](https://github.com/avocado-framework/avocado/pull/5349/files#r859144848).

I also have the impression that this is `ping_check()` could be used for the same purposes (by checking for a `ping_check()` failure, you can't you infer packet loss?).

But, let's hear from @TasmiyaNalatwad, @abdhaleegit and @PraveenPenguin .

abdhaleegit:
> I agree with the revert on the grounds of the comments from @beraldoleal [here](https://github.com/avocado-framework/avocado/pull/5349/files#r859144848).
> 
> I also have the impression that this is `ping_check()` could be used for the same purposes (by checking for a `ping_check()` failure, you can't you infer packet loss?).
> 
> But, let's hear from @TasmiyaNalatwad, @abdhaleegit and @PraveenPenguin .

Thanks for your suggestions

Initial thought was to keep same ping_check(), in our case most of our NIC adapters was passing the ping_check though there were some packet loss for immediate ping_check just after configuring ip for the adapter and when we retry ping after some delay I see there is no packet loss.. so we added a sleep(6) in our test code, but than we thought why not write a function for packet loss check instead of ping_check() as it is already being used in many NIC tests.

But yes after seeing your suggestion we can extend same ping_check to infer packet loss

```
       try:
            run_command(cmd, self.host)
            <<add code to check for packet">>
            return False
        except Exception as ex:
            raise NWException(f"Failed to ping: {ex}")
        return True
```
@TasmiyaNalatwad  see if you can retry the ping command inside ping_check() for given timeout=30, instead of wait_for in test code

abdhaleegit:
@beraldoleal @clebergnu just wondering what if ping command exited with 0 but with 0.016% of packet loss, should we still consider ping_check fail ?

PraveenPenguin:
> @beraldoleal @clebergnu just wondering what if ping command exited with 0 but with 0.016% of packet loss, should we still consider ping_check fail ?

yes I agree with @abdhaleegit as the `ping` utility just checks ping, we need to have this check we should not packet drop

beraldoleal:
> I agree with the revert on the grounds of the comments from @beraldoleal [here](https://github.com/avocado-framework/avocado/pull/5349/files#r859144848).
> 
> I also have the impression that this is `ping_check()` could be used for the same purposes (by checking for a `ping_check()` failure, you can't you infer packet loss?).

I would split the methods. Ping is just a ping with checking the return code 0 or not 0. And packet loss a different method. My main concern right now is the support for both LocalHost and RemoteHost. This method as it is is not supporting them.

clebergnu (approved):
OK, so even if there's need for another utility, this one is breaking the general assumption of module (being able to run either local or remote).

Please @TasmiyaNalatwad if you would be so kind to send another version with that in mind.

BTW, we have a feature freeze on April 29, 2022, so please consider sending a new version ASAP.

Thanks!

TasmiyaNalatwad:
@beraldoleal  @clebergnu Please find the PR with the reviews addressed https://github.com/avocado-framework/avocado/pull/5361

Issue #5357, opened 2022-04-27
failfast option should be available for any "NOT OK" status


richtja:
**Is your feature request related to a problem? Please describe.**
Based on our [documentation](https://avocado-framework.readthedocs.io/en/latest/guides/user/chapters/introduction.html?highlight=failfast#interrupting-the-job-on-first-fail-failfast), the --failfast option interrupts the job on first fail. And this is working as expected. But IMO this shouldn't be limited only to failure, but to any "NOT OK" status `[ERROR, FAIL, INTERRUPTED]`. Because when you don't want to run next tests after failure, there is no difference between fail and others.
This example shows the current behavior.
```
$ avocado run --nrunner-max-parallel-tasks=1 --failfast ../avocado/examples/tests/errortest.py /bin/true /bin/false ../avocado/examples/tests/passtest.py 
JOB ID     : 868593a1d9855d5ceaaa83042df0042d2f96df90
JOB LOG    : /home/jarichte/avocado/job-results/job-2022-04-07T11.40-868593a/job.log
 (1/4) ../avocado/examples/tests/errortest.py:ErrorTest.test: STARTED
 (1/4) ../avocado/examples/tests/errortest.py:ErrorTest.test: ERROR: This should end with ERROR. (0.04 s)
 (2/4) /bin/true: STARTED
 (2/4) /bin/true: PASS (0.01 s)
 (3/4) /bin/false: STARTED
 (3/4) /bin/false: FAIL (0.01 s)
Interrupting job (failfast).
RESULTS    : PASS 1 | ERROR 1 | FAIL 1 | SKIP 1 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB HTML   : /home/jarichte/avocado/job-results/job-2022-04-07T11.40-868593a/results.html
JOB TIME   : 2.45 s

Test summary:
../avocado/examples/tests/errortest.py:ErrorTest.test: ERROR
/bin/false: FAIL
```

**Describe the solution you'd like**
```
$ avocado run --nrunner-max-parallel-tasks=1 --failfast ../avocado/examples/tests/errortest.py /bin/true /bin/false ../avocado/examples/tests/passtest.py 
JOB ID     : 868593a1d9855d5ceaaa83042df0042d2f96df90
JOB LOG    : /home/jarichte/avocado/job-results/job-2022-04-07T11.40-868593a/job.log
 (1/4) ../avocado/examples/tests/errortest.py:ErrorTest.test: STARTED
[...]

PR #5358, opened 2022-04-27
failfast for any "NOT OK" status

richtja:
This PR will add `failfast` option for any "NOT OK" status. Until
now, the `failfast` was working only for `fail` status. This change adds
the same ability to `error` and `interrupted` statuses as well.

Reference: #5357

### DIFF
--- a/avocado/core/task/statemachine.py
+++ b/avocado/core/task/statemachine.py
@@ -5,6 +5,7 @@ import multiprocessing
 import time
 
 from avocado.core.exceptions import TestFailFast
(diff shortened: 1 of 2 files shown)


mcondotta:
Looks good to me.
It will definitely benefit the CI tests that fail with ERROR or TIMEOUT and are pre-requisites for the next tests in the list.
Our CI which uses Avocado and Avocado-VT to test virtualization stack (KVM, QEMU and Libvirt) will benefit a lot from this as it will cut the time that the pipeline jobs take to fail and finish when there are problems.

PR #5372, opened 2022-05-05
Messages timestamp changes

richtja:
This change the message timestamp value from time when the message was
created to time when the message was published. This will fix the issue
where the publishing order of messages was different from order by time
and the status repo discard messages in the wrong order.

Reference: #5327

### DIFF
--- a/avocado/core/nrunner/app.py
+++ b/avocado/core/nrunner/app.py
@@ -4,6 +4,7 @@ import json
 import os
 import re
 import sys
+import time
 
 import pkg_resources
 
@@ -105,7 +106,7 @@ class BaseRunnerApp:
     )
 
     def __init__(self, echo=print, prog=None, description=None):
-        self.echo = echo
+        self._echo = echo
         self.parser = None
         if prog is None:
             prog = self.PROG_NAME
@@ -113,6 +114,10 @@ class BaseRunnerApp:
             description = self.PROG_DESCRIPTION
         self._class_commands_method = self._get_commands_method()
         self._setup_parser(prog, description)
+
+    def echo(self, message):
+        message['time'] = time.monotonic()
+        self._echo(message)
 
     def _setup_parser(self, prog, description):
         self.parser = argparse.ArgumentParser(prog=prog,
@@ -220,7 +225,7 @@ class BaseRunnerApp:
         of a given kind, or identifying if a runner script has a given feature
         (as implemented by a command).
         """
-        self.echo(json.dumps(self.get_capabilities()))
+        self._echo(json.dumps(self.get_capabilities()))
 
     def command_runnable_run(self, args):
         """
--- a/avocado/core/nrunner/runner.py
+++ b/avocado/core/nrunner/runner.py
@@ -1,5 +1,3 @@
-import time
-
 from avocado.core.nrunner.runnable import \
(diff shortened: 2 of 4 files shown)


richtja:
Hi @luckyh can you please check, if this will fix your problem with discarded messages. Thanks.

avocado/core/nrunner/app.py
```diff
@@ -114,6 +115,10 @@ def __init__(self, echo=print, prog=None, description=None):
         self._class_commands_method = self._get_commands_method()
         self._setup_parser(prog, description)
 
+    def echo(self, message):
+        message['time'] = time.monotonic()
+        self._echo(message)
+
```
beraldoleal, line 121: Usually it is "strange" to have a method with the same name as an argument. This leads to confusing when using the API:

```python
r = Runner(echo=print)
r.echo # this will be completely different from what passed by the user.
```

What about renaming `s/echo/send_message/`? Or rename the `echo` argument to `echo_method`.

Also, if an argument is being used it is ok to leave as public, since it was passed during the initialization.

beraldoleal (changes requested):
Hi @richtja thanks for this. LGTM. Just one small comment about the OOP API usage.

luckyh:
> Hi @luckyh can you please check, if this will fix your problem with discarded messages. Thanks.

Hi @richtja thanks for the fix, right now I'm going to do the verification and will update the result later, thanks!

richtja:
After the discussion, I am moving this to Draft, because there is another solution in #5374

luckyh:
> After the discussion, I am moving this to Draft, because there is another solution in #5374

Okay, just see this comment, anyway, I have finished the first round of testing and didn't meet this issue, thanks!
```
# git diff
diff --git a/avocado/core/status/repo.py b/avocado/core/status/repo.py
index b9f1a38f..cfb9f7cd 100644
--- a/avocado/core/status/repo.py
+++ b/avocado/core/status/repo.py
@@ -110,6 +110,8 @@ class StatusRepo:
                 index = len(self.get_all_task_data(task_id))
                 self._status_journal_summary.append((task_id, status, time,
                                                      index))
+            else:
+                LOG.debug("Discard (%s vs %s): %s", time, current_time, message)
             if time > current_time:

# grep -c Discard workspace/job-results/latest/job.log 
0
```

richtja:
> > After the discussion, I am moving this to Draft, because there is another solution in #5374
> 
> Okay, just see this comment, anyway, I have finished the first round of testing and didn't meet this issue, thanks!
> 
> ```
> # git diff
> diff --git a/avocado/core/status/repo.py b/avocado/core/status/repo.py
> index b9f1a38f..cfb9f7cd 100644
> --- a/avocado/core/status/repo.py
> +++ b/avocado/core/status/repo.py
> @@ -110,6 +110,8 @@ class StatusRepo:
>                  index = len(self.get_all_task_data(task_id))
>                  self._status_journal_summary.append((task_id, status, time,
>                                                       index))
> +            else:
> +                LOG.debug("Discard (%s vs %s): %s", time, current_time, message)
>              if time > current_time:
>                  self._status[task_id] = (status, time)
> 
> # grep -c Discard workspace/job-results/latest/job.log 
> 0
> ```

Thanks, it is great to know that we are on the right path to solve this problem.

PR #5374, opened 2022-05-10
Discarding older messages from status repo fix

richtja:
The status repo was discarding messages when the message timestamp was
lower than the latest save message. This change removes timestamp
checking and ensures the status changes only in expected life-cycle of a
test: started->running->finished. This means that change from running to
finished is possible, but from finished to running is not.

This change also stores every message into the `status_journal_summary`
list and sorts them by the timestamp. This sort is not optimal because,
the `runner_nrunner` plugin reads data from the status journal during
runner to server communication. This means that status repo sorts
messages which haven't been read yet by `runner_nrunner` plugin, but
can't ensure the order of the whole communication.

Reference: #5327

### DIFF
--- a/avocado/core/status/repo.py
+++ b/avocado/core/status/repo.py
@@ -1,3 +1,4 @@
+import heapq
 import logging
 
 from avocado.core.status.utils import json_loads
@@ -26,8 +27,7 @@ class StatusRepo:
         #: Contains the most up to date status of a task, and the time
         #: it was set in a tuple (status, time).  This is keyed
         #: by the task ID, and the most up to date status is determined by
-        #: the "timestamp" in the "time" field of the message, that is,
-        #: it's *not* based by the order it was received.
+        #: the status type of message.
         self._status = {}
         #: Contains a global journal of status updates to be picked, each
         #: entry containing a tuple with (task_id, status, time).  It discards
@@ -94,8 +94,11 @@ class StatusRepo:
             return None
         return task_data[-1]
 
+    def status_journal_summary_pop(self):
+        return heapq.heappop(self._status_journal_summary)
+
     def _update_status(self, message):
-        """Update the latest status of a task (by time, not by message)."""
+        """Update the latest status of a task (by message)."""
         task_id = message.get('id')
         status = message.get('status')
         time = message.get('time')
@@ -103,15 +106,25 @@ class StatusRepo:
             return
         if task_id not in self._status:
             self._status[task_id] = (status, time)
-            self._status_journal_summary.append((task_id, status, time, 0))
+            heapq.heappush(self._status_journal_summary, (time,
+                                                          task_id,
+                                                          status,
+                                                          0))
         else:
-            _, current_time = self._status[task_id]
-            if time >= current_time:
-                index = len(self.get_all_task_data(task_id))
-                self._status_journal_summary.append((task_id, status, time,
-                                                     index))
-            if time > current_time:
+            current_status, _ = self._status[task_id]
+            if current_status == "finished":
+                LOG.warning('Received a %s message after finished message: %s',
+                            status, message)
+            elif status == "started":
+                LOG.warning('Received a started message when the status '
+                            'is already %s: %s', current_status, message)
+            else:
                 self._status[task_id] = (status, time)
+            index = len(self.get_all_task_data(task_id))
+            heapq.heappush(self._status_journal_summary, (time,
+                                                          task_id,
+                                                          status,
+                                                          index))
 
     def process_message(self, message):
         for required_field in ('id', 'job_id'):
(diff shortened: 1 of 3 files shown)


beraldoleal (approved):
I didn't tested this, but LGTM.

clebergnu (approved):
Looks and works great AFAICT, thanks!

pevogam:
> Looks and works great AFAICT, thanks!

Not quite though, checking out the most recent avocado master branch fails with all tests skipped for us and bisecting brought me here:
```
JOB ID     : 840db3b9d64388fd12c1966a222c105fafe3c9f7
JOB LOG    : /mnt/local/results/job-2022-06-01T10.54-840db3b/job.log
 (1-mis1/5) i2nsys.normal.nongui.quicktest.rpmlist.mis1.qemu_kvm_business_server.default_bios.no_9p_export.smallpages.no_pci_assignable.qcow2.virtio_blk.smp2.virtio_net.Linux.Intra2net.Business_Server.official: STARTED
 (1-mis1/5) i2nsys.normal.nongui.quicktest.rpmlist.mis1.qemu_kvm_business_server.default_bios.no_9p_export.smallpages.no_pci_assignable.qcow2.virtio_blk.smp2.virtio_net.Linux.Intra2net.Business_Server.official: PASS (23.37 s)
RESULTS    : PASS 1 | ERROR 0 | FAIL 0 | SKIP 4 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB HTML   : /mnt/local/results/job-2022-06-01T10.54-840db3b/results.html
JOB TIME   : 41.70 s
```
@richtja: Does Avocado VT need some adaptation too in order to make this work?

richtja:
Hi @pevogam, there shouldn't be any adaptation on Avocado VT necessary. In our avocado-vt jobs, we didn't register any problems. But because you are using your avocado runner, maybe for you might be relevant, changes to the `avocado/plugins/runner_nrunner.py` in [here](https://github.com/avocado-framework/avocado/pull/5374/files#diff-5f176e989399c44696c35a67b527b008dfeb2725c5755f360c33d95d6aa6bf4e).

pevogam:
> Hi @pevogam, there shouldn't be any adaptation on Avocado VT necessary. In our avocado-vt jobs, we didn't register any problems. But because you are using your avocado runner, maybe for you might be relevant, changes to the `avocado/plugins/runner_nrunner.py` in [here](https://github.com/avocado-framework/avocado/pull/5374/files#diff-5f176e989399c44696c35a67b527b008dfeb2725c5755f360c33d95d6aa6bf4e).

This was indeed relevant and fixed the issue for me. However, since this is a public API change in the status repo I was wondering: is there any guarantee on API stability and backwards compatibility at least for the public API-s?

richtja:
> > Hi @pevogam, there shouldn't be any adaptation on Avocado VT necessary. In our avocado-vt jobs, we didn't register any problems. But because you are using your avocado runner, maybe for you might be relevant, changes to the `avocado/plugins/runner_nrunner.py` in [here](https://github.com/avocado-framework/avocado/pull/5374/files#diff-5f176e989399c44696c35a67b527b008dfeb2725c5755f360c33d95d6aa6bf4e).
> 
> This was indeed relevant and fixed the issue for me. However, since this is a public API change in the status repo I was wondering: is there any guarantee on API stability and backwards compatibility at least for the public API-s?

True is that we don't testing the backwards compatibility for the public API and the nrunner API is still going through the big development with a lot of changes. So IMO we are not able to guarantee backwards compatibility of public API. It would be great to have that, but IMO it would create a lot of troubles, and it would make a development of new nrunner features a lot harder.

pevogam:
> > > Hi @pevogam, there shouldn't be any adaptation on Avocado VT necessary. In our avocado-vt jobs, we didn't register any problems. But because you are using your avocado runner, maybe for you might be relevant, changes to the `avocado/plugins/runner_nrunner.py` in [here](https://github.com/avocado-framework/avocado/pull/5374/files#diff-5f176e989399c44696c35a67b527b008dfeb2725c5755f360c33d95d6aa6bf4e).
> > 
> > 
> > This was indeed relevant and fixed the issue for me. However, since this is a public API change in the status repo I was wondering: is there any guarantee on API stability and backwards compatibility at least for the public API-s?
> 
> True is that we don't testing the backwards compatibility for the public API and the nrunner API is still going through the big development with a lot of changes. So IMO we are not able to guarantee backwards compatibility of public API. It would be great to have that, but IMO it would create a lot of troubles, and it would make a development of new nrunner features a lot harder.

It depends how you see "troubles" because you mostly refer to "troubles on the side of the maintainer" which is always in a tradeoff with "troubles on the side of the user". So while it seems the maintainers have fewer troubles by not having to provide a stable API to the users, the costs are simply outsourced to the users that constantly have to update their own scripts and fix breakage. In any case I still hope that this big nrunner refatoring is just a multi-year-long stage that will eventually end so that the user base can also increase and feel safe enough to update.

[*] Note that not updating and keeping up is not exactly an option since the errors and breakage from changes only gets compounded and harder to debug and fix the more it accumulates.

@clebergnu Just mentioning you here since it would be great if you are also aware of this.

PR #5409, opened 2022-06-23
Fix connection-refused race condition when spawning tasks

pevogam:
Signed-off-by: Plamen Dimitrov <plamen.dimitrov@intra2net.com>

### DIFF
--- a/avocado/core/nrunner/task.py
+++ b/avocado/core/nrunner/task.py
@@ -1,11 +1,15 @@
 import base64
 import json
+import logging
 import socket
 import tempfile
+import time
 from uuid import uuid1
 
 from avocado.core.nrunner.runnable import (
     RUNNERS_REGISTRY_STANDALONE_EXECUTABLE, Runnable)
+
+LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


beraldoleal (approved):
I tried using timeout argument for the create_connection() but it will not solve the problem, so I'm ok with that.

pevogam:
> I tried using timeout argument for the create_connection() but it will not solve the problem, so I'm ok with that. There is only one GH action failing.

Yes, the failing action is about logging which I commented to ask for your opinion first. Should we at least log the error? Or just a `pass` is enough in your opinion?

beraldoleal:
> > I tried using timeout argument for the create_connection() but it will not solve the problem, so I'm ok with that. There is only one GH action failing.
> 
> Yes, the failing action is about logging which I commented to ask for your opinion first. Should we at least log the error? Or just a `pass` is enough in your opinion?

I think we could log there at least with a LOG.debug(), or LOG.warning()

pevogam:
Alright, adapted accordingly to your suggestion.

Issue #5375, opened 2022-05-10
Issuing log.warning prior to a failed assertion results in test result as WARN instead of FAIL


on-kage:
**Describe the bug**
If a test logs a warning via the log.warning command, then no matter what happens afterwards the test result will be WARN. This is the case even if there is a later failed assertion

**Steps to reproduce**
from avocado import Test

class Test_assert(Test):
        """
        :avocado: enable
        """

        def setUp(self):
            pass

        def tearDown(self):
            pass

        def test_1(self):
            self.log.warning(f'warn message')
            self.assertTrue(False, 'assertion fail')

**Expected behavior**
In the test shown in "Steps to reproduce", I would expect the test result to be FAIL based on the failed assertTrue statement. 

**Current behavior**
The test results in a WARN

**System information (please complete the following information):**
 - OS: docker image based on `python:3.10.2-slim-buster`
 - Avocado version: Avocado 96.0
 - Avocado installation method: pip

**Additional information**
Checked with Avocado 69 which is what we were previously using and the result of the test is FAIL

richtja:
Hi @on-kage thanks for reporting this. It is definitely a problem.

PR #5425, opened 2022-07-11
Test results with warning fix

richtja:
If a test logged a warning via the log.warning command, then no matter
what happened afterwards, the test result was WARN. This fixes this
problem. So the test result will be WARN only when the test itself
succeed with warning.

Reference: #5375


clebergnu (approved):
LGTM, thanks!

beraldoleal (approved):

PR #5424, opened 2022-07-11
fix data_factory.py

wangp-h:
header: fix function make_dir_and_populate
message: When use funciton make_dir_and_populate to create a directory in basedir and populate with a number of files,
raise TypeError: a bytes-like object is required, not 'str'
This PR fix this bug
Reference: See the following drawing
企业微信截图_16575253716941
Signed-off-by:WangPeng wangpengb@uniontech.com

PR #5430, opened 2022-07-14
fix byte use func split

yangchenguang94:
Signed-off-by:Yangchenguang yangchenguang@uniontech.com

image


richtja (changes requested):
Hi @yangchenguang94  thank you for participating in Avocado project, the code LGTM, but can you please add the commit message and update your signature, based on our [contributor guide](https://avocado-framework.readthedocs.io/en/latest/guides/contributor/chapters/styleguides.html#commit-style-guide), so the Static checks can pass. You can also have a look on #5424 where I discussed the problems with Static checks. Thanks

PR #5451, opened 2022-07-29
Added one definition get_pci_info() and update one definition get_cfg()

maramsmurthy:
Added definition get_pci_info() to capture slot information using lspci command

Updated get_cfg() to capture Pysical slot, Sub vendor and Device ID from description

Signed-off-by: Maram Srimannarayana Murthy <Maram.Srimannarayana.Murthy@ibm.com>

### DIFF
--- a/avocado/utils/pci.py
+++ b/avocado/utils/pci.py
@@ -280,6 +280,28 @@ def get_pci_id(pci_address):
         return ":".join(pci_id)
 
 
+def get_pci_info(pci_address):
+    """
+    Gets PCI info of given PCI address.
+
+    :param pci_address: Any segment of a PCI address (1f, 0000:00:1f, ...)
+
+    :return: Dictionary attribute name as key and attribute value as value.
+    :rtype: Dict
+    """
+    cmd = f"lspci -Dnvmm -s {pci_address}"
+    output = process.run(cmd, ignore_status=True, shell=True).stdout_text
+    pci_info = {}
+    if not output:
+        return
+    for line in output.splitlines():
+        if line.strip():
+            pci_info[line.split(":")[0].strip()] = (
+                    (":".join(line.split(":")[1:])).strip()
+                    )
+    return pci_info
+
+
 def get_driver(pci_address):
     """
     Gets the kernel driver in use of given PCI address. (first match only)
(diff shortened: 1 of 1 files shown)


avocado/utils/pci.py
```diff
@@ -280,6 +280,24 @@ def get_pci_id(pci_address):
         return ":".join(pci_id)
 
 
+def get_pci_info(pci_address):
+    """
+    Gets PCI info of given PCI address.
+    :param pci_address: Any segment of a PCI address (1f, 0000:00:1f, ...)
+    :return: Dictionary attribute name as key and attribute value as value.
+    :rtype: Dict
+    """
```
Naresh-ibm, line 289: @maramsmurthy maintain proper line of space in docstring.

avocado/utils/pci.py
```diff
@@ -280,6 +280,24 @@ def get_pci_id(pci_address):
...
+    cmd = f"lspci -Dnvmm -s {pci_address}"
+    output = process.run(cmd, ignore_status=True, shell=True).stdout_text
+    pci_info = {}
+    if output:
+        for line in output.splitlines():
+            if line.strip():
+                s_line = line.split(":")
```
Naresh-ibm, line 296: I think above two line are not required and can be incorporate into below single line.
  ```
if output:
        for line in output.splitlines().strip():
            pci_info[s_line.split(":")[0].strip()] = (":".join(s_line.split(":")[1:])).strip()
or
if output:
        for line in output.splitlines().strip():
            line = line.split(":")
            pci_info[line[0].strip()] = (":".join(line[1:])).strip()
           
```

if condition is not required and try avoid extra variable initialization(s_line).

**Review by Naresh-ibm — CHANGES_REQUESTED** [review]

File: avocado/utils/pci.py (hunk 1 above)
Comment (maramsmurthy) [reply avocado/utils/pci.py:289]: Sure Naresh! will add line spaces as suggested

File: avocado/utils/pci.py (hunk 2 above)
Comment (maramsmurthy) [reply avocado/utils/pci.py:296]: Yes! I will remove s_line attribute which is not needed.

.strip() function is not supported on array. That the reason I wrote a different if condition. strip() function is available in numpy module. Let me know if we can use numpy module?

File: avocado/utils/pci.py [hunk 3]
```diff
@@ -280,6 +280,24 @@ def get_pci_id(pci_address):
...
+    """
+    Gets PCI info of given PCI address.
+    :param pci_address: Any segment of a PCI address (1f, 0000:00:1f, ...)
+    :return: Dictionary attribute name as key and attribute value as value.
+    :rtype: Dict
+    """
+    cmd = f"lspci -Dnvmm -s {pci_address}"
+    output = process.run(cmd, ignore_status=True, shell=True).stdout_text
+    pci_info = {}
+    if output:
```
beraldoleal, line 293: To reduce the indentation and logic, please try to invert the order:

```python
if not output:
   return

```

beraldoleal (changes requested):

avocado/utils/pci.py
maramsmurthy: Addressed suugested changes

PraveenPenguin (approved):

PR #5457, opened 2022-08-11
Fixed the volume group creation error

PavamanSubramaniyam:
The volume group name and logical volume device parameters passed for vgcreate command were not properly handled
I have fixed the creation of the command to be executed with correct conditional statements

Signed-off-by: Pavaman Subramaniyam <pavsubra@linux.vnet.ibm.com>


PavamanSubramaniyam:
As discussed with @PraveenPenguin,  that we should be sending the fixes or code changes to the files in avocado/utils/ folder to the upstream branch and not to the tagged releases of avocado.
So, I have made the code changes for the upstream branch accordingly and send in another PR for merger.

Issue #5453, opened 2022-08-04
Requirements cache interface


richtja:
**Is your feature request related to a problem? Please describe.**
Right now, the requirements cache is hidden inside avocado test process and user doesn't have any control over it. The main problem with this is, that if the data in cache doesn't correspond with the real state in the environment, the avocado won't work properly. In such situation, the only way how the user can fix avocado is to remove the requirement cache file, which is not very user-friendly approach.

**Describe the solution you'd like**
Avocado provides vmimage cache interface. With this interface, user can list the cache and make updates. I would like to have similar interface for requirements cache. With this interface, the user will be able to list all the requirements stored in cache and remove those which doesn't respect the current environment.

**Describe alternatives you've considered**
We can only create small script for removing the whole requirement cache. It will fix the problem with requirements cache, but it won't be so effective in situations when you have hundreds of entries, and you need to remove only one.

PR #5475, opened 2022-08-18
Cache interface

richtja:
This is an initial PR which crates avocado cache interface. It introduces new plugin "Cache" which will be responsible for manipulating with cache, and it's representation through avocado interface. This PR also adds vmimage cache and requirements cache to this interface.

For now, it supports only listing of cache entries, but it is possible to expand for features like prune or edit.

Reference: https://github.com/avocado-framework/avocado/issues/5453

### DIFF
--- a/avocado/core/dependencies/requirements/cache/backends/sqlite.py
+++ b/avocado/core/dependencies/requirements/cache/backends/sqlite.py
@@ -333,3 +333,35 @@ def get_all_environments_with_requirement(
             else:
                 requirements[row[0]] = [(row[1], row[2])]
     return requirements
+
+
+def get_all_requirements():
+    """Fetches all requirements from database.
+
+    :return: Dict with all environments which has requirements.
+
+    """
+    requirements = {}
+    if not os.path.exists(CACHE_DATABASE_PATH):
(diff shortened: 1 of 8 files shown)


clebergnu (changes requested):
Hi @richtja, this looks good, thanks!

Besides the typo, I think the CLI syntax/interface is not prepared for the future extensibility you described.  The current functionality is all about a `list` sub-command, and as such, should require that explicitly.

Then, the `CACHE_TYPE` parameter should also become a non-positional argument IMO: it's more explicit and allows for better extensibility in the future.

richtja:
> Hi @richtja, this looks good, thanks!
> 
> Besides the typo, I think the CLI syntax/interface is not prepared for the future extensibility you described. The current functionality is all about a `list` sub-command, and as such, should require that explicitly.
> 
> Then, the `CACHE_TYPE` parameter should also become a non-positional argument IMO: it's more explicit and allows for better extensibility in the future.

@clebergnu thanks for your review I did fix via force-push, please have a look.

clebergnu (approved):
LGTM, thanks!

PR #5500, opened 2022-10-27
Cache clear

richtja:
This adds a clear feature to cache interface. With this feature, users
will be able to remove entries from cache base on their cache type.

Reference: https://github.com/avocado-framework/avocado/issues/5453

### DIFF
--- a/avocado/core/plugin_interfaces.py
+++ b/avocado/core/plugin_interfaces.py
@@ -465,3 +465,10 @@ class Cache(Plugin):
         :return: Human-readable list of entries which will be printed to console.
         :rtype: str
         """
+
+    @abc.abstractmethod
+    def clear(self):
+        """Clear all cache entries
+
+        It removes every cache entire.
+        """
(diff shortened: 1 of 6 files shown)


richtja:
> Hi @richtja , I've caught a few typos and documentation suggestions.
> 
> I particularly disliked the current behavior because, from an innocent "avocado cache clear" execution, without specifying the type, everything was removed, including the somehow costly to download VM Images.
> 
> I understand that the UNIX philosophy is to proceed with what was requested (including a recursive removal of files). Still, I think we should reconsider it here for a global clean operation.
> 
> My suggestion is to start with a required parameter for the cache type to be cleaned. This would make sure that users won't have other types of cache cleaned due to an innocent `avocado cache clean`. Keep in mind that because of the plugin interface, a new plugin can be added between versions, and consequently, the current behavior will clean more than the user previously expected.
> 
> Also, I haven't tested what happens if the `requirements.sqlite` is cleaned while we have a running Avocado job working on it. We should make sure this type of concurrent access works as expected.

Hi @clebergnu, thanks for your review. I see your point and I agree with you that this behavior might cause an innocent cache clear, but I don't want to remove the global cache removal, because I see this feature useful. What about adding the confirmation mechanism before the actual removal. Something like:

```
These dependencies will be cleared from cache: 

requirement:
        local:
                Environment Requirement_type Requirement
                jarichte    package          podman

Do you want to continue (y/n):
```

richtja:
Hi @clebergnu, I create v2 #5536 which focuses only on metadata. Please have a look.

PR #5507, opened 2022-11-02
utils/genio.py: added method for pattern match

rohan-r-deshpande:
- added a method read_line_with_matching_pattern()
- This method takes a file name, searches the file for the occurrences of specified "pattern".
- If found returns the line/lines in the form of a list.

Signed-off-by: Rohan Deshpande <rohan_d@linux.ibm.com>

### DIFF
--- a/avocado/utils/genio.py
+++ b/avocado/utils/genio.py
@@ -10,7 +10,11 @@
 #
 # This code was inspired in the autotest project,
 # client/shared/utils.py
-# Authors: Martin J Bligh <mbligh@google.com>, Andy Whitcroft <apw@shadowen.org>
+# Authors: Martin J Bligh <mbligh@google.com>
+#          Andy Whitcroft <apw@shadowen.org>
+#
+# Modification History:
+# - Added read_line_with_matching_pattern()
 
 """
 Avocado generic IO related functions.
@@ -102,6 +106,32 @@ def read_all_lines(filename):
             contents = [line.rstrip("\n") for line in file_obj.readlines()]
     except Exception:  # pylint: disable=W0703
         pass
+    return contents
+
+
+def read_line_with_matching_pattern(filename, pattern):
+    """
+    Return the line/lines found with a given matching pattern.
+
+    This method returns the line/lines wherever the occurrence of the pattern,
+    that is passed as an input, is found in the file.
+
+    All the occurrences are stored in the list and the list is returned as
+    an output.
+
+    :param filename: Path to the file to be read.
+    :type filename: str
+    :param pattern: List of all the patterns, intended to be searched
+    :type pattern: list
+
+    :return: All those lines from the file, matching the pattern.
+    :rtype: list
+    """
+    contents = []
+    with open(filename, "r", encoding="utf-8") as file_obj:
+        for line in file_obj.readlines():
+            if pattern in line:
+                contents.append(line.rstrip("\n"))
     return contents
 
 


avocado/utils/genio.py
```diff
@@ -10,7 +10,12 @@
 #
 # This code was inspired in the autotest project,
 # client/shared/utils.py
-# Authors: Martin J Bligh <mbligh@google.com>, Andy Whitcroft <apw@shadowen.org>
+# Authors: Martin J Bligh <mbligh@google.com>
+#          Andy Whitcroft <apw@shadowen.org>
+#
+# Modification History:
+# Rohan Deshpande <rohan_d@linux.ibm.com>
+# - Added read_line_with_matching_pattern()
```
PraveenPenguin, line 18: as it is small addition so i feel we can drop that and we can get git history what added

PraveenPenguin (changes requested):

avocado/utils/genio.py
rohan-r-deshpande: Done

abdhaleegit (commented):
@rohan-r-deshpande what if the none of the lines matches the pattern, should we handle that

rohan-r-deshpande:
> @rohan-r-deshpande what if the none of the lines matches the pattern, should we handle that

@abdhaleegit Good point. 

Yes we **should** handle it.

As per the current code flow,  in case of no-match-found for the said patterns, eventually we would be returning an empty list to the caller.  So in my opinion, instead of handling this scenario here, we could might as well incorporate the same from the place where this function is being referenced. 

Reason for this approach?
 - We can neither raise an exception, nor fail it here, for not having the desired pattern found in the file.

Please let me know your opinion!

abdhaleegit:
> > @rohan-r-deshpande what if the none of the lines matches the pattern, should we handle that
> 
> @abdhaleegit Good point.
> 
> Yes we **should** handle it.
> 
> As per the current code flow, in case of no-match-found for the said patterns, eventually we would be returning an empty list to the caller. So in my opinion, instead of handling this scenario here, we could might as well incorporate the same from the place where this function is being referenced.
> 
> Reason for this approach?
> 
> * We can neither raise an exception, nor fail it here, for not having the desired pattern found in the file.
> 
> Please let me know your opinion!

Yes, I agree . it is always good to handle exception at the calling program

abdhaleegit (changes requested):
@rohan-r-deshpande while reading content of a file, better to have it encoded `encoding="utf-8"`  the open call provides this option to encode

rohan-r-deshpande:
> > > @rohan-r-deshpande what if the none of the lines matches the pattern, should we handle that
> > 
> > 
> > @abdhaleegit Good point.
> > Yes we **should** handle it.
> > As per the current code flow, in case of no-match-found for the said patterns, eventually we would be returning an empty list to the caller. So in my opinion, instead of handling this scenario here, we could might as well incorporate the same from the place where this function is being referenced.
> > Reason for this approach?
> > 
> > * We can neither raise an exception, nor fail it here, for not having the desired pattern found in the file.
> > 
> > Please let me know your opinion!
> 
> Yes, I agree . it is always good to handle exception at the calling program

Great! So, without changing anything further here, will handle the exceptions at function calls. Thanks!

rohan-r-deshpande:
> > @rohan-r-deshpande while reading/writing content to a file, better to have it encoded `encoding="utf-8"` the open call provides this option to encode
> 
> Noted, will do the changes.

@abdhaleegit Done and re-submitted the patch. Thanks!

abdhaleegit (approved):
LGTM

PraveenPenguin (approved):

PR #5527, opened 2022-11-25
utils/software_manager/backends/yum.py: make dnf repo options customizable

lkotek:
Method `add_repo` is extended by **optional** argument `opt_params` containing dict with optional repo options.

There are some options already defined in `add_repo` method definition: `gpg_check` and `enabled`. If the same options are passed in `opt_params`, then values in `opt_params` take precedence.

Original usage of method is intact.

_Example of expected usage:_
```
repo_options = {
    "priority": "1",
    "gpgcheck": "0",
    "skip_if_unavailable": "1"
    }

yum_backend = YumBackend()
yum_backend.add_repo("http://some.path.with.dnf.repo", repo_options)
```
### Note
PR  ixes `url` vs. `baseurl` usage preventing method to work.

PR sets `space_around_delimiters=False` for write function of `configparser`, so syntax:
```
skip_if_unavailable = False
````
is replaced by the following one:
```
skip_if_unavailable=False
```
which is used in case of yum `.repo` files.

Signed-off-by: Lukas Kotek <lkotek@redhat.com>

### DIFF
--- a/avocado/utils/software_manager/backends/yum.py
+++ b/avocado/utils/software_manager/backends/yum.py
@@ -111,16 +111,20 @@ class YumBackend(RpmBackend):
         except process.CmdError:
             return False
 
-    def add_repo(self, url):
+    def add_repo(self, url, **opt_params):
         """
         Adds package repository located on [url].
 
         :param url: Universal Resource Locator of the repository.
+        :param opt_params: Dict for optional repo options, eg. {"priority": "1"}.
         """
+        # Set default values for dnf repo options
+        repo_options = {"enabled": "1", "gpgcheck": "0"}
+        repo_options.update(opt_params)
         # Check if we URL is already set
         for section in self.repo_config_parser.sections():
             for option, value in self.repo_config_parser.items(section):
-                if option == "url" and value == url:
+                if option == "baseurl" and value == url:
                     return True
 
         # Didn't find it, let's set it up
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by lkotek** [comment]
Following PRs were tested together:
- https://github.com/autotest/tp-qemu/pull/3550
- https://github.com/avocado-framework/avocado/pull/5527

Repo is created properly (using configuration from expected usage):
```
# cat /etc/yum.repos.d/avocado-managed.repo 
[software_manager_FjPU]
name=Avocado managed repository
baseurl=http://some.path.with.dnf.repo/
enabled=1
gpgcheck=0
priority=1
skip_if_unavailable=1
```

**Comment by yanan-fu** [comment]
@clebergnu Could you help approve the running workflow for @lkotek's first contribute in avocado. Many thanks!

File: avocado/utils/software_manager/backends/yum.py [hunk 1]
```diff
@@ -111,16 +111,20 @@ def remove(self, name):
         except process.CmdError:
             return False
 
-    def add_repo(self, url):
+    def add_repo(self, url, opt_params={}):
```
yanan-fu, line 114: As the CI check error show, it is dangerous to use a dict as the default arguments of a function. 
Suggest using `**kwargs`.

lkotek: That's good point, thanks! 

It is true, we call the `add_repo` method only once (and nobody currently uses `opt_params`), so it should be safe for our use case from https://github.com/autotest/tp-qemu/pull/3550, but this is definitely thing to be avoided. (Fixed in both PRs.)

clebergnu (approved):
Hi @lkotek, these changes LGTM.  I'll wait on @yanan-fu formal ACK before merging it though.

lkotek:
> Hi @lkotek, these changes LGTM. I'll wait on @yanan-fu formal ACK before merging it though.

@clebergnu Thank you a lot! 

Let me also confirm that the latest changes using `**kwargs` also works properly.

yanan-fu (approved):
LGTM

Acked-by: Yanan Fu <yfu@redhat.com>

PR #5561, opened 2022-12-20
Remove triage tag from issue templates

richtja:
Since we moved to the new project board, we don't need tag for triaging issues, because we have special status for that in new project board.

Reference: https://github.com/orgs/avocado-framework/projects/1


clebergnu (approved):
LGTM, thanks!

PR #5575, opened 2023-01-10
Network utilities for all device types and to get interface name via hwaddrs

abdhaleegit:
This patch series adds
+ utilities added to identify different Network device interface 
type like virtual NIC, virther ethernet SEA, SRIOV logical interface etc and 
+  added utility to get interface names via unique hardware address 
+ With some pylint changes

### DIFF
--- a/avocado/utils/network/interfaces.py
+++ b/avocado/utils/network/interfaces.py
@@ -124,8 +124,7 @@ class NetworkInterface:
             mode, slaves = run_command(cmd, self.host).splitlines()
             return {"mode": mode.split(), "slaves": slaves.split()}
         except Exception:
-            raise NWException(
-                f"Slave interface not found for " f"the bond {self.name}")
+            raise NWException(f"Slave interface not found for " f"the bond {self.name}")
 
     def _move_file_to_backup(self, filename, ignore_missing=True):
         destination = f"{filename}.backup"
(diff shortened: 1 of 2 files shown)


avocado/utils/network/interfaces.py
```diff
@@ -612,6 +612,52 @@ def is_bond(self):
...
 
+    def is_veth(self):
+        """Check if interface is a Virtual Ethernet.
+
+        This method checks if the interface is a Virtual Ethernet or not.
+
+        rtype: bool
+        """
+        if not os.path.isfile(f"/sys/class/net/{self.name}/device/devspec"):
+            raise ValueError("Please provide valid inerface name")
```
PraveenPenguin, line 623: handle this gracefully may  raise NWException kind of way

avocado/utils/network/interfaces.py
```diff
@@ -612,6 +612,52 @@ def is_bond(self):
...
+        cmd = f"cat /sys/class/net/{self.name}/device/devspec"
+        output = run_command(cmd, self.host)
+        if "vnic" in output:
+            return True
+        return False
```
PraveenPenguin, line 643: can utility  is_vnic and is_veth combined name can be like 

```
is_virtualnet(self, mode=''):
     if mode='vnic':
        check_string= 'vnic'
     elif mode='veth':
          check_string= '"l-lan'
     else:
           raise NWException("Not supported Format")
  
   if not os.path.isfile(f"/sys/class/net/{self.name}/device/devspec"):
            raise NWException ("Please provide valid inerface name")
        cmd = f"cat /sys/class/net/{self.name}/device/devspec"
        output = run_command(cmd, self.host)
        if checkstring  in output:
            return True
        return False
  ```

PraveenPenguin (changes requested):

avocado/utils/network/interfaces.py
abdhaleegit: utility is only to know mode.. if it is already known.. than no need to check fro if it is vnic .. you got the point ?

abdhaleegit: with suggested way.. it is like you are passing vnic and returning vnic

PraveenPenguin: yes reasonable just looking to minimize the code uses may be common part can be push in another module

avocado/utils/network/interfaces.py
abdhaleegit: Done

PraveenPenguin (approved):

PR #5595, opened 2023-01-30
Increase connection establish timeout and fix its error object

pevogam:
Signed-off-by: Plamen Dimitrov <plamen.dimitrov@intra2net.com>

### DIFF
--- a/avocado/core/nrunner/task.py
+++ b/avocado/core/nrunner/task.py
@@ -47,7 +47,7 @@ class TaskStatusService:
             host, port = self.uri.split(":")
             port = int(port)
             if self.connection is None:
-                for _ in range(30):
+                for _ in range(600):
                     try:
                         self.connection = socket.create_connection((host, port))
                         break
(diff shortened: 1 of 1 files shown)


avocado/core/nrunner/task.py
```diff
@@ -47,15 +47,15 @@ def post(self, status):
             host, port = self.uri.split(":")
             port = int(port)
             if self.connection is None:
-                for _ in range(30):
+                for _ in range(600):
```
richtja, line 50: Hi @pevogam can you please give me more context to this change? I don't understand why you need to increase the timeout. Thank you.

pevogam: Hi @richtja at least empirically most containers as well as remote hosts could take much longer than 30 seconds in various situations (hardware load, being booted, etc.). As the above loop will be interrupted earlier if the connection is established earlier it won't slow down faster cases but will give more tolerance to slower cases (I have obtained multiple timeouts here due to 30 seconds being too restrictive).

richtja (approved):
Hi @pevogam, LGTM, thanks.

avocado/core/nrunner/task.py
richtja: I see, thanks for the explanation.

